PR #117, opened 2014-05-15
Issue with prompt after inquirer.ui.BottomBar have been declared

fnhipster:


Coriou:
I believe this bug is still in today

SBoudrias:
@Coriou I'll be happy to look at a new PR regarding this bug.

Issue #167, opened 2014-09-30
multi-line validate messages screw up prompt rendering


tonylukasavage:
### problem

Whenever I add `validate` to a question and it returns an error with more than one line of text (either explicitly defined as multi-line in the code, or overrun in the terminal), the subsequent rendering of the prompts gets all muddled up. It appears as though the prompts assume that the error will only be one line.

### test case
```js
var inquirer = require('inquirer');

inquirer.prompt([
	{
		name: 'name',
		type: 'input',
		message: 'name:',
		validate: function(input) {
			if (input !== 'test') {
				return 'this is a multi-line\n' +
					'error message that is going to\n' +
					'really screw up how inquirer.js\n' +
					're-renders the prompts.'
			} else {
				return true;
			}
		}
	}
], function(answers) {
	console.log(answers);
});
```

### results
```
? name: ThisNameFailsValidate
>> this is a multi-line
error message that is going to
? name: 
re-renders the prompts.
```

flovan:
I think I found another bug that overlaps with the behaviour reported above.  
When the default value of a prompt spans across a new line, the error message gets cut off by the amount of characters already present on that line.

Example result with response "this is an error message":
````
? This is a question? (and this is a very long
default answer) WrongAnswer
? This is a question? (and this is a very long
default answer) essage
````

SBoudrias:
Fixed on master. Please test and let me know how it goes.

PR #169, opened 2014-10-01
Fix multi-line error rendering for input type prompts (#167)

tonylukasavage:
There was a relatively shallow, but present, rabbit-hole when addressing this issue. For ease of evaluation, I broke the commits associated with this PR into 3 distinct pieces, which I'll explain below.

#### fix rendering on multi-line errors (2976045)

This resolved the direct issue. It simply moves the cursor up based on the number of individual lines in the erorr message. Pretty simple.

#### prefix each error line with red >> (f110f1a)

When multi-line errors were printed, only the first line was prefixed with the red ">>" symbol. This commit prefixes each line in the resulting error message with that symbol for a more uniform output.

#### add errorHeight for appropriate next render (3368940)

This was the slightly tricky one. Once a multi-line error is rendered correctly, subsequent prompts, or the final result, would be rendered over it, rather than it being cleaned before moving on. For example, using the code in issue #167, the output might look something like this before this commit:

```
≫ node index.js 
? name: test
{ name: 'test' }
>> error message that is going to
(~/development/inq)how inquirer.js
≫  re-renders the prompts.
```

With this commit I added the `this.errorHeight` property that keeps track of the height of the rendered error message, which will then be subsequently be used in the `Prompt.prototype.onEnd()` function to reposition and clear the message upon successful validation. This was the best, most lightweight way I could come up with to both solve the problem and keep the `error()` and `onEnd()` functions loosely coupled.  Now it looks like this:

```
≫ node index.js 
? name: test
{ name: 'test' }
```

### DIFF
--- a/lib/prompts/base.js
+++ b/lib/prompts/base.js
@@ -103,9 +103,10 @@ Prompt.prototype.error = function( error ) {
   var lines = (error || "Please enter a valid value")
     .split(/(?:\r\n|\n|\r)/)
     .map(function(line) { return chalk.red(">> ") + line; });
+  var len = this.errorHeight = lines.length;
   this.write( lines.join(os.EOL) );
 
-  return this.up( lines.length );
+  return this.up( len + this.height - 1 );
 };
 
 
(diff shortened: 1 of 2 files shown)


SBoudrias:
I have a branch out refactoring or rendering that will eventually fix the line wrapping issues.

I'll close this CR as it's pretty old and I don't intend on merging it anymore.

PR #181, opened 2014-10-30
automatically pick the right height for scrollable choices.

### LINKED ISSUE #180: don't hardcode pagination settings
It would be nice if the number of entries that caused paging to be enabled was adjustable, or at least defaulted to something closer to [`process.stdout.rows`](http://nodejs.org/api/tty.html#tty_ws_rows).

```js
Choices.prototype.setRender = function( render ) {
  this.renderingMethod = (this.choices.length > 9) ? this.paginateOutput(render) : render;
};
```
https://github.com/SBoudrias/Inquirer.js/blob/master/lib/objects/choices.js#L143-145

```js
Choices.prototype.paginateOutput = function( render ) {
  var pageSize = 7;
```
https://github.com/SBoudrias/Inquirer.js/blob/master/lib/objects/choices.js#L155

dylang:
fixes #180

### DIFF
--- a/lib/objects/choices.js
+++ b/lib/objects/choices.js
@@ -141,7 +141,7 @@ Choices.prototype.render = function() {
  */
 
 Choices.prototype.setRender = function( render ) {
-  this.renderingMethod = (this.choices.length > 9) ? this.paginateOutput(render) : render;
+  this.renderingMethod = (this.choices.length > process.stdout.rows - 1) ? this.paginateOutput(render) : render;
 };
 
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
@dylang Any chance you could finish this up?

dylang:
@sindresorhus I'd love to but I don't think I'll have time to look at it for a few weeks.

SBoudrias:
So, the code have been changing a lot lately, this is no longer merge-able - and it's been pretty much a dead PR for while. I'll close.

PR #197, opened 2014-12-15
UI: options 'input' & 'output', defaults to ttys

naholyr:
Rebased version of #188

### DIFF
--- a/lib/inquirer.js
+++ b/lib/inquirer.js
@@ -23,8 +23,8 @@ inquirer.ui = {
  * Create a new self-contained prompt module.
  */
 inquirer.createPromptModule = function () {
-  var promptModule = function( questions, allDone ) {
-    var ui = new inquirer.ui.Prompt( promptModule.prompts );
+  var promptModule = function( questions, allDone, opt ) {
+    var ui = new inquirer.ui.Prompt( promptModule.prompts, opt );
     ui.run( questions, allDone );
     return ui;
   };
--- a/lib/ui/baseUI.js
+++ b/lib/ui/baseUI.js
@@ -5,6 +5,7 @@
 var _ = require("lodash");
 var tty = require("../utils/tty");
 var readlineFacade = require("readline2");
+var ttys = require("ttys");
 
 
 /**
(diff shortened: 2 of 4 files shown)


rmg:
Similar to @adriengibrat for me. I've got yeomen generators failing in CI all over the place once 0.8.1 was released.

SBoudrias:
@rmg I think your issue is actually not the same, last release broke in non tty (which is kinda expected, it was weird it ever worked - but we're working on a fix)

For your Yeoman-Generator, use the RunContext to test them. yeoman.io/authoring/testing.html

rmg:
@SBoudrias the generator _is_ using the test helpers and the stacktrace shows `var helpers = require('yeoman-generator').test;` right near the bottom.

To be clear:
 * the error is `Error: ENXIO, no such device or address '/dev/tty'`
 * it comes from [TooTallNate/ttys/index.js:9](https://github.com/TooTallNate/ttys/blob/2bf157923599f71a3269e24c9093b4198e1fdb6e/index.js#L9)
 * is triggered on load via `require('yeoman-generator')`

The root of the problem is that the ttys module assumes there actually is a TTY to attach to. The result is this PR adds that same erroneous assumption to inquirer.js, which in turn adds it to yeoman.

Since tty hasn't been updated in 3 years (as far as I can tell), I'm inclined to think that `require('ttys');` added by this PR should be wrapped in a try/catch rather than relying on an update to the tty module.

rmg:
Forgot to mention the reproduction steps:
 1. `ssh -T localhost "cd $PWD; npm test"`  (where $PWD is some generator's checkout)
 2. there is no step 2, `ssh -T localhost` is a quick'n'dirty way to drop your TTY.

SBoudrias:
I actually unpublished 0.8.1, will investigate this more deeply, for the time being you can still use 0.8.0

n2liquid:
Can't wait to implement a line picker once this is implemented to do things like this:

    $ vim $(find . -name config.json | picklines)
    ? Pick lines: path/1/to/config.json, path/2/to/config.json
     ◉ path/1/to/config.json
    ❯◉ path/2/to/config.json
     ◯ path/to/some-other/config.json
    2 files to edit
    [Vim session]

It's hard to do that when stdin / stdout are being used by Inquirer itself :(

PR #204, opened 2015-01-20
get rid of some moot dependencies

sindresorhus:
and some minor other cleanup

`mute-stream` isn't used anywhere

`cli-color` is moot as you can do the same without

-

Can you push a new release when this is merged?

### DIFF
--- a/lib/prompts/rawlist.js
+++ b/lib/prompts/rawlist.js
@@ -4,7 +4,6 @@
 
 var _ = require("lodash");
 var util = require("util");
-var clc = require("cli-color");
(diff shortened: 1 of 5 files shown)


lib/prompts/rawlist.js
```diff
@@ -132,7 +131,6 @@ Prompt.prototype.onError = function() {
   this.hasError = true;
   this
     .error("Please enter a valid index")
-    .write( clc.bol(0, true) )
```
sindresorhus: @SBoudrias This actually fixes a bug where the message above was not shown on error.

PR #218, opened 2015-02-24
Add password hidden input type so that password character length isnt shown

timmyg:


SBoudrias:
Pretty sure you're not on the right HEAD here...

timmyg:
eek, going to close and redo

PR #219, opened 2015-02-24
adding password prompt that doesnt display any characters; resolves #177

### LINKED ISSUE #177: Password prompt should display no characters
Password prompt should display no characters when entering text, or there should be an option to enable this behaviour. This would mimic the behaviour of applications like ssh-add.

timmyg:


SBoudrias:
What I meant on the ticket was to create a custom prompt as an external plugin. I'm currently working towards making the creation of custom prompt easier, but it's totally possible to do it today.

PR #222, opened 2015-02-27
Updated lodash and the rest of the "dependencies" to the latest versions...

igorlino:
Grunt tests are all passing without errors.

Solves https://github.com/SBoudrias/Inquirer.js/issues/220


SBoudrias:
Thanks!

Issue #233, opened 2015-04-10
[feature] Construct custom prompts types


dakal-oleksandr:
Hi. 
I added little feature here:
https://github.com/dakal-oleksandr/Inquirer.js/commit/d265d20894ca910cfb8f4e78c72f544b4a94b195

`bottomText` will give ability to render additional information under list that can depend on selected choice (for example it's useful to render long hint to selected list item). Bottom text is added in `Prompt.prototype.render` so `utils.writeMessage` will include additional text in height and prompt clean will work correctly.

Please review changes. And give your opinion

dakal-oleksandr:
So I had a lot of experiments with Inquirer and the best solution is to add functionality to create custom prompts types based on already available prompts.

I created pull request with proposed changes

PR #238, opened 2015-04-27
#233 [feature] Construct custom prompts types

dakal-oleksandr:
Functionality to create custom prompt types.

### DIFF
--- a/examples/custom-prompt.js
+++ b/examples/custom-prompt.js
@@ -1,0 +1,46 @@
+/**
+ * List prompt example
+ */
+
+"use strict";
+var inquirer = require("../lib/inquirer");
+var util = require('util');
+
+var BaseList = inquirer.prompt.getPromptConstructor('list');
+
+var CustomPrompt = function () {
+  BaseList.apply(this, arguments);
+};
+
+util.inherits( CustomPrompt, BaseList );
+
+CustomPrompt.prototype.prefix = function( str ) {
+  str || (str = "");
+  return "Question: " + " " + str;
+};
+
+inquirer.registerPrompt('custom', CustomPrompt);
+
+inquirer.prompt([
+  {
+    type: "custom",
+    name: "theme",
+    message: "What do you want to do?",
+    choices: [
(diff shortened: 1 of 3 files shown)


SBoudrias:
Hey, I see what you're trying to accomplish here, but I don't think that's a maintainable solution.

Exposing random modules and function as class property is probably the biggest technical debt you'll encounter anywhere. It freeze interfaces and tightly couple everything.

This is definitively not the way I want to go about providing base helpers to help the community to create custom prompts.

The way forward is to decouple all the class based helpers functions there's right now in Inquirer to utilities and then create an external packages out of them. We could then do the same thing with the base prompt class if necessary.

We would also need to clarify the prompt interface. What is a prompt object? How does it work? Clearly define the responsibility of each part of the inquirer system.

It is important these things stay versioned out of inquirer so people can reliably expect them to work even if we do breaking change over here. Also, we have a growing use case of user who abstract GUI on top of Inquirer, so considering their use case and how they can leverage our core system without rewriting the whole thing would be a welcome addition.

dakal-oleksandr:
>The way forward is to decouple all the class based helpers functions there's right now in Inquirer to utilities and then create an external packages out of them. We could then do the same thing with the base prompt class if necessary.

You mean access helpers from Inquirer Client interfaces. Same approch like Separator? 
`inquirer.Utils= require("./utils/utils");`
`inquirer.Observe= require("./utils/events");`

>We would also need to clarify the prompt interface. What is a prompt object? How does it work? Clearly define the responsibility of each part of the inquirer system.

Well that means that we need to define proper interface for already existing Prompts. Am I correct?

SBoudrias:
> You mean access helpers from Inquirer Client interfaces. Same approch like Separator? 

No, I mean like separate packages `require('inquirer-util')`. But before you jump there, we need to define interfaces in Inquirer so they're easy to separate.

> Well that means that we need to define proper interface for already existing Prompts. Am I correct?

Yes, for the existing stuff. This is a hard work.

PR #239, opened 2015-04-27
Fix input prompt repeating value when value is longer than terminal width. Closes #214

### LINKED ISSUE #214: inputs repeating question when input wraps
Sorry for the rubbish title, this is better explained by example.

screen shot 2015-02-04 at 16 19 58

If the input goes onto the next line, the question repeats for every character over the new line. It's a bit strange.

I'm using zprezto in iTerm2. Let me know if you need to know anything else.

jviotti:
The problem happens because inquirer cleans only the current line on
each key stroke, without paying attention to the actual length of the
value, which might be longer than one line.

This PR was only tested on OS X. Help from people with other OSes would be appreciated.

### DIFF
--- a/lib/prompts/input.js
+++ b/lib/prompts/input.js
@@ -2,6 +2,7 @@
  * `input` type prompt
  */
 
+var os = require("os");
 var _ = require("lodash");
 var util = require("util");
 var chalk = require("chalk");
@@ -82,7 +83,7 @@ Prompt.prototype.onEnd = function( state ) {
     this.status = "answered";
 
     // Re-render prompt
-    this.clean(1).render();
+    this.clean( this.lines() ).render();
 
     // Render answer
     this.write( chalk.cyan(filteredValue) + "\n" );
@@ -100,5 +101,14 @@ Prompt.prototype.onError = function( state ) {
  */
 
 Prompt.prototype.onKeypress = function() {
-  this.clean().render().write( this.rl.line );
+  var lines = this.lines();
+
+  // Windows duplicates the line only when the
+  // first character wraps to another line
+  if ( os.platform() === "win32" && this.previousLines === lines + 1 ) {
+    this.clean( 1 ).render();
+  }
+
+  this.previousLines = lines;
+  this.clean( this.previousLines - 1 ).render().write( this.rl.line );
 };
--- a/lib/utils/tty.js
+++ b/lib/utils/tty.js
@@ -2,11 +2,54 @@
  * TTY mixin helpers
  */
 
+var os = require("os");
 var _ = require("lodash");
 var readline = require("readline");
-var cliWidth = require("cli-width");
+var unicodeLength = require("unicode-length");
+var utils = require("./utils");
 
 var tty = module.exports;
+
+/**
+ * Get the number of lines the current line occupies
+ * @return {Number}  The number of lines
+ */
+
+tty.lines = function() {
+  var terminalWidth = utils.cliWidth();
+
+  if ( terminalWidth === 0 ) {
+
+    // Preserve original behaviour
+    return 1;
+  }
+
+  var value = this.rl.line;
+
+  if ( !value ) {
+    if ( this.rl.history ) {
+      value = this.rl.history[0];
+    } else {
+      value = this.opt.default;
+    }
+  }
+
+  var lineLength = unicodeLength.get( this.rl._prompt + value );
+
+  // Windows seems to wraps the line one character
+  // before the window width for some reason
+  if ( os.platform() === "win32" ) {
+    lineLength += 1;
+  }
+
+  var result = lineLength / terminalWidth;
+
+  if ( result % 1 === 0 ) {
+    return result;
+  } else {
+    return Math.floor( result ) + 1;
+  }
+};
 
 
 /**
@@ -21,7 +64,7 @@ tty.clean = function( extra ) {
   var len = this.height + extra;
 
   while ( len-- ) {
-    readline.moveCursor(this.rl.output, -cliWidth(), 0);
+    readline.moveCursor(this.rl.output, -utils.cliWidth(), 0);
     readline.clearLine(this.rl.output, 0);
     if ( len ) readline.moveCursor(this.rl.output, 0, -1);
   }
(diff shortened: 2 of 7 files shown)


lib/utils/utils.js
```diff
@@ -121,3 +122,25 @@ utils.writeMessage = function ( prompt, message ) {
...
+
+/**
+ * Helper for getting the length of a string containing unicode symbols
+ * From https://github.com/mathiasbynens/esrever/blob/master/scripts/export-data.js
+ *
+ * @param {String} input  - The unicode string
+ * @return {Number}       - The string length
+ */
+
+var REGEX_SYMBOLS = /([\0-\u02FF\u0370-\u1DBF\u1E00-\u20CF\u2100-\uD7FF\uDC00-\uFE1F\uFE30-\uFFFF]|[\uD800-\uDBFF][\uDC00-\uDFFF]|[\uD800-\uDBFF])([\u0300-\u036F\u1DC0-\u1DFF\u20D0-\u20FF\uFE20-\uFE2F]+)/g;
```
jviotti: This line gives a long line warning from the linter. What can we do about it?

callumacrae: Most linters can be configured to ignore lines with regular expressions in, for that exact reason. You can use the constructor function, but then you're compromising code quality, which isn't great.

callumacrae:
Is there any way you can add a test for this?

jviotti:
@callumacrae Let me see what can I do.

lib/utils/tty.js
```diff
@@ -5,9 +5,27 @@
...
 
+/**
+ * Get the number of lines the current line occupies
+ * @return {Number}  The number of lines
+ */
+
+tty.lines = function() {
+  var value = this.rl.line || this.rl.history[0];
+  var lineLength = utils.getUnicodeStringLength( this.rl._prompt + value );
+  var result = lineLength / cliWidth();
```
SBoudrias: `cliWidth` can return 0 which is definitively not what we want.

jviotti: @SBoudrias In which cases can it return 0?

SBoudrias: In this case: https://github.com/knownasilya/cli-width/blob/master/index.js#L17

Pretty sure all the time on Windows.

jviotti: Hm, I guess my fix will not work in Windows in that case. Let me borrow a PC and give it a try there. Maybe there is an alternative way to get the terminal width in Windows.

SBoudrias: You can get free Windows VM to test on https://www.modern.ie/en-us/virtualization-tools#downloads 

Pretty sure you can install git and checkout inquirer with them.

jviotti: @SBoudrias I've tested in Windows 8.1 and `cliWidth` returns the correct width on that platform. I had to implement some windows-specific workaround for me patch to work fine, but I can confirm it works as expected on `cmd.exe`.

There still exist the possibility of `cliWidth()` returning 0, maybe in older Windows versions? I propose checking the return value of `cliWidth`, and if it is zero, don't make use of my patch, thus resulting in the actual behaviour. What do you think?

callumacrae: That's what I would vote for :smiley:

jviotti:
Works great in Ubuntu as well.

lib/utils/tty.js
SBoudrias: Seems like a potential workaround.

Another is to manually handle line returns so we stay consistent. But that might just bring new kinds of issues.

jviotti: @callumacrae @SBoudrias Workaround implemented. `tty.lines` now returns `1` if `cliWidth` is 0, which, preserves the original behaviour.

jviotti:
I was able to write some unit tests that check that the corresponding number of lines are cleaned.

test/specs/prompts/input.js
```diff
@@ -53,4 +54,71 @@ describe("`input` prompt", function() {
...
+    it("should clean short lines appropriately", function( done ) {
+      var prompt = new Input( this.fixture, this.rl );
+      var cleanSpy = sinon.spy( prompt, "clean" );
+      prompt.run(function() {
+        expect( cleanSpy.callCount ).to.equal(1);
+        expect( cleanSpy.calledWith(1) ).to.be.true;
+        done();
+      }.bind(this));
+
+      this.rl.line = "hello";
```
jviotti: For some reason, I have to assign `this.rl.line` directly, as emitting a line event doesn't assign the actual line to neither `this.rl.line` or `this.rl.history`, where `tty` expects them.

lib/prompts/input.js
```diff
@@ -82,7 +83,7 @@ Prompt.prototype.onEnd = function( state ) {
     this.status = "answered";
 
     // Re-render prompt
-    this.clean(1).render();
+    this.clean( this.lines() ).render();
```
jviotti: @SBoudrias Do you have any insight on how to get this function (`Prompt.prototype.onEnd`) called in the unit tests? Currently only `onKeypress` seems to be called there, even when emitting a line event.

test/specs/prompts/input.js
SBoudrias: That is correct as we use a mocked readline. So there's no usual Readline behavior during testing.

lib/prompts/input.js
SBoudrias: Sorry, I'm not sure I understand this question. Mind clarifying?

jviotti: @SBoudrias If you take a look at my latest changes in the PR, I was able to unit test `Prompt.prototype.onKeypress()` by using `this.rl.emit()`, however calling `this.rl.emit()` in the unit tests doesn't call `Prompt.prototype.onEnd()` (it does call it when not mocking readline), so I wonder if you know a way to make `Prompt.prototype.onEnd()` get called as well.

jviotti:
@SBoudrias  Is there anything else needed for the PR to be merged?

SBoudrias:
Hey, I'd have a couple comments on the code changes, but I'll hold onto them for now because this fix is very specific to the input prompt and I have a solution in mind that would fix this issue for every prompt type. I'm experimenting with it ATM and just need some more time to make sure it'll work fine.

SBoudrias:
I'm going forward with the rendering process redesign. I'll close this PR as I don't intend to merge it.

HolySuicidalTurkey:
seriously, please, **please** merge this PR. We're still having problems with it 6 years later.

PR #243, opened 2015-05-11
'checked' improperly handles truthy values

### LINKED ISSUE #242: Truthy value for 'checked' shows selection, but item is missing from answers
Suppling a truthy value to 'checked' results in the list item being selected visually, but it is missing from the selected elements list.

The following test case fails as the second item is omitted from the answers:

```js
  it("provide an array of checked choice to validate", function( done ) {
    this.fixture.choices = [
      { name: "1", checked: true  },
      { name: "2", checked: 1 },
      { name: "3", checked: false }
    ];
    this.fixture.validate = function( answer ) {
      expect(answer).to.eql([ "1", "2" ]);
      return true;
    };
    this.checkbox = new Checkbox( this.fixture, this.rl );
    this.checkbox.run(function() {
      done();
    });
    this.rl.emit("line");
  });
```

plemarquand:
Suppling a truthy value to 'checked' resulted in the list item
being selected visually, but it was missing from the selected
elements list.

'checked' is now coerced to a boolean when selected items are
queried.

This patch closes #242.

### DIFF
--- a/test/specs/prompts/checkbox.js
+++ b/test/specs/prompts/checkbox.js
@@ -71,11 +71,11 @@ describe("`checkbox` prompt", function() {
   it("provide an array of checked choice to validate", function( done ) {
     this.fixture.choices = [
       { name: "1", checked: true  },
-      { name: "2", checked: false },
+      { name: "2", checked: 1 },
       { name: "3", checked: false }
     ];
     this.fixture.validate = function( answer ) {
-      expect(answer).to.eql([ "1" ]);
+      expect(answer).to.eql([ "1", "2" ]);
       return true;
     };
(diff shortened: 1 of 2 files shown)


test/specs/prompts/checkbox.js
```diff
@@ -85,6 +85,8 @@ describe("`checkbox` prompt", function() {
     this.rl.emit("line");
   });
 
+
```
SBoudrias: can you remove the extra whitespace

plemarquand: @SBoudrias Done

SBoudrias:
Thanks!

PR #246, opened 2015-05-21
Fix #214 by not redrawing the question on keypress

### LINKED ISSUE #214: inputs repeating question when input wraps

FremyCompany:
I found out there was actually no good reason to do this. At least, I didn't find one, and removing this code make the input works for me again (I'm doing some sort of REPL so it's frequent to see my input wrapping).

### DIFF
--- a/lib/prompts/input.js
+++ b/lib/prompts/input.js
@@ -100,5 +100,5 @@ Prompt.prototype.onError = function( state ) {
  */
 
 Prompt.prototype.onKeypress = function() {
-  this.clean().render().write( this.rl.line );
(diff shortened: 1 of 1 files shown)


SBoudrias:
We'll take care of input wrapping in a correct fix. This is a very specific hack to the input prompt.

PR #252, opened 2015-06-22
Change the separator color

stevemao:
Just to make it more obvious its a separator.

Before: 
screen shot 2015-06-22 at 4 29 03 pm

After:
screen shot 2015-06-22 at 4 28 29 pm

My instinct always told me that "Run a generator" was selectable and I ended up selecting "Get me out of here!" :cry: 
This change makes it very clear that "Run a generator" is not selectable and it's obviously a part of the separator.

### DIFF
--- a/lib/objects/separator.js
+++ b/lib/objects/separator.js
@@ -22,7 +22,7 @@ module.exports = Separator;
 
 function Separator( line ) {
   this.type = "separator";
-  this.line = chalk.dim(line || new Array(15).join(figures.line));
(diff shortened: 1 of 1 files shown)


SBoudrias:
I don't think this is clearer - the line actually look like the currently selected element.

stevemao:
What would you expecting? expose the color as the second argument?

SBoudrias:
Users can overwrite it in their own programs.

stevemao:
oh ok

PR #254, opened 2015-06-25
make the current answers available within validate fn

### LINKED ISSUE #198: access anwsers in validate callback
is it possible to retrieve the answers within the validate callback?

the scenario is that i wanna validate an input by the previous answers.

Whoaa512:
fix #198 

This was the simplest fix I could see. If you don't like making the answers available on the prompt itself, I'm open to suggestions about how else it could be done.


SBoudrias:
Hey, thanks for taking care of this (and providing a unit test +++)

PR #255, opened 2015-06-29
Fix wrapping bug on lists

adriancooney:
This should fix the issue #214 where wrapping lines caused extra lines to be added to the prompt. The problem was where the lines wrapped and it caused the screens count of lines to clear to fall out of sync. I'm not exactly sure how to unit test this but from manual tests, it seems to work fine. Another potential bug is when the terminal resizes, it will go out of sync because it only updates the line count on render so that will need to be accounted for in future.

#### Behaviour differences
Here is a demonstration of the behaviour differences introduced by this PR:
* Current (`master`): https://asciinema.org/a/c11odt2jpxncuz7jmdaymefch
* PR changes (`wrapping-bug`): https://asciinema.org/a/bsaszkvz3skkixpgtuzoy40re

### DIFF
--- a/lib/utils/screen-manager.js
+++ b/lib/utils/screen-manager.js
@@ -23,8 +23,13 @@ ScreenManager.prototype.render = function (content, opt) {
   this.rl.output.unmute();
   this.clean(this.prevCursor);
 
-  var lines = content.split(/\n/);
-  this.height = lines.length;
+  var lines = content.split(/\n/),
+    width = cliWidth();
+
+  this.height = lines.reduce(function(count, line) {
+    count += Math.ceil(line.length / width);
+    return count;
+  }, 0);
 
   // Write message to screen and setPrompt to control backspace
   var prompt = lines[lines.length - 1 - opt.cursor];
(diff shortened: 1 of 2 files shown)


lib/utils/screen-manager.js
```diff
@@ -23,8 +23,13 @@ ScreenManager.prototype.render = function (content, opt) {
   this.rl.output.unmute();
   this.clean(this.prevCursor);
 
-  var lines = content.split(/\n/);
-  this.height = lines.length;
+  var lines = content.split(/\n/),
+    width = cliWidth();
```
SBoudrias: use one `var` statement per line.

adriancooney: Sorry, fixed!

adriancooney:
Actually, there's an issue with this. It count's hidden characters such as color and incorrectly assumes the length of the line. Actual length of the line needs to be detected either by rendering then counting or somehow flattening colors and text decoration.

SBoudrias:
Yeah, you can use `strip-ansi` for that. I have a branch on the side that's doing a very similar logic. Another issue with your implementation is that it won't consider the cursor position if you're outsetting it with a long line. Also, I'm pretty sure  it'll break our input prompts where the user answer is wrapping.

So this is not mergeable as is, but I might base something out of your solution.

adriancooney:
That's cool! Glad I could help somehow. Any rough ETA on a fix?

SBoudrias:
My fix is on master now.

Issue #225, opened 2015-03-12
auto-complete list


malandrew:
It would be super nice to have an autocomplete list input type which is a hybrid of the `input` and `list` input types. Having it be capable of async (so you can make remote requests and redraw the filtered options in the terminal) would be a plus. 

(cc: @lxe)

yangchenyun:
+1

yangchenyun:
Hi SBoudrias@, I notice that you are also author of this awesome repo!

How much work do you evaluate to implement this feature? If this is within reasonable time, I could implement it and send you PR for review.

SBoudrias:
I think it'd be fairly simple. A couple hours at most, but it might be more since you might need to get used to the core principles - and the core of Inquirer date from ~2 years ago, so it's not the cleanest code I've written out there.

yangchenyun:
Wow, I didn't know this almost 2 years old! So far this is most friendly module to work with terminal app. 

Could you point me to some main abstractions you have defined in the core to help me start: 
- Abstraction to work the terminal
- Abstraction to work with user input

SBoudrias:
Oh, I think the public API is fairly simple and pretty feature complete. I was more thinking of the behind the scenes abstractions who're bloated around classes. Redoing the core code today, most of this would be in functional style as a lot of class base abstraction currently in use are not needed. But you should be able to wrap your head around it.

I think the list prompt implementation is the most similar to the suggested kind of prompt: https://github.com/SBoudrias/Inquirer.js/blob/master/lib/prompts/list.js

We use RxJS to handle user inputs https://github.com/SBoudrias/Inquirer.js/blob/master/lib/prompts/list.js#L65-L69

Then you just react to those events by writing and clearing the terminal with the new layout.

btakita:
is this still in the works?

patrick-motard:
Is this request different from https://github.com/SBoudrias/Inquirer.js/pull/265, the auto-complete prompt? If so, should this be closed?

SBoudrias:
Right this is now published as a standalone plugin https://github.com/mokkabonna/inquirer-autocomplete-prompt

SBoudrias:
Maybe of interest, PR open to include such a prompt in the core https://github.com/SBoudrias/Inquirer.js/pull/1479

PR #265, opened 2015-07-31
Feature: Autocomplete prompt

mokkabonna:
I have created a POC for a autocomplete prompt. See the example for a live example.

Is this something that would be considered for inclusion in this library?

I have tried to use the api as used by other prompts although when introducing async I feel it is a bit weird. Any suggestions on how to do this instead? like the choices:

If you now define a function as your choices (as with other prompts) you have to return a inner function that gets called whenever the users input changes. This works fine, but maybe could be done better, and might be confusing.

Fake search function below (from the example): 
```javascript
choices: function (answers) {
  return function(searchTerm) {
    return new Promise(function(resolve, reject) {
      setTimeout(function() {
        resolve(states.filter(function(state) {
          return new RegExp(searchTerm, 'i').exec(state) !== null;
        }));
      }, 400); // simulate 400 ms delay for async request
    });
  };
}
```

Since the choices also aren't available at creation of the prompt it won't create a Choices object like normally, so I am creating that each time the search function is returning it's result. Is this necessary? Ideally I would just use a simple array, but there might be good reasons why you want them created as choice objects.

It is also an assumption that the search function returns a promise. I feel that promises are the way to go with async these days. Thoughts?

This PR is not complete, but meant as a proof of concept and as as place to discuss it. I have however already some tests.

I have also added a grunt develop task, just for my own sake really, I'll rebase and remove if not wanted.

The style and formatting of the code might also be a bit off, will gladly comply to whatever style is desired, just wasn't my focus right now.

As suggested here https://github.com/SBoudrias/Inquirer.js/issues/225 I have based it of the list prompt.

image

### DIFF
--- a/lib/prompts/autocomplete.js
+++ b/lib/prompts/autocomplete.js
@@ -61,6 +61,7 @@ Prompt.prototype._run = function(cb) {
   events.line.take(1).forEach(self.onSubmit.bind(this));
   events.keypress.takeUntil(events.line).forEach(self.onKeypress.bind(this));
 
+  //call once at init
   self.currentPromise = self.opt.choices();
 
   self.currentPromise.then(function inner(choices) {
@@ -130,7 +131,6 @@ Prompt.prototype.onSubmit = function() {
 };
 
 Prompt.prototype.search = function() {
-
   var self = this;
   self.selected = 0;
 
@@ -138,6 +138,7 @@ Prompt.prototype.search = function() {
   self.currentChoices = new Choices([]);
   self.render(); //now render current searching state
 
+  self.lastSearchTerm = this.rl.line;
   self.currentPromise = self.opt.choices(self.rl.line);
 
   self.currentPromise.then(function inner(choices) {
@@ -164,7 +165,11 @@ Prompt.prototype.onKeypress = function(e) {
     this.render();
   } else {
     this.render(); //render input automatically
-    this.search(); //trigger new search
+
+    //Only search if input have actually changed, not because of other keypresses
+    if (this.lastSearchTerm !== this.rl.line) {
+      this.search(); //trigger new search
+    }
   }
 };
 
--- a/lib/prompts/autocomplete.js
+++ b/lib/prompts/autocomplete.js
@@ -1,0 +1,198 @@
+/**
+ * `list` type prompt
+ */
+
+var _ = require("lodash");
+var util = require("util");
+var chalk = require("chalk");
+var figures = require("figures");
+var Base = require("./base");
+var Choices = require("../objects/choices");
+var observe = require("../utils/events");
+var utils = require("../utils/readline");
+var Paginator = require("../utils/paginator");
+
+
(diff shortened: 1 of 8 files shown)


lib/prompts/autocomplete.js
```diff
@@ -0,0 +1,203 @@
...
+  if (!this.opt.choices) {
+    this.throwParamError("choices");
+  }
+
+  this.currentChoices = [];
+
+  this.firstRender = true;
+  this.selected = 0;
+
+  var def = this.opt.default;
```
mokkabonna: Not sure what to do with default for this prompt, since we have a external unknown at runtime resultset. If you want a default, maybe have the service return that choice if no match?

lib/prompts/autocomplete.js
```diff
@@ -0,0 +1,203 @@
...
+  var events = observe(self.rl);
+
+  events.line.take(1).forEach(self.onSubmit.bind(this));
+  events.keypress.takeUntil(events.line).forEach(self.onKeypress.bind(this));
+
+  //call once at init
+  self.currentPromise = self.opt.choices();
+
+  self.currentPromise.then(function inner(choices) {
+    choices = new Choices(choices);
```
mokkabonna: Creating a choices object with the resultset when done.

lib/prompts/autocomplete.js
```diff
@@ -0,0 +1,203 @@
...
+  this.done(choice.value);
+
+};
+
+Prompt.prototype.search = function() {
+  var self = this;
+  self.selected = 0;
+
+  self.searching = true;
+  self.currentChoices = new Choices([]);
```
mokkabonna: Empty results as we are about to search and the old results aren't valid anymore.

lib/prompts/autocomplete.js
```diff
@@ -0,0 +1,203 @@
...
+  self.render(); //now render current searching state
+
+  self.lastSearchTerm = this.rl.line;
+  self.currentPromise = self.opt.choices(self.rl.line);
+
+  self.currentPromise.then(function inner(choices) {
+    choices = new Choices(choices);
```
mokkabonna: Again, create Choices object.

derhuerst:
Looking forward to see this getting merged. :+1:

lazabogdan:
+1

SBoudrias:
Hey! Thanks for working on this.

This prompt should be published as it's own npm package. Inquirer allow to be extended with custom prompt type:

```js
inquirer.registerPrompt('search', require('your-module');
inquirer.prompt({ type: 'search', ... });
```

mokkabonna:
For anyone interested, I have started progress on the prompt here: https://github.com/mokkabonna/inquirer-autocomplete-prompt

Not published as a npm package yet. Feedback is welcome.

patrick-motard:
Looks like inquirer-autocomplete-prompt is now published on npm.

PR #271, opened 2015-08-18
Mask password default

### LINKED ISSUE #270: Password default should be masked
When using a default for the password it should be masked.

Eg:
```sh
? What is your npm password? (**************)
```

My use case is that I want to store the password with [keytar](https://github.com/atom/node-keytar) and want to allow the user to change it.

jo:
Closes #270.


SBoudrias:
See my answer on the issue. This should be handle on the user side.

PR #280, opened 2015-09-05
Adds an option to configure input/output streams, fixes #279

### LINKED ISSUE #279: Allow setting input and output
Resurrecting the ideas from #188 #197 

I just stumbled upon a use case where having the `ttys` module support that @naholyr added would have saved the day (basically reading from stdin and prompting user with info taken from there).

By reading through the history of these PRs it seems to me that `ttys` turned out to be a very unreliable module that unfortunately only worked on some terminals and eventually all the changes got reverted :disappointed: 

@SBoudrias would you be open to implement back the possibility of setting input/output but defaulting to `process.stdin`/`process.stdout` instead? That would at least allow use cases like the one I have to deal with now, to opt-in for `ttys` or to use some other better solution in the future.

ruyadorno:
Adds a third option to the `inquirer.prompt` method to allow for configurable input/output streams.

When no value is supplied defaults to `process.stdin`/`process.stdout`.

@SBoudrias

### DIFF
--- a/lib/inquirer.js
+++ b/lib/inquirer.js
@@ -22,9 +22,9 @@ inquirer.ui = {
 /**
  * Create a new self-contained prompt module.
  */
-inquirer.createPromptModule = function () {
+inquirer.createPromptModule = function (opt) {
   var promptModule = function (questions, allDone) {
-    var ui = new inquirer.ui.Prompt(promptModule.prompts);
+    var ui = new inquirer.ui.Prompt(promptModule.prompts, opt);
     ui.run(questions, allDone);
     return ui;
   };
--- a/lib/ui/baseUI.js
+++ b/lib/ui/baseUI.js
@@ -10,7 +10,9 @@ var readlineFacade = require('readline2');
 var UI = module.exports = function (opt) {
   // Instantiate the Readline interface
   // @Note: Don't reassign if already present (allow test to override the Stream)
-  this.rl || (this.rl = readlineFacade.createInterface());
+  if (!this.rl) {
+    this.rl = readlineFacade.createInterface(opt);
+  }
   this.rl.resume();
 
   this.onForceClose = this.onForceClose.bind(this);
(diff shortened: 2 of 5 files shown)


lib/ui/baseUI.js
```diff
@@ -21,7 +29,7 @@ var UI = module.exports = function (opt) {
   process.on('exit', this.onForceClose);
 
   // Propagate keypress events directly on the readline
-  process.stdin.addListener('keypress', this.onKeypress);
+  this._input.addListener('keypress', this.onKeypress);
```
SBoudrias: I'm fairly certain we can just call `this.rl.output`/`this.rl.input` without requiring us to keep the reference around.

lib/inquirer.js
```diff
@@ -23,8 +23,8 @@ inquirer.ui = {
  * Create a new self-contained prompt module.
  */
 inquirer.createPromptModule = function () {
-  var promptModule = function (questions, allDone) {
-    var ui = new inquirer.ui.Prompt(promptModule.prompts);
+  var promptModule = function (questions, allDone, opt) {
```
SBoudrias: I think `inquirer.createPromptModule()` should take input/output options. `inquirer.prompt` should stay as is.

lib/ui/baseUI.js
```diff
@@ -43,12 +51,12 @@ UI.prototype.onForceClose = function () {
...
   // Remove events listeners
   this.rl.removeListener('SIGINT', this.onForceClose);
-  process.stdin.removeListener('keypress', this.onKeypress);
+  this._input.removeListener('keypress', this.onKeypress);
   process.removeListener('exit', this.onForceClose);
 
   // Restore prompt functionnalities
   this.rl.output.unmute();
-  process.stdout.write('\x1B[?25h'); // show cursor
+  this._output.write('\x1B[?25h'); // show cursor
```
SBoudrias: FWIW you can delete this line. I forgot to remove it when I moved cursor support to cli-cursor.

ruyadorno:
:+1: addressed all comments

SBoudrias:
Awesome, thanks a ton!

ruyadorno:
:confetti_ball:

Issue #290, opened 2015-09-29
SyntaxError thrown if CLI width cannot be obtained


vsimonian:
If cli-width returns `0`, which is [possible in certain environments](https://github.com/anvilresearch/connect-cli/issues/51), Inquirer will crash on [line 108 of `lib/utils/screen-manager.js`](https://github.com/SBoudrias/Inquirer.js/blob/84b2bbcac664bf26ffb4b2a51d2354807bcb2260/lib/utils/screen-manager.js#L108):

```javascript
  var regex = new RegExp('.{1,' + cliWidth() + '}', 'g');
```

`/.{1,0}/g` is an invalid regex as the max number of characters is smaller than the minimum. This throws an error:

```
SyntaxError: Invalid regular expression: /.{1,0}/: numbers out of order in {} quantifier.
```

The cli-width package supports setting the default width to a number other than `0` by setting the value of ` cliWidth.defaultWidth`. It seems reasonable to have a default of `80` instead of `0`.

artem-malko:
:+1: got this bug on WIndows.

artem-malko:
@SBoudrias is there any news about this bug?

SBoudrias:
@artem-malko Feel free to send a PR.

lostPixels:
Just encountered this one too when users of my app were attempting to execute a script in the Intellij shell.

jviotti:
I've sent a PR for setting the default value to `80`. I've experimented with it in a Windows 10 PC and seems to be working fine.

Omitting tests since `screen-manager.js` it not being unit tested at all.

sbritoig:
A temporary fix is setting an environment variable named CLI_WIDTH to anything greater than 1.

PR #297, opened 2015-10-15
Set a default cli width of 80 characters

jviotti:
In some environment, `cli-width` is known to return a default value of
`0`, which causes inquirer to crash in such a context.

Fixes: https://github.com/SBoudrias/Inquirer.js/issues/290


SBoudrias:
Thanks!

PR #298, opened 2015-10-15
Remove unused dependencies in lib/

jviotti:


jviotti:
Found them while making the previous PR.

SBoudrias:
Ah, thanks for the help

PR #300, opened 2015-11-04
added pageSize option for prompts that use paginator

id2iose:
Prompts that use the paginator (checkbox, expand, list, and rawList) can now be passed an optional pageSize parameter. Defaults to 7.


SBoudrias:
Thanks!

PR #307, opened 2015-11-23
implements lazy-caching for all module dependencies

jonschlinkert:
Reduces (passive) init time from ~110ms to 18ms. Let me know if you have any questions about implementation specifics, or want anything changed. 

***
## before 

Initialization averaged between 95-120ms

screen shot 2015-11-22 at 7 23 48 pm

## after

Initialization averages 17-18ms. 

screen shot 2015-11-22 at 8 16 27 pm

With lazy caching implemented, half of the passive init time now comes from the following line of code:

```js
inquirer.prompt = inquirer.createPromptModule();
```

If we can make `prompt` a getter (depending on when `createPromptModule()` needs to be initialized), it will take init time down to 10ms.

### DIFF
--- a/lib/inquirer.js
+++ b/lib/inquirer.js
@@ -1,3 +1,5 @@
+require('time-require');
+
 /**
  * Inquirer.js
  * A collection of common interactive command line user interfaces.
--- a/lib/inquirer.js
+++ b/lib/inquirer.js
@@ -1,5 +1,3 @@
-require('time-require');
-
 /**
  * Inquirer.js
  * A collection of common interactive command line user interfaces.
--- a/lib/objects/choice.js
+++ b/lib/objects/choice.js
@@ -1,5 +1,5 @@
 'use strict';
-var utils = require('../utils/');
+var utils = require("../utils/");
 
 /**
  * Choice object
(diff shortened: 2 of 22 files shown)


jonschlinkert:
as a sidenote, to make inquirer initialize faster during actual use, something it might be worth considering using something like https://www.npmjs.com/package/kind-of for checking native types instead of lodash. Not because the type-checking would be better or faster, but because in some cases that is the only thing lodash is used for, which means the entire lodash module might be loaded up, adding 40-70ms to initto, just do `Array.isArray()`. if you want i can do this too

SBoudrias:
Hey, I won't be merging this.

The main reason is ES6 modules don't support conditionals/nested loading. So this is not future friendly.

Also, having God object with a bunch of stuff on it are hard to maintains on the long term.

The commonjs module system is the standard in the Node environment, we'll stick to it for now.

Thanks for sending this PR out though.

jonschlinkert:
Silliness. It's trivial to swap out that file in a build step. Of course, that's your call, but you should have said that before saying you wanted the pr. It's a shame too, since this is by far the least performant library we used in node.js.

SBoudrias:
@jonschlinkert this was a misunderstanding on the goal. The issue was about lazily evaluating the terminal environment to prevent crashes on load in non-tty environment; not about using lazy-require.

PR #309, opened 2015-12-02
Add suggestion for Vorpal.js as a CLI utility

dthree:
I noticed you were kind enough to add a plug for Commander in your readme. 

[Vorpal.js](https://github.com/dthree/vorpal) is rapidly gaining traction thanks to awesome [things like this](http://developer.telerik.com/featured/creating-node-js-command-line-utilities-improve-workflow/), and a lot of people are now apparently finding it really useful, from their feedback.

I'm trying to spread the word around and thought it'd be awesome if your page could suggest Vorpal as well as Commander. You can totally say no, but I thought I'd at least suggest it. :smiley: 

As Vorpal both uses Inquirer and is inspired by Commander, I take definite strides to promote both of those projects wherever I can as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ A collection of common interactive command line user interfaces.
 - *validating* answers
 - managing *hierarchical prompts*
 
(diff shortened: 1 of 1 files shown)


SBoudrias:
Oh that's simple enough. I'm happy to promoto vorpal here.

dthree:
Thanks! You're awesome! :fireworks:

PR #311, opened 2015-12-10
add directory type prompt for choosing a relative folder

nicksrandall:
I have taken the "list" type prompt and converted it to be a recurive "directory" selector. 

## Use Case
Many scoffolding tools (yeoman, plop, slush, ect.) internally rely on Inquirer as a method of getting user input before execuiting certain commands. This new prompt type will allow these tools to ask questions like "Where you would like to place this new component?" and allow their uses to select a folder (relative to a given path). 

## Limitations
Currently, this only works for selecting directories. However, it could easily be cloned to work for files.

## Issues
Testing. I need some help with the unit tests. My current test is passing but it is basically testing the file structure of this repo so that seems like a nightmare to maintain. I'm hopeing somebody has an idea on a better way to test this functionality.

### DIFF
--- a/examples/directory.js
+++ b/examples/directory.js
@@ -1,0 +1,17 @@
+/**
+ * List prompt example
+ */
+
+"use strict";
+var inquirer = require("../lib/inquirer");
+
+inquirer.prompt([
+  {
+    type: "directory",
+    name: "path",
+    message: "In what directory would like to perform this action?",
+    basePath: "./"
+  }
+], function( answers ) {
+    console.log( JSON.stringify(answers, null, "  ") );
+  });
--- a/lib/inquirer.js
+++ b/lib/inquirer.js
@@ -48,6 +48,7 @@ inquirer.createPromptModule = function (opt) {
 
   promptModule.restoreDefaultPrompts = function () {
     this.registerPrompt('list', require('./prompts/list'));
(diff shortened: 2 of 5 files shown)


SBoudrias:
Hey, this can be an inquirer plugin. There's no need to add it to the core.

nicksrandall:
Ahh, looks like travis didn't like my unit tests. I agree that this could be a plugin, but I was thinking that this was a common enough use case that it could be added into core so that individual users didn't have to seperatly implement it.

SBoudrias:
I'm planning to eventually move every prompt in the core to plugins. Yours would really fit better as a plugin.

nicksrandall:
Ok. It was worth a try!

Signus:
@SBoudrias, do you plan to implement a plugin guide for instances like this? I see Nick simply implemented a new Prompt prototype, which is the convention I'm following for this very same problem.

SBoudrias:
@Signus that's the way to go currently. I eventually want to split the prompt core and the inquirer system in different modules to ease the creation of plugin. I just need more time to work on that specific feature.

I'm not sure about creating a guide right away as I feel the API is pretty sketchy and I don't want to bind people to it. - if that makes sense.

(also closing as this initial issue is resolved with a module)

PR #316, opened 2015-12-30
Fix offset separator

fredericgrati:
Fix bug when a list prompt has separator and is longer than pageSize (using paginator)

The offset of the separator was added to the cursor position but the index of the first line to print was wrong.

### DIFF
--- a/lib/prompts/checkbox.js
+++ b/lib/prompts/checkbox.js
@@ -96,7 +96,8 @@ Prompt.prototype.render = function (error) {
     message += chalk.cyan( this.selection.join(", ") );
   } else {
     var choicesStr = renderChoices(this.opt.choices, this.pointer);
-    message += "\n" + this.paginator.paginate(choicesStr, this.pointer, this.opt.pageSize);
+    var indexPosition = this.opt.choices.indexOf(this.opt.choices.getChoice(this.pointer));
+    message += "\n" + this.paginator.paginate(choicesStr, indexPosition, this.opt.pageSize);
   }
 
   if (error) {
@@ -168,7 +169,6 @@ Prompt.prototype.toggleChoice = function( index ) {
   var checked = this.opt.choices.getChoice(index).checked;
   this.opt.choices.getChoice(index).checked = !checked;
 };
-
 
 /**
  * Function for rendering checkbox choices
--- a/examples/checkbox.js
+++ b/examples/checkbox.js
@@ -11,23 +11,40 @@ inquirer.prompt([
     message: "Select toppings",
     name: "toppings",
     choices: [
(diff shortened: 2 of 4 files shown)


SBoudrias:
Can you explain clearly how to reproduce the bug you're having?

fredericgrati:
Try to have a long checkbox list with separator. 
You will see that the cursor will disappear.

I have modified the examples/checkbox.js in order to have a long list. try it with the master version ;)

lib/prompts/checkbox.js
```diff
@@ -96,7 +96,8 @@ Prompt.prototype.render = function (error) {
     message += chalk.cyan( this.selection.join(", ") );
   } else {
     var choicesStr = renderChoices(this.opt.choices, this.pointer);
-    message += "\n" + this.paginator.paginate(choicesStr, this.pointer, this.opt.pageSize);
+    message += "\n" + this.paginator.paginate(choicesStr, this.pointer,
+     this.opt.pageSize, this.getSeparatorOffset(this.pointer));
```
SBoudrias: We should pass both the pointer and the offset. A single value (pointer + offset) should work just as fine in this case.

SBoudrias: Also, I feel like the correct position really just is `this.opt.choices.choices.indexOf(activeChoice)`. As this will position the choice item in the full array (with separators).

fredericgrati:
You are right, I have made the modifications :)

lib/prompts/checkbox.js
```diff
@@ -96,7 +96,8 @@ Prompt.prototype.render = function (error) {
     message += chalk.cyan( this.selection.join(", ") );
   } else {
     var choicesStr = renderChoices(this.opt.choices, this.pointer);
-    message += "\n" + this.paginator.paginate(choicesStr, this.pointer, this.opt.pageSize);
+    var indexPosition = this.opt.choices.choices.indexOf(this.opt.choices.getChoice(this.pointer));
```
SBoudrias: Can you add an `indexOf` function on the `Choices` collection object rather than accessing inner properties of the object (`choices.indexOf` rather than `choices.choices.indexOf`). This is already done for other `Array.prototype` methods here: https://github.com/SBoudrias/Inquirer.js/blob/master/lib/objects/choices.js#L98

SBoudrias:
One last comment and the rest looks good (and way simpler!)

Once you've fixed the last thing, please squash your commits and we'll be good to merge.

fredericgrati:
Thank you for your help on this PR. ;)
Can you update the package.json version after the merge ? thx

SBoudrias:
Awesome! Happy new year :)

PR #322, opened 2016-01-16
fix incorrect cursor position

yibn2008:
I found two bugs when use input prompt.

## First One

When prompt message contains multibyte characters (e.g. Chinese/Japanese), the input cursor position will move away:

qq20160116-0

The reason is result of `width()` function in screen-manager.js doesn't match with display length.

## The Second

When I moved LEFT key after typed some text in answer area, the cursor does't move; but when I type again, the characters will come out from the front of cursor.

qq20160116-1

When re-adjust cursor one the right in ScreenManager.render, the column size does't match with real cursor in mute stream

### DIFF
--- a/lib/utils/screen-manager.js
+++ b/lib/utils/screen-manager.js
@@ -3,16 +3,13 @@ var _ = require('lodash');
 var util = require('./readline');
 var cliWidth = require('cli-width');
 var stripAnsi = require('strip-ansi');
+var stringWidth = require('string-width');
 
 // Prevent crashes on environments where the width can't be properly detected
 cliWidth.defaultWidth = 80;
 
 function height(content) {
   return content.split('\n').length;
-}
-
-function width(content) {
-  return stripAnsi(content).length;
 }
 
 function lastLine(content) {
@@ -35,8 +32,6 @@ var ScreenManager = module.exports = function (rl) {
 };
 
 ScreenManager.prototype.render = function (content, bottomContent) {
-  var cursorPos = this.rl._getCursorPos();
-
   this.rl.output.unmute();
   this.clean(this.extraLinesUnderPrompt);
 
@@ -55,6 +50,9 @@ ScreenManager.prototype.render = function (content, bottomContent) {
     prompt = prompt.slice(0, -this.rl.line.length);
   }
   this.rl.setPrompt(prompt);
+
+  // setPrompt will change cursor position, now we can get correct value
+  var cursorPos = this.rl._getCursorPos();
 
   content = forceLineReturn(content);
   if (bottomContent) {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "inquirer",
-  "version": "0.11.2",
+  "version": "0.11.3",
```
SBoudrias: Please don't bump versions yourself. The maintainers will take care of that.

lib/utils/screen-manager.js
```diff
@@ -12,7 +14,19 @@ function height(content) {
 }
 
 function width(content) {
-  return stripAnsi(content).length;
+  var chars = stripAnsi(content);
+  var len = 0;
+
+  for (var i = chars.length - 1; i >= 0; i--) {
```
SBoudrias: Let's drop the custom code in favor a module doing what we want - in this case https://www.npmjs.com/package/string-length will do both astral points and ANSI.

SBoudrias:
Thanks for bringing that bug up and offering a fix! That's awesome :)

lib/utils/screen-manager.js
yibn2008: @SBoudrias I'm afraid that string-length cannot be used in this case, for example, the `stringLength("哈")` is 1, but what we want is 2, the real width displayed on terminal. Actually the code is reference to the _getDisplayPos in https://github.com/SBoudrias/readline2/blob/master/index.js#L67-L95.

Maybe we should extract the width calculate logic from readline2 to a npm module.

yibn2008:
I found https://github.com/sindresorhus/string-width-cli can solve the CLI character width problem, I will fix with it later.

lib/utils/screen-manager.js
```diff
@@ -3,6 +3,9 @@ var _ = require('lodash');
 var util = require('./readline');
 var cliWidth = require('cli-width');
 var stripAnsi = require('strip-ansi');
+var codePointAt = require('code-point-at');
```
SBoudrias: We don't use these two modules anymore. You can remove them.

SBoudrias:
Okay, code LGTM. Please remove the extra dependencies you kept and squash your commits. We'll be good to merge after.

yibn2008:
@SBoudrias Now the extra dependencies has removed.

SBoudrias:
Last thing, can you squash your commits into a single one? See https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/

yibn2008:
All commits has squashed to a single one.

SBoudrias:
Awesome, thanks a ton!

PR #326, opened 2016-01-25
add sanity check for screen width

bcoe:
string-width can return a 0 width screen, which causes inquirer to blow up:

https://gist.github.com/bcoe/6bc16d0dbe5190b8d302

I'm submitted a [pull request](https://github.com/knownasilya/cli-width/pull/6) to `cli-width`, but I felt it might be worthwhile to defend against this in inquirer as well.

### DIFF
--- a/lib/utils/screen-manager.js
+++ b/lib/utils/screen-manager.js
@@ -17,10 +17,11 @@ function lastLine(content) {
 }
 
 function normalizedCliWidth() {
+  var width = cliWidth() || cliWidth.defaultWidth;
   if (process.platform === 'win32') {
-    return cliWidth() - 1;
+    return width - 1;
   }
-  return cliWidth();
+  return width;
 }
 
 var ScreenManager = module.exports = function (rl) {


SBoudrias:
We set a default value at the top of this file.

bcoe:
@SBoudrias yeah, but unfortunately as of right now `cli-width` only respects the default value if `process.stdout.getWindowSize` and `tty.getWindowSize()` are not available

In the case of Alpine running on Docker, `process.stdout.getWindowSize()` returns a width of `0` -- which `cli-width` in turn happily returns; I think better behavior might be to return the default value if `process.stdout.getWindowSize()` returns anything that's not truthy.

SBoudrias:
Alright, sounds good then!

bcoe:
@SBoudrias thanks for patching this so fast \o/ now I have no reason to procrastinate from releasing the next version of npm On-Site.

Mind pinging me when you release the patch to npm?

SBoudrias:
@bcoe done!

PR #327, opened 2016-01-25
add test for #326

bcoe:
Hey @SBoudrias maybe worth adding a test?


bcoe:
@SBoudrias in the very least, you can try commenting out `|| cliWidth.defaultWidth;` and get a feel for the error.

SBoudrias:
Ah, thanks for that

PR #328, opened 2016-01-27
Add Emacs up/down keybindings

dougluce:
FYI: the README specifies "wip" as the branch to target for feature adds. I'm guessing that's outdated.


SBoudrias:
Yeah, need to update that part of the readme.

Thanks for the PR

PR #330, opened 2016-02-07
Use "short" property in expanding prompts

eush77:
Choice objects may have a `"short"` property which is defined as a text displayed after selection. However, it doesn't work for all prompt types.

This patch makes `"expand"` work properly with `"short"` properties, that is, prefer it to `"name"` after selection.


SBoudrias:
Thanks

Issue #285, opened 2015-09-14
stdin/stdout cli-width issue


ruyadorno:
One small issue I found while playing with the possibility of setting an `stdin`/`stdout` introduced on #280 is that the [cli-width](https://github.com/knownasilya/cli-width) dependency is still using `process.stdout` directly.

In my specific case I could get a workaround by using its `defaultWidth` property but it's definitely going to be problematic for people wanting to define their own input/output streams.

SBoudrias:
You should probably open an issue on cli-width package.

PR #334, opened 2016-02-09
Updates cli-width dependency

ruyadorno:
Greetings my friend @SBoudrias,

**cli-width** finally got a major release with a nicer API to define `defaultWidth` AND the possibility of configure an option **stdout** as described on #285 

This PR updates the dependency to its latest version `v2.0.0` - it's also worth noting that this version also contains a fix to the annoying issue where the default value wouldn't be used, reported and patched on #326

Unfortunately, I had to do a bit of moving around to be able to get the value of `this.rl.output` - please let me know if that's good enough.

Thanks :zap:

### DIFF
--- a/lib/utils/screen-manager.js
+++ b/lib/utils/screen-manager.js
@@ -5,23 +5,12 @@ var cliWidth = require('cli-width');
 var stripAnsi = require('strip-ansi');
 var stringWidth = require('string-width');
 
-// Prevent crashes on environments where the width can't be properly detected
-cliWidth.defaultWidth = 80;
-
 function height(content) {
   return content.split('\n').length;
 }
 
 function lastLine(content) {
   return _.last(content.split('\n'));
-}
-
-function normalizedCliWidth() {
-  var width = cliWidth() || cliWidth.defaultWidth;
-  if (process.platform === 'win32') {
-    return width - 1;
-  }
-  return width;
 }
 
 var ScreenManager = module.exports = function (rl) {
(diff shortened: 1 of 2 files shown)


lib/utils/screen-manager.js
```diff
@@ -108,10 +97,21 @@ ScreenManager.prototype.done = function () {
...
+    defaultWidth: 80,
+    output: this.rl.output
+  });
+  if (process.platform === 'win32') {
+    return width - 1;
+  }
+  return width;
+};
+
+ScreenManager.prototype.breakLines = function (lines) {
```
SBoudrias: How about we pass the width as a parameter instead of making those class method?

ruyadorno: do you mean the output or really just the returning value of `cliWidth()`? :confused: anyways `this.rl.output` is the reason I had to turn them into class methods

The first thing I tried was sending the output as a parameter but then it cascades to `breakLines()` and `forceLineReturn()` having to add an extra parameter to them just to send down `this.rl.output` - it was a much uglier solution to be honest

I'm pretty sure it's going to look the same thing if we were going to do the same thing with the `width`

SBoudrias: I mean `breakLines(text, width);`

ruyadorno: oh yeah :stuck_out_tongue: only for `breakLines` and `forceLineReturn` might be better, let me try in a sec

ruyadorno: what about it now?

lib/utils/screen-manager.js
```diff
@@ -55,14 +44,14 @@ ScreenManager.prototype.render = function (content, bottomContent) {
   // setPrompt will change cursor position, now we can get correct value
   var cursorPos = this.rl._getCursorPos();
 
-  content = forceLineReturn(content);
+  content = forceLineReturn(content, this.normalizedCliWidth());
```
SBoudrias: Let's calculated the width value only once at the beginning of the function and reuse it elsewhere.

ruyadorno: yeah, I just thought about it too, makes sense :+1:

ruyadorno: oh yeah, much much better :trophy:

SBoudrias:
Awesome thanks!!

ruyadorno:
:confetti_ball: thank you!

SBoudrias:
0.12 is out with this fix in

Issue #275, opened 2015-08-25
Use lodash modules


Downchuck:
Lodash comes in modular forms, which may be more appropriate for the light use in this library:
https://www.npmjs.com/browse/keyword/lodash-modularized

Currently lodash is 2.1M of the total 2.9M that inquirer.js distributes as when unpacked.

SBoudrias:
Feel free to send a PR

sindresorhus:
:+1:

seethemhigh:
Allow me to chime in. I've got the PR ready for the submission. Shall I send it out? @Downchuck does not mind?

Downchuck:
That's fine with me. We did find in another project, that if you include many lodash modules, the aggregate size of the license files is unreasonable.

seethemhigh:
Thanks!

That's amusing. Same thing happens here. That's only increased the size of the package. Not worth it. I think the issue can be closed.

dthree:
@SBoudrias this is a really big deal for me. 

As you know [Vorpal](https://github.com/dthree) is a big user of Inquirer. 

I have been trying to make Vorpal as slim as possible:

- Like Inquirer, it's a user-downloaded module. That means someone types `npm install` stares at terminal waiting for the library depending on Inquirer / Vorpal to download. Speed is of the essence.

- The slimmer the module, the more willing someone will be do include it as a dependency. 

Right now, the latest lodash version is taking 61% of Inquirer's total size. Vorpal has the same problem, also depending on lodash.

I've been thinking long and hard about how to handle this, and so I'm proposing a solution I hope you're okay with.

---

[I published a repo](https://www.npmjs.com/package/mindash) that mirrors the latest build of lodash and matches its semver. It's simplied the minified build of the core library. 

- That means it's `68KB` unzipped, instead of `1.07MB`. **Thats a big difference.**

- This reduces Inquirer's size from `1.71MB` unziped to `727KB`.

As you may have a concern on depending on something unofficial, know I've already published it in Vorpal and so have a vested interest in keeping it up to date. *I've also given you publishing rights to the module, so you don't have to depend on me if you need it updated right away.*

---

I've submitted a Pull Request that uses the minified module instead of lodash, with passing tests:

https://github.com/SBoudrias/Inquirer.js/pull/335

---

Does this sound okay to you? Thanks.

jdalton:
[...]

PR #335, opened 2016-02-20
minify lodash

dthree:
This updates Inquirer's lodash dependency to a minified version, reducing the footprint of Inquirer by nearly `1MB`.

See issue 275:

https://github.com/SBoudrias/Inquirer.js/issues/275


TrySound:
I don't think it's a good idea. Better split lodash into atomic functions.

PR #336, opened 2016-02-20
Split lodash

### LINKED ISSUE #275: Use lodash modules

TrySound:
Fix #275


lib/prompts/base.js
```diff
@@ -15,13 +17,13 @@ var ScreenManager = require('../utils/screen-manager');
...
   // Setup instance defaults property
-  _.assign(this, {
+  assign(this, {
     answers: answers,
     status : 'pending'
   });
 
   // Set defaults prompt options
-  this.opt = _.defaults(_.clone(question), {
+  this.opt = defaults(clone(question), {
```
TrySound: I think here we should reduce something

package.json
```diff
@@ -27,7 +27,16 @@
...
+    "lodash.assign": "^4.0.3",
+    "lodash.clone": "^4.3.0",
+    "lodash.defaults": "^4.0.1",
+    "lodash.filter": "^4.2.0",
+    "lodash.flatten": "^4.1.0",
+    "lodash.isarray": "^4.0.0",
+    "lodash.isplainobject": "^4.0.2",
+    "lodash.last": "^3.0.0",
+    "lodash.map": "^4.2.0",
+    "lodash.uniq": "^4.2.0",
```
dthree: Did you add up the combined package size of each of these modules? The problem as stated earlier is that, inflated with licenses, markdown, etc. these come out to the same size as lodash as a whole.

TrySound: But we don't need to update whole lodash only because one function is changed. Common packages via npm3 looks pretty good. Better than large bundle. If we will find solution in reducing assign, clone and defaults download size will be smaller.

TrySound:
@SBoudrias Any thoughts?

TrySound:
We already did it in gulp

SBoudrias:
We're using lodash pretty extensively in this package. I don't see reason to split in multiple package.

Especially, given how npm cache modules and how network requests goes. It's probably faster to load one version of lodash rather than 15 others modules. See the discussion on the issue this PR is linked to.

PR #338, opened 2016-03-04
Move cursor to end (of bottom content) on forced close

vweevers:
If a (custom) prompt renders some bottomContent (e.g. `this.screen.render(msg, 'bottom')`), but the user aborts, the cursor should be moved to the end of the last active screen.

A gif to show the problem and the fix (after `npm link inquirer`):

inquirer-cursor-fix-cut

(this is running [my fork](https://github.com/vweevers/inquirer-autocomplete-prompt/tree/inquirer-0.11) of [inquirer-autocomplete-prompt](https://github.com/mokkabonna/inquirer-autocomplete-prompt), which renders the autocomplete list as bottomContent so that the cursor stays at the normal text input above it. And in both instances, I'm pressing `a` to trigger the search, then `ctrl-c`.)

### DIFF
--- a/lib/ui/prompt.js
+++ b/lib/ui/prompt.js
@@ -76,12 +76,10 @@ PromptUI.prototype.processQuestion = function (question) {
 };
 
 PromptUI.prototype.fetchAnswer = function (question) {
+  var screen = this.createScreen();
   var Prompt = this.prompts[question.type];
-  var prompt = new Prompt(question, this.rl, this.answers);
+  var prompt = new Prompt(question, screen, this.rl, this.answers);
   var answers = this.answers;
-
-  // Remember the screen for when there's a forced close
-  this.setCurrentScreen(prompt.screen);
 
   return utils.createObservableFromAsync(function () {
     var done = this.async();
--- a/lib/ui/prompt.js
+++ b/lib/ui/prompt.js
@@ -79,9 +79,14 @@ PromptUI.prototype.fetchAnswer = function (question) {
   var Prompt = this.prompts[question.type];
   var prompt = new Prompt(question, this.rl, this.answers);
   var answers = this.answers;
+  var self = this;
+
   return utils.createObservableFromAsync(function () {
     var done = this.async();
+    self.activeScreen = prompt.screen;
+
     prompt.run(function (answer) {
+      self.activeScreen = null;
       answers[question.name] = answer;
       done({ name: question.name, answer: answer });
     });
--- a/lib/ui/prompt.js
+++ b/lib/ui/prompt.js
@@ -54,7 +54,7 @@ PromptUI.prototype.onCompletion = function () {
   this.close();
 
   if (_.isFunction(this.completed)) {
-    this.completed(this.answers);
+    this.completed(null, this.answers);
   }
 };
 
--- a/lib/ui/prompt.js
+++ b/lib/ui/prompt.js
@@ -83,10 +83,10 @@ PromptUI.prototype.fetchAnswer = function (question) {
 
   return utils.createObservableFromAsync(function () {
     var done = this.async();
-    self.activeScreen = prompt.screen;
+    self.setActiveScreen(prompt.screen);
 
     prompt.run(function (answer) {
-      self.activeScreen = null;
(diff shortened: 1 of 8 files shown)


SBoudrias:
PR welcomed.

lib/utils/screen-manager.js
```diff
@@ -19,6 +19,7 @@ var ScreenManager = module.exports = function (rl) {
   this.extraLinesUnderPrompt = 0;
 
   this.rl = rl;
+  if (rl) rl.activeScreen = this;
```
SBoudrias: Okay, so if we do this, we need clear relationship. Modifying an object outside your own scope is pretty bad practice and it'll make it hard to debug in the future.

vweevers: Gotcha. Though it's difficult to do this cleanly with the current class hierarchy.

vweevers:
How about this?

vweevers:
OK, ready for review. The base UI's `close()` now always calls `screen.releaseCursor()`, and `screen.done()` clears the screen's internal state. So in a normal close, `screen.releaseCursor()` is effectively a noop.

Maybe even better would be to have the base UI instantiate and remember the screen manager, then from PromptUI, pass it to the Prompt:

```js
var screen = this.createScreen();
var prompt = new Prompt(question, screen, this.rl, this.answers);
var answers = this.answers;

return utils.createObservableFromAsync(function () {
  var done = this.async();
  prompt.run(function (answer) {
    answers[question.name] = answer;
    done({ name: question.name, answer: answer });
  });
});
```

However, this would be a breaking change to the Prompt constructor, and break a lot of tests, so I decided against it.

SBoudrias:
Maybe we should instantiate the screen manager inside the UI class. Then we don't have to reach for properties on children objects.

vweevers:
> Maybe we should instantiate the screen manager inside the UI class. Then we don't have to reach for properties on children objects.

Like I said:

> Maybe even better would be to have the base UI instantiate the screen manager [..]. However, this would be a breaking change to the Prompt constructor, and break a lot of tests

If you're OK with that, I'll go ahead

SBoudrias:
Oh yeah, sorry missed that part. Let's go ahead with that!

lib/ui/prompt.js
```diff
@@ -76,9 +76,11 @@ PromptUI.prototype.processQuestion = function (question) {
 };
 
 PromptUI.prototype.fetchAnswer = function (question) {
+  var screen = this.createScreen();
   var Prompt = this.prompts[question.type];
-  var prompt = new Prompt(question, this.rl, this.answers);
+  var prompt = new Prompt(question, screen, this.rl, this.answers);
```
vweevers: @SBoudrias Before I start updating the tests, I wanna make sure we're on the same page. Happy with this?

SBoudrias: This approach is way better in my opinion, but I think we can do something better:

Would it makes sense for the BaseUI to call a method (`forceClose` or `cleanup`) on the Prompt instance. Then the prompt release the cursor in screen manager. That approach would really mean we don't pass or reference the screen manager in objects that don't own it.

vweevers: By Prompt, do you mean PromptUI or Prompt? Either way, this would only make sense if we go all the way and have the screen manager do all the cleanup (unmuting and ending output etc). Because now, the order of things is quite specific. And if we were to replace the `releaseCursor()` call below with a `cleanup()` call, it's not exactly clear what the responsibilities of that `cleanup` method are.

```js
UI.prototype.close = function() {
  // ..
  this.rl.output.unmute();

  if (this.currentScreen) {
    this.currentScreen.releaseCursor();
  }

  this.rl.output.end();
  this.rl.pause();
  this.rl.close();
};
```

SBoudrias:
Hey 1.0 just landed. Mind rebasing on it and then I'll take a look at this PR code.

mokkabonna:
Would be awesome if this got implemented so I can change the autocomplete prompt to use this.

vweevers:
@mokkabonna feel free to take over (tbh, this is a low priority for me, might never get around to it)

PR #339, opened 2016-03-08
Expose UI

vweevers:
In another module that wraps Inquirer, I rely on `inquirer.prompt()` to return the UI. But a recent change (68c46073e58afd8970bbe455db7b8f7e7de6584c) adding Promise support broke it.

This PR makes the UI object publicly accessible just like `rl` and `process`, that is, by monkey patching the promise. Out of curiosity, did you have any plans to refactor that?

Additionally, I renamed all the `ui` variables in the tests to `promise`, to reflect the aforementioned change.

### DIFF
--- a/lib/inquirer.js
+++ b/lib/inquirer.js
@@ -47,6 +47,7 @@ inquirer.createPromptModule = function (opt) {
     // promise object so they remain publicly accessible.
     promise.rl = ui.rl;
     promise.process = ui.process;
+    promise.ui = ui;
 
     return promise;
   };
--- a/lib/inquirer.js
+++ b/lib/inquirer.js
@@ -43,10 +43,8 @@ inquirer.createPromptModule = function (opt) {
       });
     });
 
-    // Monkey patch the readline and Observable `process` object on the
-    // promise object so they remain publicly accessible.
-    promise.rl = ui.rl;
-    promise.process = ui.process;
+    // Monkey patch the UI on the promise object so
+    // that it remains publicly accessible.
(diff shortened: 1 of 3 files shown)


lib/inquirer.js
```diff
@@ -47,6 +47,7 @@ inquirer.createPromptModule = function (opt) {
     // promise object so they remain publicly accessible.
     promise.rl = ui.rl;
     promise.process = ui.process;
+    promise.ui = ui;
```
SBoudrias: Would it be too cumbersome to remove `promise.rl` and reference it as `promise.ui.rl` in the tests?

SBoudrias:
> Out of curiosity, did you have any plans to refactor that?

Yes and no. I want to keep the reactive interface hook exposed for now. For semi backward compatibility and just because some people really like the feature. And promises have been a major request for a while, so I felt it was important to ship it soon, especially as Promises are now native in Node.

I'm not sure how to conciliate all these needs for now. I don't like the monkey patching, but that was the easiest and simplest way to move forward without breaking most people code.

vweevers:
Updated to only expose `ui`. `rl` and `process` can be referenced as `promise.ui.x`.

Re promises: the downside of the current approach, besides the monkey patching, is that it swallows errors (including errors thrown in user land). People who use a regular callback, will not think to add a `catch` handler on the promise. Nor should they have to.

SBoudrias:
Good point. Please open an issue to discuss solutions and way forward further.

PR #370, opened 2016-05-12
Recommend args

leo:
I've recently spent some time on creating a smaller yet powerful [alternative](https://github.com/leo/args) to commander. If you ask me, it's a better way to build CLIs. If you think like that as well, I would be very happy if it could be added to the recommendations... :blush:

Oh. And I've also suggested inquirer in the description of course (when it comes to user input)!

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ A collection of common interactive command line user interfaces.
 - *validating* answers
 - managing *hierarchical prompts*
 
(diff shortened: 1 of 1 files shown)


SBoudrias:
Happy to add it. I like the interface.

leo:
@SBoudrias Yey! Thank you.

PR #372, opened 2016-05-20
Add missing options to question type description

sirreal:
The description of possible question `type` did not include

* `expand`
* `checkbox`

Add these options (which are present in the `type` detail section).


SBoudrias:
Thanks!

PR #376, opened 2016-05-29
fix #373

### LINKED ISSUE #373: Confusing exception when using expand, if the value does not match up with the key.
I had an existing prompt config:

```code
{
        name: 'isTeamGame',
        message: 'Does the game have teams?',
        type: 'expand',
        choices:  [
          {
            key: 'y',
            name: 'yes, the game has teams',
            value: true
          },
          {
            key: 'n',
            name: 'no, the game does not have teams',
            value: false
          }
        ]
 }
```

This worked fine, until I upgrade to the latest version of Inquirer.  When "n" was selected, I get the confusing error:

TypeError: Cannot read property 'short' of undefined
    at Prompt.onSubmit (/usr/local/lib/node_modules/yo/node_modules/yeoman-environment/node_modules/inquirer/lib/prompts/expand.js:164:23)

It's due to this line of code:
```javascript
  var choice = this.opt.choices.where({value: state.value})[0];
```

There should be a more descriptive error message, like:
"Error: value 'n' does not exist in list of choices (true, false)"

akatsukle:
This patch aims to fix [#373](https://github.com/SBoudrias/Inquirer.js/issues/373).
It checks whether the attribute `value` is really a string, and throw an error if needed.

Updated : it now allows the value to be a String, a Boolean or a Integer. (Nb, it would probably be good to also support Objects)

### DIFF
--- a/lib/prompts/expand.js
+++ b/lib/prompts/expand.js
@@ -193,17 +193,11 @@ Prompt.prototype.onKeypress = function () {
 
 Prompt.prototype.validateChoices = function (choices) {
   var formatError;
-  var typeError;
   var errors = [];
   var keymap = {};
   choices.filter(Separator.exclude).forEach(function (choice) {
     if (!choice.key || choice.key.length !== 1) {
       formatError = true;
-    }
-    if (choice.value !== undefined) {
-      if (typeof choice.value !== 'string' && typeof choice.value !== 'boolean' && typeof choice.value !== 'number') {
-        typeError = true;
-      }
     }
     if (keymap[choice.key]) {
       errors.push(choice.key);
@@ -214,9 +208,6 @@ Prompt.prototype.validateChoices = function (choices) {
 
   if (formatError) {
     throw new Error('Format error: `key` param must be a single letter and is required.');
-  }
-  if (typeError) {
-    throw new Error('Type error: `value` param must be a string, a number or a boolean.');
   }
   if (keymap.h) {
     throw new Error('Reserved key error: `key` param cannot be `h` - this value is reserved.');
--- a/lib/prompts/expand.js
+++ b/lib/prompts/expand.js
@@ -190,11 +190,17 @@ Prompt.prototype.onKeypress = function () {
 
 Prompt.prototype.validateChoices = function (choices) {
   var formatError;
+  var typeError;
   var errors = [];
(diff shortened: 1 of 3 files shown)


lib/prompts/expand.js
```diff
@@ -115,7 +115,10 @@ Prompt.prototype.getCurrentValue = function (input) {
     return null;
   }
 
-  return selected.value || selected.name;
+  if (selected.value === undefined) {
```
SBoudrias: Why change this line? The new code does the same thing in 4 lines instead of one.

SBoudrias:
How about we just check the result of the value lookup in the `where()` and if we get no value, then we throw an helpful error message. That'd be better than a bunch of type checkings that doesn't actually fix the issue.

lib/prompts/expand.js
SBoudrias: Re-reading the issue, this section here actually is the only bug.

It'll ignore falsy value. As so, that's actually the only fix we need.

akatsukle:
Yes the fix that actually distinct between false and falsy (like undefined) is actually enough for the bug.
However, how should values like `undefined` or `null` be treated?
`choices:  [ {key: 'n', name: 'no, the game does not have teams', value: false}, {key: 'y', name: 'yes, the game have teams', value: undefined}]` produces the same error than in #373 (TypeError: Cannot read property 'short' of undefined)
`choices:  [ {key: 'n', name: 'no, the game does not have teams', value: false}, {key: 'y', name: 'yes, the game have teams', value: null}]` does not allow me to select 'y'.
If the decision if to forbid the usage of these values, then it should be specified in the API, and the error should be handled correctly. 
Otherwise, they should be returned correctly.
Maybe I can treat these cases in another pull request?

akatsukle:
`Bad response:422 {"message":"Couldn't find a repository matching this job.","error":true}` did I broke something?
Sorry about the coverage decrease, I don't get which line does not get covered :/.

SBoudrias:
I ran some tests and fixed the bug in this commit https://github.com/SBoudrias/Inquirer.js/commit/8a49df659d54ada610c8437c89a02f3f1572ecd5

Thanks for sending this out!

akatsukle:
Thanks, awesome :+1:

PR #386, opened 2016-06-11
Convert JS to TS

luisvt:


SBoudrias:
Hi, thanks for your interest, but I have no plan to change the core language of inquirer.

PR #387, opened 2016-06-17
feat(example/hierarchical): Added an example of hierarchical flow

kaustavha:
Hi there,
This is an example of hierarchical flow using inquirer.
e.g. 
```
A - B1 - C1
       \
         B2 - C2
```
Its kind of trivial since deriving this from other examples is easy, but I was looking for something like that for a second.


SBoudrias:
Thanks!

PR #388, opened 2016-06-19
Fix `subscribe` example in README

GeorgeTaveras1231:
After some debugging, found that the `process` attribute is nested under `ui`


SBoudrias:
Oh yeah, missed that :(

Thanks for the PR!

Issue #99, opened 2014-03-01
Add support for multiline input


brianweil:
I would like to add support for multiple lines of text in the input type. I use the prompt for commit messages and sometimes one line is not enough. The UI could be as simple as pressing shift+enter to move to the next line of input. Feedback from the contributors is welcome and appreciated.

jameswomack:
This would be great for doing things like blog posts from the command-line. Hexo would do well to integrate this.

RonaldTreur:
Just to make sure this stays on the radar: +1

adjohnson916:
+1 Any updates on this? I support the `Shift`+`Enter` idea, either as a default to insert newline on any option, or swapped for `Enter` behavior which instead inserts newline on options explicitly configured as multiline.

SBoudrias:
@adjohnson916 PR welcomed.

SBoudrias:
This is now shipped with `v1.1.0`

0phoff:
I know there is editor support, but is there any chance that you will also support multiline?
I prefer would prefer multiline over editor for my application.

Multiline should allow the user to enter multiple lines, and close when it encounters an empty line (or line with only a dot) or when the user presses CTRL+D.
At least that is the standard behaviour across various command line applications I use.

I tried to look at your code to implement it myself, but I don't really understand how to implement a custom prompt... Is there any documentation or tutorial?

aleclarson:
I have a working implementation [here](https://github.com/aleclarson/inquirer/commit/3b5c14efa29be0efb4c713c40e9b7225822c9739). It adds a `multiline` option to `input` prompts.

I don't think it will merge cleanly into this repository, because I made other changes. But you could try if you're desperate.

KoStard:
Has this feature been merged? I don't see how to receive multi-line input from the user from terminal without opening external windows.

```
import { input } from '@inquirer/prompts';

const answer = await input({
  message: 'Enter a description',
  multiline: true,
});
console.log(answer);
```

I tried this, but it only accepted one line. Also I don't see anything about multiline support in the homepage: https://github.com/SBoudrias/Inquirer.js

I think this issue can be reopened to track the interest in this request.

SBoudrias:
Hi @KoStard, the multi-line comes through the `editor` prompt. There's been some more recent talk over here too https://github.com/SBoudrias/Inquirer.js/discussions/1583 (I'm open to improving this, but transparently, it's not on my todo list ATM. So it's up for the taking.)

PR #389, opened 2016-06-23
Implemented editor input support. (Originally multiline) 

mrkmg:
Uses env VISUAL, EDITOR, or notepad(windows) or vim(linux)

Addresses #99 

I was not sure if I should add the multiline option to the "Objects" -> "Question" section of the readme.

I also am not sure how to write a test for this, although I am very willing to (as the Contributing sections notes). Looking for input from @SBoudrias on how to implement a test.

### DIFF
--- a/test/specs/prompts/editor.js
+++ b/test/specs/prompts/editor.js
@@ -9,7 +9,7 @@ describe('`editor` prompt', function () {
   beforeEach(function () {
     this.previousVisual = process.env.VISUAL;
     // Writes the word "testing" to the file
-    process.env.VISUAL = 'node test/helpers/write.js testing';
+    process.env.VISUAL = 'node ./test/bin/write.js testing';
     this.fixture = _.clone(fixtures.editor);
     this.rl = new ReadlineStub();
   });
@@ -25,7 +25,7 @@ describe('`editor` prompt', function () {
     this.rl.emit('line', '');
 
     return promise.then(function (answer) {
-      expect(answer).to.equal('testing');
+      return expect(answer).to.equal('testing');
     });
   });
 });
--- a/test/specs/prompts/editor.js
+++ b/test/specs/prompts/editor.js
@@ -8,7 +8,8 @@ var Editor = require('../../../lib/prompts/editor');
 describe('`editor` prompt', function () {
   beforeEach(function () {
     this.previousVisual = process.env.VISUAL;
-    process.env.VISUAL = 'truncate --size 0';
+    // Writes the word "testing" to the file
+    process.env.VISUAL = 'node test/helpers/write.js testing';
     this.fixture = _.clone(fixtures.editor);
     this.rl = new ReadlineStub();
   });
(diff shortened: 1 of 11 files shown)


SBoudrias:
I think the multiline input should be it's own prompt. How it work is very different then how the input one is working.

mrkmg:
I am willing to do that! Do you have a preference on naming? Maybe add a type called "editor"?

SBoudrias:
@mrkmg oh, maybe just `multiline`?

mrkmg:
I think `editor` makes the most sense, since it's not really `multiline`, its a completely different paradigm. This feels more correct because `multiline` to me implies that it would be more like `input` but with multi-line support, where `editor` is more descriptive of what is actually happening.

mrkmg:
Also, I am going to omit the .png for the new type so it can be made on the same system with terminal colors as the rest of them (and be hosted in the same dropbox account).

SBoudrias:
@mrkmg no worries about the screenshots

mrkmg:
Please look at the test, as it is using the Linux command `truncate` to alter the tempFile to verify that the $VISUAL is actually getting called. It was the most simple way I could think to test that the VISUAL is actually getting called.

mrkmg:
It appears spawnSync is not available in node .10

Going to look into implementing with async spawn.

mrkmg:
There is a problem with using spawn. I implemented it, but about half of my keypresses are absorbed by the node script and not sent to the editor. Now using a polyfill for spawnSync. Should be good for review.

mrkmg:
@SBoudrias I have wrote a new package to handle the external editor by itself and then used that package in Inquirer. That way people can use just the editor if they want. This removes the complexity from Inquirer. The tests still verify that everything is working correctly. I also updated the output to not output the multiline output as it could be very large and take away from the beauty of Inquirer. If this should be changed back, let me know. 

Also, your very welcome for the PR. I am love Inquirer and use it on a couple projects.

test/specs/prompts/editor.js
```diff
@@ -0,0 +1,45 @@
...
+  });
+
+  afterEach(function () {
+    process.env.VISUAL = this.previousVisual;
+  });
+
+  it('should use raw value from the users editor', function (done) {
+    var input = new Editor(this.fixture, this.rl);
+
+    input.run().then(function (answer) {
```
SBoudrias: Can you return the promise instead of using done (like the second test).

test/specs/prompts/editor.js
```diff
@@ -0,0 +1,45 @@
...
+var _ = require('lodash');
+var ReadlineStub = require('../../helpers/readline');
+var fixtures = require('../../helpers/fixtures');
+
+var Editor = require('../../../lib/prompts/editor');
+
+describe('`editor` prompt', function () {
+  beforeEach(function () {
+    this.previousVisual = process.env.VISUAL;
+    process.env.VISUAL = 'truncate --size 0';
```
SBoudrias: Any things we could do to get some input back to the prompt? It'd be nice if we could `expect(answer).to.equal('dummy answer');`.

The empty string is not super insightful in this case :(

SBoudrias:
The feature LGTM. Some minor reviews on the testing itself.

mrkmg:
Something weird with the test. Trying to figure out whats going on...

SBoudrias:
Awesome!!

Issue #264, opened 2015-07-30
Prompt.prototype.prefix is not used in getQuestion


juancoen:
In this line (https://github.com/SBoudrias/Inquirer.js/blob/v0.9.0/lib/prompts/base.js#L169), the call to prefix was changed with:

```
chalk.green('?')
```

In version 0.8.5, the prefix function was being used. Is there a reason for not using it anymore?

Thanks!

SBoudrias:
Yeah, I'm removing all the crust around the huge classes so it can easily be exported to a standalone module.

Not sure what I want to do about prefixes/suffixes yet.

juancoen:
it would be nice to make prefix and suffix configurable, can that be done in other way?

jsonmaur:
I agree with this, I don't like having the `?` suffix for each question asked, especially when it doesn't go along with my UI. It really should be configurable so I can remove it. @SBoudrias is this something that you'd take a PR for? I'd be happy to do it.

SBoudrias:
@jsonmaur Well, it depends on the solution implemented.

What do you have in mind?

jsonmaur:
@SBoudrias Maybe just a `messagePrefix` option on the question object that would default to `?`. Seems like you could just add it as a default at `lib/prompts/base.js` line 21, then just reference it in `Prompt.prototype.getQuestion`. But that's from me looking at the code for all of 5 minutes, haha.

That may get tedious to set the prefix on every question, so maybe a second parameter passed to `inquirer.prompt` could be global options (unless that's reserved for a callback). That could be useful for more than just setting the prefix, such as defining a global validators, etc. Thoughts?

jsonmaur:
Any update on this? My ideas above may be overkill, but I need to at least hide it. My current solution is really hacky (overriding the `getQuestion` prototype with my own function).

SBoudrias:
@jsonmaur maybe we could set it up as a constant `inquirer.PREFIX` or something similar. Then it could be replaced without too much trouble and without adding extra API surface to the prompt method.

jsonmaur:
That would work. With that in mind, it would also be nice to provide a way to customize prefixes for other things such as validation error messages, prompt arrows, checkboxes, etc. Maybe a list of several different constants, or having the `PREFIX` constant accept an object map?

SBoudrias:
@jsonmaur an object map would be better.

heylookltsme:
[...]

PR #390, opened 2016-06-27
Adds customizable checkbox formatting.

mbleigh:
I like Inquirer, but I'm not a fan of the radio button format for checkboxes. This PR adds a configurable `format` option to checkbox types that lets you specify your own formats.

I'd be happy to write a test for this but I couldn't figure out how to get it to print the prompt and then read the output.


SBoudrias:
See the discussion at https://github.com/SBoudrias/Inquirer.js/issues/264 for a better solution to the issue.

PR #403, opened 2016-08-07
Use individual lodash modules

dguettler:
As part of ember-cli/ember-cli#6149, this PR aims to reduce the number of files downloaded for this package.

Switch to individual lodash modules in production code to reduce dependencies.

note that lodash.extend is an alias for lodash.assignIn (see [here](https://lodash.com/docs#assignIn))

This will save 758 dep files in production


SBoudrias:
So instead if one cache lodash version, well be loading over 10 other dependencies? This is not fixing an actual issue.

dguettler:
@SBoudrias I'd like to ask you to reconsider. Yes it does add additional dependencies, however it reduces number of files, disk space and install time.
@stefanpenner might have some more details on this.

Based on some naive testing I get the following results
```
# With PR applied
time npm install --production

real	0m4.390s
user	0m2.084s
sys	0m0.688s

781	node_modules
25	lib
24	test
14	examples
4	assets
1	package.json
1	gulpfile.js
1	README.md
1	LICENSE-MIT
0	coverage

Size: 10 MB

# Without PR applied
time npm install --production

real	0m8.518s
user	0m2.289s
sys	0m0.863s

1739	node_modules
25	lib
24	test
14	examples
4	assets
1	package.json
1	gulpfile.js
1	README.md
1	LICENSE-MIT
0	coverage

Size: 14 MB
```

SBoudrias:
Lodash announced they were reducing size and dependencies on their next release. IMO that's where you should look if you want to help fix size issues.

PR #404, opened 2016-08-08
Update README.md

walfly:
@SBoudrias could be mistaken, but it doesn't look like 'ui' is a property of the return value from the prompt method.  It seems that the "process" property you want to subscribe to is a top-level property.


SBoudrias:
Thanks

Issue #363, opened 2016-04-21
Error handler callback


iongion:
I do not know if this is the expected behavior, but I am having trouble handling the errors, I want to display the user an input that if invalid, will display an error and retry otherwise continue, what I am doing is this("inquirer": "1.0.2"):

```
function asumeRole(cb) {
  const prompt = {
    name: 'MFA',
    type: 'input',
    message: 'Please enter the generated MFA code',
    validate: function validate(response) {
      const done = this.async();
      if (response.length === 0) {
        done('The MFA code cannot be empty');
      } else {
        done(null, response);
      }
    }
  };
  inquirer.prompt([prompt])
  .then((response) => {
    console.log('got response', response);
  })
  .catch((error) => {
    console.log('got error', error);
  });
}
gulp.task('deploy', [], (cb) => {
  asumeRole((mfaCode) => {
    console.log('using mfa code', mfaCode);
  });
});
```

The done callback with error is called but then the catch is never called rendering impossible to do something on error except then bailing out.

The output looks like this:

```
? Please enter the generated MFA code
W:\Workspace\MyProject\node_modules\rx\dist\rx.js:77
    throw e;
    ^
 The MFA code cannot be empty
```

SBoudrias:
Yes, failing validation doesn't fail the prompt. It warn the users and ask a valid answer.

zckrs:
Same issue. 

```
validate(input) {
        return new Promise((resolve, reject) => {
          console.log(input);
          reject('Wrong choices. You should select other choice!');
        });
      }
```

Output:
```
/home/zckrs/.nvm/versions/node/v6.2.0/lib/node_modules/yo/node_modules/rx/dist/rx.js:77
    throw e;
    ^
Wrong choices. You should select other choice!
                                                                                                                                                                                                                   
project7 git/master  
❯ 

```

How can we reject user choices without stop process ?

SBoudrias:
@zckrs so you're saying rejecting the promise end up throwing an unhandled error?

With every prompt? Only one prompt? Please give me more context + code example if possible.

zckrs:
Sorry for missing informations

I edit the first prompt in [examples/list.js](https://github.com/SBoudrias/Inquirer.js/blob/master/examples/list.js#L23) 

```
/**
 * List prompt example
 */

'use strict';
var inquirer = require('..');

inquirer.prompt([
  {
    type: 'list',
    name: 'theme',
    message: 'What do you want to do?',
    choices: [
[...]

PR #407, opened 2016-08-11
Add error handling for errors produced in filter function

jakepruitt:
Changes behavior of errors produced in async filter functions from
throwing to printing the error and reprompting the user for a
non-erroring input. Addresses #363

cc/ @rclark @SBoudrias @iongion


SBoudrias:
Thanks for sending this out :) Sorry for the delay to review.

PR #415, opened 2016-08-30
Ensure provided options aren't overwritten on successive prompt calls

### LINKED ISSUE #414: Calling a custom prompt module repeatedly fails
While building out some of the behavior for https://github.com/TryGhost/Ghost-CLI, I noticed some odd behavior regarding a custom prompt module.

In short, calling a custom prompt module multiple times breaks the internal output stream for the prompt.

Take this code:
```js
var inquirer = require('inquirer'),
    prompt = inquirer.createPromptModule({
        input: process.stdin,
        output: process.stdout
    });

prompt([{
    type: 'confirm',
    name: 'test',
    message: 'Testing 1'
}]).then(function (values) {
    console.log(values.test);

    return prompt([{
        type: 'confirm',
        name: 'test2',
        message: 'Testing 2'
    }]);
}).then(function (values) {
    console.log(values.test2);
});
```

When that is run, the first prompt ("Testing 1") will appear correctly. However, once the user has responded to the first prompt, the second prompt does not appear. The program does wait as if the the prompt is _supposed_ to appear, leading me to believe it's an issue with how the output stream is handled. Also to note is that typing `Y` or `n` when the second prompt should be there works - as the `console.log` in the second then block outputs the correct result (so it appears that stdin works just fine).

If the code above is switched from using the prompt to using `inquirer.prompt` it runs fine.

Will have a look at the code later on today or tomorrow hopefully and see if I can come up with a fix.

acburdine:
closes #414
- fix issue with custom prompt modules on successive calls

### DIFF
--- a/lib/ui/baseUI.js
+++ b/lib/ui/baseUI.js
@@ -54,14 +54,16 @@ function setupReadlineOptions(opt) {
   opt = opt || {};
 
   // Default `input` to stdin
-  opt.input = opt.input || process.stdin;
+  var input = opt.input || process.stdin;
 
   // Add mute capabilities to the output
   var ms = new MuteStream();
   ms.pipe(opt.output || process.stdout);
-  opt.output = ms;
+  var output = ms;
 
   return _.extend({
-    terminal: true
-  }, opt);
+    terminal: true,
+    input: input,
+    output: output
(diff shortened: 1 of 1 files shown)


SBoudrias:
Thanks!

PR #424, opened 2016-10-03
add support for overriding the "?" prefix for questions

starpit:
This pull request adds support for clients overriding the "?" prefix that appears on question prompts, by specifying a prefixMessage property.

### DIFF
--- a/lib/prompts/base.js
+++ b/lib/prompts/base.js
@@ -118,7 +118,7 @@ Prompt.prototype.handleSubmitEvents = function (submit) {
  */
 
 Prompt.prototype.getQuestion = function () {
-  var message = chalk.green('?') + ' ' + chalk.bold(this.opt.message) + ' ';
+  var message = (this.opt.hasOwnProperty('prefixMessage') ? this.opt.prefixMessage : chalk.green('?') + ' ') + chalk.bold(this.opt.message) + ' ';
 
   // Append the default if available, and if question isn't answered
   if (this.opt.default != null && this.status !== 'answered') {


lib/prompts/base.js
```diff
@@ -118,7 +118,7 @@ Prompt.prototype.handleSubmitEvents = function (submit) {
  */
 
 Prompt.prototype.getQuestion = function () {
-  var message = chalk.green('?') + ' ' + chalk.bold(this.opt.message) + ' ';
+  var message = (this.opt.hasOwnProperty('prefixMessage') ? this.opt.prefixMessage : chalk.green('?') + ' ') + chalk.bold(this.opt.message) + ' ';
```
SBoudrias: Would `(this.opt.prefixMessage || chalk.green('?') + ' ')` work just as fine?

starpit:
hi, thanks for the quick feedback on my pull request!

we could go a number of ways here. i wanted to support `prefixMessage: ''`, which javascript, in its glory, evaluates to a non-truthy condition. this is why i added the hasOwnProperty.

SBoudrias:
Hey, sorry for letting this linger. I don't really like the solution as this mean it needs to be provided for each questions.

Is there any use case where that's what we want rather than a "global" option applied everywhere?

starpit:
howdy, no worries. we do have a release coming up soon, but i can always point to my fork, for now.

this is a good suggestion, to make "?" a global preference. however, as far i can tell, each prompt impl has no reference back to the enclosing PromptModule (which would seem to be the natural home for a global preference).

we could change `promptModule.registerPrompt` so that it calls `prompt.globalOptions = opt`, and then prompts/base.js could have `this._opt = _.defaults(_.clone(question), globalOptions, { ... defaults ... }`

starpit:
ok yeah, that more or less would work. we'd have to update ui/prompt.js to preserve `this.opt = opt`, and update prompts/base.js to accept a globalOptions fourth parameter...

SBoudrias:
The ideal API would be:

```js
inquirer.setOptions({prefix: '>'});
```

This would be unique per inquirer instance.

Skywalker13:
> This pull request adds support for clients overriding the "?" prefix that appears on question prompts, by specifying a prefixMessage property.

I think that you should provide a way to remove completely the prompt, not just a way for changing.

SBoudrias:
@Skywalker13 you'd just pass an empty string for remove it.

SBoudrias:
Closing this PR as inactive and I don't plan to merge it as is.

If someone wants to see this merge, they can open a new PR with the code review concern addressed.

PR #425, opened 2016-10-04
Add support for object paths in question names

jviotti:
Consider the following set of questions:

``` js
{
  type: 'input',
  name: 'name.first',
  message: 'What\'s your first name?'
},
{
  type: 'input',
  name: 'name.last',
  message: 'What\'s your last name?'
}
```

The resulting answer object will look something like this:

``` js
{
  'name.first': 'John',
  'name.last': 'Doe'
}
```

If the `name` properties contains periods, it makes sense that Inquirer
will interpret them as paths in the final answers object, resulting in
something like this instead:

``` js
{
  name: {
    first: 'John',
    last: 'Doe'
  }
}
```

This change is implemented by making use of `_.set` when assigning the
value to the answers hash.

Signed-off-by: Juan Cruz Viotti jviotti@openmailbox.org

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ A question object is a `hash` containing question related values:
 
 - **type**: (String) Type of the prompt. Defaults: `input` - Possible values: `input`, `confirm`,
 `list`, `rawlist`, `expand`, `checkbox`, `password`, `editor`
-- **name**: (String) The name to use when storing the answer in the answers hash.
+- **name**: (String) The name to use when storing the answer in the answers hash. If the name contains periods, it will define a path in the answers hash.
 - **message**: (String|Function) The question to print. If defined as a function, the first parameter will be the current inquirer session answers.
 - **default**: (String|Number|Array|Function) Default value(s) to use if nothing is entered, or a function that returns the default value(s). If defined as a function, the first parameter will be the current inquirer session answers.
 - **choices**: (Array|Function) Choices array or a function returning a choices array. If defined as a function, the first parameter will be the current inquirer session answers.
--- a/lib/ui/prompt.js
+++ b/lib/ui/prompt.js
@@ -40,7 +40,7 @@ PromptUI.prototype.run = function (questions) {
 
   return this.process
     .reduce(function (answers, answer) {
-      this.answers[answer.name] = answer.answer;
+      _.set(this.answers, answer.name, answer.answer);
       return this.answers;
     }.bind(this), {})
     .toPromise(Promise)
--- a/test/specs/inquirer.js
+++ b/test/specs/inquirer.js
@@ -72,6 +72,33 @@ describe('inquirer.prompt', function () {
     });
   });
 
+  it('should take a prompts array with nested names', function () {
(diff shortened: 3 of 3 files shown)


jviotti:
If backwards compatibility with applications that may be using periods in their
names is a concern, I propose allowing the `name` property to be an array of
string, in which case its interpreted as a path:

```js
{
  type: 'input',
  name: [ 'name', 'first' ],
  message: 'What\'s your first name?'
},
{
  type: 'input',
  name: [ 'name', 'last' ],
  message: 'What\'s your last name?'
}
```

SBoudrias:
Do you have any specific use case where this is really desirable?

I see how this can be a syntax sugar, but I don't see why or where this would be necessary.

jviotti:
@SBoudrias, Its just a matter of being able to better organize the resulting answers object without any intermediary transformation (so we get the exact answer object that we need).

We ([Etcher](https://www.etcher.io)) are building a configuration/reconfiguration solution for OS images (see https://github.com/resin-io/etcher/issues/718#issuecomment-251005065 for details), and will be presenting complex Inquirer forms to users that would greatly benefit from grouping answers in different nested levels in order to plug the answers directly to our (re)configuration engines.

There are also cases where we ask similar sets of questions repeatedly in a dynamic fashion (for example, we ask ssid and key for several wifi networks), and without support for nested properties, a lot of top-level properties get unwieldy very quickly.

SBoudrias:
I've been thinking about this for a little while, and I think that's a good and easy improvements. I'm merging :)

Thanks for sending this out!

tandrewnichols:
Is there a way to not this? In my case, the names of the questions are generated dynamically based on filenames, so I end up with answers something like:

```
{
  'foo/bar/baz' {
    js: true
  }
}
```

I don't actually object to the feature. It's pretty cool really. But it would be nice if there were a way to turn it off or otherwise say "don't do it this time."

jviotti:
Ugh, I didn't think of these edge cases when implementing the patch, I'm very sorry about that! Let me think of a way we can fix this...

jviotti:
What about my second suggestion in this thread, of only enabling this functionality if the question name is an array of strings (which `_.set` happily takes as an alternative)?

I'll prepare a PR for it to speed things up in case @SBoudrias agrees :)

tandrewnichols:
That could be one way to do it, although I feel like my use case is the _more_ edgy edge case. I could see wanting to have it work the other way most of the time. I think filenames are not going to be _that_ common, so I'm willing to use whatever workaround you come up with. Right now, as a workaround, I'm replacing "." with "|" when generating the questions and then in the callback, mapping the answers back to "." (which isn't great, but it does work).

jviotti:
I guess an array of paths its the only non-intrusive way to introduce this, since I imagine there might be other people using periods in their question names for some reason. Adding a boolean option to enable/disable this behaviour seems too much of a burden IMHO.

I made the change I proposed above here: https://github.com/SBoudrias/Inquirer.js/pull/475. Looking forward to @SBoudrias' input!

PR #427, opened 2016-10-07
Allow ESCAPE in list prompt

guigrpa:
This (small) PR allows pressing ESC while in the list prompt. When this happens, the prompt is closed exactly in the same way as when it is normally answered, but a `null` value is returned.

I've included a unit test to validate the new functionality.

### DIFF
--- a/lib/prompts/list.js
+++ b/lib/prompts/list.js
@@ -147,7 +147,6 @@ Prompt.prototype.onCancel = function () {
   this.done(null);
 };
 
-
 /**
  * When user press a key
  */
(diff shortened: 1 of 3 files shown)


guigrpa:
Maybe I'm wrong, but I believe the failed test has nothing to do with my additions.

SBoudrias:
What's the use case for escaping from the list prompt? Why would a user escape rather than just `ctrl-c` out of the process?

guigrpa:
We have an app in production that would greatly benefit from this:

* It uses the list prompt to present the user with several nested menus. The user can go down the menu hierarchy by choosing a menu option, and up by choosing the "Back" option. Pressing ESC would be a faster way, since some menus are quite long.

* For some menu options, it requests several answers from the user, some of them via the list prompt. Cancelling via the ESC key would also be great.

In none of these cases ctrl-c is useful, since we don't want to exit the app.

guigrpa:
@SBoudrias Any chance this might land? Would you need any additional info?

guigrpa:
I've noticed that this approach works on Windows but **not on Linux machines**, given how Linux treats ESCAPE sequences. Furthermore, I've developed a new extension for our particular needs (a nested tree of menus based on the List prompt) that allows shortcuts to be associated with list choices (F5, backspace, etc.).

For both reasons, I'm closing this PR. Please tell me if you're interested in a new PR adding shortcut support to the List prompt (simple shortcuts only for the time being, i.e. no modifier keys, but more complex shortcuts could be added afterwards).

Issue #442, opened 2016-11-06
editor issues


hakase:
Using editor type a second time in a question list causes all sorts of weird behaviour. This happens whether it follows the first directly or there are intermediate questions of other types. Tested on Arch in terminator with nano & vim and in console as well.
```
'use strict';
var inquirer = require('inquirer');

var questions = [
  {
    type: 'editor',
    name: 'bio',
    message: 'short bio',
  },
  {
    type: 'editor',
    name: 'rabbit',
    message: 'Please write about saru',
  }
];

inquirer.prompt(questions).then(function (answers) {
  console.log(JSON.stringify(answers, null, '  '));
});
```
The first editor works normally. Sometimes the message from the second editor type is visible over the top of nano as an overlay. Keyboard input works briefly although the first character is often swallowed then the cursor jumps all over the place, random letters occur when you type and many keys seem to be disabled. Pasting anything into nano immediately prints gibberish and disables half the keyboard.

After killing nano the following error is visible:
```
/home/someuser/src/karasu/scripts/addlores/node_modules/rx/dist/rx.js:710
    if (disposable.isDisposed) { throw new ObjectDisposedError(); }
                                 ^
ObjectDisposedError: Object has been disposed
```
Vim is pretty much the same except that you can exit it without killing the process since ctrl-keys aren't required and the terminal is completely messed up afterwards.

moqmar:
I have the exact same issue, but it doesn't even require two editors - the problem occurs always if the editor is not the last question:

```javascript
require("inquirer").prompt([
    { type: "editor", name: "test1", message: "Enter some text:" },
    { type: "input", name: "test2", message: "Something else:" }
]).then(answers => console.log(answers))
```

PR #451, opened 2016-11-13
Fix for #442 (editor comes up again if it's not the last question / ObjectDisposedError)

moqmar:
The editor result object was properly disposed (which then causes the error message), but the editor stayed subscribed to line events (which causes it to come up again after the next question).  
That subscription is now disposed properly, too, so editor questions can be used again if they are not the last question. So, this fixes issue #442 (editor issues).


moqmar:
Argh, I can't even commit messages... Hope that's still fine :stuck_out_tongue:

SBoudrias:
Thanks!

Issue #454, opened 2016-11-15
Allow cancelling input with Esc key


laggingreflex:
I want the user to be able to escape out of the input by hitting <kbd>Esc</kbd> key. 

There should be an option which if turned on tells Inquirer to reject the promise of the prompt when user hits <kbd>Esc</kbd>, and let the program handle that rejection. 

It's kind of similar to #148 but it shouldn't exit, just reject. 

Was this almost implemented in #427 but only for list input? I haven't looked into it in detail but if it's what I think it is, that's exactly what I wanted but for all kinds of input - answer, checkbox etc. 

And would it also solve #291? It asks for a similar feature but using <kbd>Ctrl</kbd><kbd>c</kbd>. I think <kbd>Esc</kbd> would be much better. 

Thoughts?

episodeyang:
this is such an important issue. I hope this is merged.

sidharthv96:
It is not possible to cancel List input either using `Ctrl + C` or `Esc`

SBoudrias:
With the new API, you can now cancel a prompt cleanly. [More on this in the README](https://github.com/SBoudrias/Inquirer.js#canceling-prompt)

So for this use case, you'd set an event listener for the escape key, and cancel the prompt (in user code, so there's a bit of plumbing to do.) You can take inspiration from this recipe here [doing the same thing, but with a timeout](https://github.com/SBoudrias/Inquirer.js#get-default-value-after-timeout).

If you need something more precise, consider building a custom prompt. This should also be simpler than before with [the new core](https://github.com/SBoudrias/Inquirer.js#create-your-own-prompts).

PR #455, opened 2016-11-15
[WIP] Add escape key support

laggingreflex:
For #454. Work in progress for adding allowing escape key support (for cancelling input).

This adds some preliminary stuff. 

Lets you define `opt.onEscapeKey` function that's invoked when user presses escape key

Also allow passing a value to `opt.onEscapeKey`

```js
inquirer.prompt({
  //...
  onEscapeKey: 'some value'
  // equivalent to:
  onEscapeKey: function() { this.done('some value') }
});
```

### DIFF
--- a/lib/prompts/base.js
+++ b/lib/prompts/base.js
@@ -8,6 +8,7 @@ var chalk = require('chalk');
 var runAsync = require('run-async');
 var Choices = require('../objects/choices');
 var ScreenManager = require('../utils/screen-manager');
+var observe = require('../utils/events');
 var Promise = require('pinkie-promise');
 
 var Prompt = module.exports = function (question, rl, answers) {
@@ -53,6 +54,7 @@ var Prompt = module.exports = function (question, rl, answers) {
  */
 
 Prompt.prototype.run = function () {
+  this.events = observe(this.rl);
   return new Promise(function (resolve) {
     this._run(function (value) {
       resolve(value);
--- a/lib/prompts/base.js
+++ b/lib/prompts/base.js
@@ -28,6 +28,9 @@ var Prompt = module.exports = function (question, rl, answers) {
     },
     when: function () {
       return true;
+    },
+    onEscapeKey: function () {
+      return true;
     }
   });
 
@@ -55,6 +58,7 @@ var Prompt = module.exports = function (question, rl, answers) {
 
 Prompt.prototype.run = function () {
   this.events = observe(this.rl);
+  this.events.escapeKey.takeUntil(this.events.line).forEach(this.onEscapeKey.bind(this));
   return new Promise(function (resolve) {
     this._run(function (value) {
       resolve(value);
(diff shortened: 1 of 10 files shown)


laggingreflex:
Lets you define `opt.onEscapeKey` function that's invoked when user presses escape key

```js
inquirer.prompt({
  type: 'input',
  //...
  onEscapeKey: () => { /* when user presses escape key */ }
});
```

Works for all prompt types

laggingreflex:
Also allow passing a value to `opt.onEscapeKey`

```js
inquirer.prompt({
  //...
  onEscapeKey: 'some value'
  // equivalent to:
  onEscapeKey: function() { this.done('some value') }
});
```

SBoudrias:
Okay, so I've been thinking about this PR and the functionality it provides. Overall, I don't think this is the right approach.

Inquirer shouldn't provide stricter APIs surface to the end user for specific use cases. I want a more generic way to allow these kind of behavior to be define by the user outside of Inquirer. Inquirer should then only provide lower level API allowing you to circumvent the current flow.

What I wish would happen is that outside of Inquirer, you set a listener for the escape key. Then inside the escape key press callback you abort the current question, maybe return to the previous prompt or use the RX api to push a different prompt in the queue - basically anything you want.

I'll close this PR as it doesn't go in a direction I'd be willing to merge. But feel free to think this issue over and send another PR reversing the control from Inquirer to the api user.

laggingreflex:
No problem, appreciate the feedback. I did actually anticipate this might not go well with the existing code the more I began to grasp it. I'll see if I can come up with something better. 

Could you expand on this:

>outside of Inquirer, you set a listener for the escape key

Firstly, other than the "outside" aspect of it, was the `onEscapeKey` function at least close to the listener that you wanted? The user still could've done anything in there, even throw an error to abort the whole inquiry. I imagined more could've been done with `this` in it but I didn't explore much.

Or I'm guessing you don't want anything exposed that lets you use `this` of Inquirer? 

What would be the ideal thing that could be provided to the user to set a listener for the escape key? `readline` module? 

I haven't checked out the RX api in detail yet, but it seems to be an important wrapper around `readline`.

SBoudrias:
@laggingreflex I think the user should use the readline module themselves to listen to the esc key. In pseudo code:

```js
rl.on('keypress', () => {
  inquirer.abort();
});
```

I'd explore what option there is using the RX interface in this case.

whitecolor:
@SBoudrias It is not exactly clear how to correctly implement it with `rl.on('keypress`

I tried to insert before prompt which could be canceled and return to main menu made:
```
    const rl = readline.createInterface({
      input: process.stdin,
      output: process.stdout
    })
    rl.on('keypress', () => {
      showInitialMenu()
    });
```

But it doesn't seem to work. Can you elaborate?

c0bra:
@whitecolor this worked for me:

whitecolor:
Yeah I know it thanks.

Issue #456, opened 2016-11-16
'expand' - make help default


ihorskyi:
What is a correct syntax to make `expand`'s `help` item to be default instead ot the very first `Item A`?

```javascript
const QUESTION = {
  type    : 'expand',
  name    : 'root',
  message : 'Choose an item',
  default : 'help',    // <------------ kind of like this
  choices : [
    { key:'a', value:'item_a',  name:'Item A'  },
    { key:'b', value:'item_b',  name:'Item B'  }
  ]
}
```

SBoudrias:
Just use a `rawlist` for a very similar UI. Expand prompts cannot be expanded by default.

ihorskyi:
@SBoudrias but `rawlist` won't accept any other shortcut keys, except item's index?

SBoudrias:
Yes, that's correct.

If you just want to present a list of options, then we consider indexes to be a better UX for the end user.

ihorskyi:
This is a workflow I'm trying to achieve:

1. User launches a cli app and sees lot of options (let's say 15)
2. User forgot which key corresponds to item so he just press `Enter`
3. Help shows items and their short-cut keys (**same as used in CLI version**)
4. User clicks that key and action is triggered

Thus having same shortcut keys is very important, b/c in most cases user will remember most used ones in CLI and will easily use same in inquirer

PR #457, opened 2016-11-17
Default value for expand can be 'help'

ihorskyi:
#456


SBoudrias:
This wouldn't display the prompt expanded, it would just default to expand if the user press enter. Is that correct? Is that what you want?

ihorskyi:
@SBoudrias yep, that's it

SBoudrias:
How about we just make help the default option if no `default` key is provided?

ihorskyi:
@SBoudrias That would make even more sense

SBoudrias:
@ihorskyi cool, I think changing that will make this review good to merge.

ihorskyi:
@SBoudrias great, will change that tomorrow.

lib/prompts/expand.js
```diff
@@ -223,10 +223,14 @@ Prompt.prototype.validateChoices = function (choices) {
  */
 Prompt.prototype.generateChoicesString = function (choices, defaultIndex) {
   var defIndex = 0;
```
SBoudrias: Just change this to `this.opt.choices.length - 1;`

lib/prompts/expand.js
```diff
@@ -222,11 +222,15 @@ Prompt.prototype.validateChoices = function (choices) {
...
  */
 Prompt.prototype.generateChoicesString = function (choices, defaultIndex) {
-  var defIndex = 0;
+  var defIndex = this.opt.choices.length - 1;
+  var defStr = this.opt.choices.pluck('key');
+
   if (_.isNumber(defaultIndex) && this.opt.choices.getChoice(defaultIndex)) {
     defIndex = defaultIndex;
+  } else if (_.isUndefined(defaultIndex)) {
+    defIndex = defStr.length - 1;
```
SBoudrias: You don't need this anymore

lib/prompts/expand.js
```diff
@@ -222,11 +222,13 @@ Prompt.prototype.validateChoices = function (choices) {
  * @return {String} The rendered choices key string
  */
 Prompt.prototype.generateChoicesString = function (choices, defaultIndex) {
-  var defIndex = 0;
+  var defStr = this.opt.choices.pluck('key');
```
SBoudrias: I see you reverted the use of `this.opt.choices.length`, any reason for that? Doesn't look like the pluck is useful here as we only want the length.

lib/prompts/expand.js
ihorskyi: @SBoudrias this was changed in following commit, it is outdated now

lib/prompts/expand.js
ihorskyi: @SBoudrias The reason to use `this.opt.choices.pluck('key')` is to get choices with `keys` excluding separators. 
`this.opt.choices.length` === 6 
`defStr.length` === 5

SBoudrias: There's `choices.realLength`

SBoudrias:
Great, thanks!

PR #475, opened 2016-12-21
Only interpret arrays of strings as object paths in question names

jviotti:
The following pull-request: https://github.com/SBoudrias/Inquirer.js/pull/425 introduced a change
where we would use `_.set` to set a value in the answers object, since
that Lodash function has the nice property that it will interpret
strings as object paths, and will create nested objects if the question
name contains periods.

As raised in a comment after the pull-request was merged, this causes
problems when the question names represent file names containing
extensions, since `_.set` will incorrectly interpret the extension
period as a path delimiter.

In order to solve this regression, while still preserving this
functionality, this commit makes use of `_.set` only of the question
name consists of an array of strings.

In summary, a question name such as `[ foo, bar, baz ]` will be
interpreted as an object path, while `foo.bar.baz` will be interpreted
literally.

See: https://github.com/SBoudrias/Inquirer.js/pull/425
Signed-off-by: Juan Cruz Viotti <jviotti@openmailbox.org>

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ A question object is a `hash` containing question related values:
 
 - **type**: (String) Type of the prompt. Defaults: `input` - Possible values: `input`, `confirm`,
 `list`, `rawlist`, `expand`, `checkbox`, `password`, `editor`
-- **name**: (String) The name to use when storing the answer in the answers hash. If the name contains periods, it will define a path in the answers hash.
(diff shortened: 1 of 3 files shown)


SBoudrias:
Hey, I'm not really feeling like publishing a new major bump to revert that change. Given I haven't seen other people complain about that feature, I think it's fine to keep it as is for now.

Issue #485, opened 2017-01-22
BottomBar.log.write doesn't work


zaide-chris:
After updateBottomBar is called any calls to log.write don't seem to do anything.

This example works up to 0.12.0 then stops with 1.0.0:
```javascript
const inquirer = require("inquirer");

var clock = [
  '/ Ticking', '| Ticking', '\\ Ticking', '- Ticking',
  '/ Tocking', '| Tocking', '\\ Tocking', '- Tocking'
];
var i = 0;
var ui = new inquirer.ui.BottomBar();

var tickInterval = setInterval(() =>{
  ui.updateBottomBar(clock[i++ % clock.length]);
}, 250);

ui.log.write('Setting 10 Sec Timer:');
setTimeout(function(){
  ui.log.write('5 Sec left.');
}, 5000);
setTimeout(function(){
  ui.log.write('3 Sec left.');
}, 7000);
setTimeout(function(){
  ui.log.write('2 Sec left.');
}, 8000);
setTimeout(function(){
  ui.log.write('1 Sec left.');
}, 9000);
setTimeout(function(){
  ui.log.write('Time Up.');
  clearInterval(tickInterval)
  ui.updateBottomBar('Done');
  ui.close();
}, 10000);
```

Expected output after the 10 sec wait is something like:
```
Setting 10 Sec Timer:
5 Sec left.
3 Sec left.
2 Sec left.
1 Sec left.
Time Up.
Done
```
Output I get is:
```
Setting 10 Sec Timer:
Done
```

My fix for this in 2.0.0 is replacing: 
```javascript
Prompt.prototype.writeLog = function (data) {
  rlUtils.clearLine(this.rl, 1);
  this.rl.output.write(this.enforceLF(data.toString()));
  return this.render();
};
```
With:
```javascript
Prompt.prototype.writeLog = function (data) {
  this.rl.output.unmute();
  this.clean();
  this.rl.output.write(this.enforceLF(data.toString()));
  this.render();
  this.rl.output.mute();
  return this;
};
```
In "lib\ui\bottom-bar.js"

SBoudrias:
Feel free to send a PR given it looks like you figured out a fix :)

daviestar:
I wonder if this issue is related to my [recent question on SO](http://stackoverflow.com/questions/41791589)?

I can confirm this fix works in my project.

PR #486, opened 2017-01-23
Fix BottomBar.log.write not working

zaide-chris:
Fixes SBoudrias/Inquirer.js#485

I hope I did this correctly new to using Git.


SBoudrias:
Thanks!

Issue #257, opened 2015-07-02
Is it possible to adjust menu list height?


kokujin:
I am using a prompt type "list" with several items, it starts to "scroll-wrap" after 6 items and below is a notice, "Move up and down to reveal more choices".

Is there a way to stop the wrapping or adjust the number of items displayed? I would like to display all choices at the same time, for example. 

Thanks

SBoudrias:
There's no option for this currently, and no plan to add it.

We might in the future be smarter and use the terminal height to determine the number of choice we display.

ihorskyi:
Any updates on this?

oleander:
+1. This is how my console looks right now. Would be nice if the height could adjust to the console.

![skarmavbild 2016-11-11 kl 18 33 24](https://cloud.githubusercontent.com/assets/220827/20224166/6fac0732-a83d-11e6-82dc-9441b1cf638e.png)

SBoudrias:
PR welcomed to auto adjust the height.

raisen:
It looks like this is possible by changing the pageSize parameter for the paginate prototype on utils/paginator.js - is that right?

PR #492, opened 2017-02-02
Auto height for pageSize

ihorskyi:
With `pageSize : 'auto'` height will be adjusted to terminal height.

Also refers to #257

### DIFF
--- a/test/specs/utils/paginator.js
+++ b/test/specs/utils/paginator.js
@@ -1,4 +1,5 @@
 /* eslint no-multi-spaces: "off" */
+/* eslint space-in-parens: "off" */
 /* eslint brace-style: "off" */
 /* eslint max-statements-per-line: "off" */
 var _ = require('lodash');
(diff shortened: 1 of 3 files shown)


SBoudrias:
Hey thanks for sending this out. I think this is heading in the right direction.

I think we just need to change two things:

1. Make `auto` the default behavior. If `pageHeight` is not defined, then automatically use the terminal window size.
2. We need to be smarter about letting room for the question message and the validation message. Instead of a hardcoded amount of lines to keep, let's calculate the messages sizes and remove them from the page height.

ihorskyi:
Hi @SBoudrias, first part is clear and I just pushed commit for it. But I'm not sure about second part, could you elaborate, please?

test/specs/utils/paginator.js
```diff
@@ -0,0 +1,50 @@
+/* eslint no-multi-spaces: "off" */
```
SBoudrias: Let's not overwrite linting rules on a per file basis.

test/specs/utils/paginator.js
```diff
@@ -0,0 +1,50 @@
...
+var sinon = require('sinon');
+var Paginator = require('../../../lib/utils/paginator');
+
+describe('Paginator', function () {
+  describe('Auto height (Terminal mode)', function () {
+    var paginator;
+    var height;
+
+    beforeEach(function () {
+      process.stdout.getWindowSize = sinon.stub().returns([80, 20]);
```
SBoudrias: This doesn't get restored (same with other test).

In this case you'd want `sinon.stub(process.stdout, 'getWindowSize').returns([80, 20])`

test/specs/utils/paginator.js
```diff
@@ -0,0 +1,50 @@
...
+      paginator = new Paginator();
+      height = paginator.getHeight.bind(paginator);
```
SBoudrias: Call `paginator.getHeight()` directly in tests.

To make it shorter, maybe iterate over the different test case rather than creating one `it()` block for each value.

sindresorhus:
ping @ihorskyi

PR #496, opened 2017-02-15
Scroller fixed to middle of list

mokkabonna:
This fixes a somewhat broken implementation of pageSize. It always fixed the list at position 4. So any pageSize below that caused the currently selected item to be out of the visible area.

Now we fix it to the middle of the list.


SBoudrias:
Looks good to me, thanks :)

PR #497, opened 2017-02-16
Clears screen at bottom of prompt

mokkabonna:
Based on https://github.com/SBoudrias/Inquirer.js/pull/338

TBH I don't fully understand these changes. But seems to do the trick, however many tests fail. :)  An extended QA should be done.

image

Seems the changes i inquirer.js is not required.

### DIFF
--- a/lib/ui/baseUI.js
+++ b/lib/ui/baseUI.js
@@ -2,6 +2,7 @@
 var _ = require('lodash');
 var MuteStream = require('mute-stream');
 var readline = require('readline');
+var ScreenManager = require('../utils/screen-manager');
 
 /**
  * Base interface class other can inherits from
@@ -33,6 +34,15 @@ UI.prototype.onForceClose = function () {
 };
 
 /**
+ + * Create a new Screen Manager and remember it for when there's a forced close
+ + */
+
+UI.prototype.createScreen = function () {
+  this.currentScreen = new ScreenManager(this.rl);
+  return this.currentScreen;
+};
+
+/**
  * Close the interface and cleanup listeners
  */
 
@@ -43,6 +53,12 @@ UI.prototype.close = function () {
 
   // Restore prompt functionnalities
   this.rl.output.unmute();
+
+  // In case this is a forced close, move the cursor to
+  // the end of the last active screen
+  if (this.currentScreen) {
+    this.currentScreen.releaseCursor();
+  }
 
   // Close the readline
   this.rl.output.end();
--- a/lib/ui/baseUI.js
+++ b/lib/ui/baseUI.js
@@ -17,6 +17,7 @@ var UI = module.exports = function (opt) {
   this.rl.resume();
 
   this.onForceClose = this.onForceClose.bind(this);
+  this.currentScreen = new ScreenManager(this.rl);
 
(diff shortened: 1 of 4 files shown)


lib/ui/prompt.js
```diff
@@ -72,8 +72,9 @@ PromptUI.prototype.processQuestion = function (question) {
 };
 
 PromptUI.prototype.fetchAnswer = function (question) {
+  var screen = this.createScreen();
   var Prompt = this.prompts[question.type];
-  var prompt = new Prompt(question, this.rl, this.answers);
+  var prompt = new Prompt(question, screen, this.rl, this.answers);
```
mokkabonna: @SBoudrias is this something you could live with? Ref the other discussion.

mokkabonna: Will take a look at the tests if it is ok. If not I might need some help with an alternative approach.

lib/ui/prompt.js
```diff
@@ -74,6 +74,8 @@ PromptUI.prototype.processQuestion = function (question) {
 PromptUI.prototype.fetchAnswer = function (question) {
   var Prompt = this.prompts[question.type];
   var prompt = new Prompt(question, this.rl, this.answers);
+  prompt.screen  = this.createScreen();
```
mokkabonna: I now set screen here. Is this ok?

lib/ui/baseUI.js
```diff
@@ -33,6 +34,15 @@ UI.prototype.onForceClose = function () {
 };
 
 /**
+ + * Create a new Screen Manager and remember it for when there's a forced close
+ + */
+
+UI.prototype.createScreen = function () {
+  this.currentScreen = new ScreenManager(this.rl);
```
SBoudrias: Let's move this line into the class constructor function.

lib/ui/prompt.js
```diff
@@ -72,8 +72,9 @@ PromptUI.prototype.processQuestion = function (question) {
 };
 
 PromptUI.prototype.fetchAnswer = function (question) {
+  var screen = this.createScreen();
```
SBoudrias: We'll be able to get rid of this function once we move the initialization to the constructor.

lib/ui/prompt.js
SBoudrias: I wonder if it'd make sense to store the prompt instead of the screen; eg`this.activePrompt = new Prompt();`

Then we could call `this.activePrompt.stop()` and that function would be in charge of clearing the screen. It's an extra layer, but it'd prevent having to pass `this.screen` as argument everywhere.

mokkabonna:
If we use the activePrompt idea, what do we then do with calls to this.screen.render in prompts?
Like this one:
https://github.com/SBoudrias/Inquirer.js/blob/master/lib/prompts/input.js#L69

I think we need to pass screen in.

mokkabonna:
I made some quick changes that you can see here: https://github.com/SBoudrias/Inquirer.js/compare/master...mokkabonna:no-artifacts-alternative?expand=1

And then I do in my prompt:

```
Prompt.prototype.close = function() {
  utils.down(this.rl, this.opt.pageSize);
}
```

Pagesize might not be the best, but this is just a POC.

PR #500, opened 2017-02-20
Clears screen at bottom of prompt - alternative

mokkabonna:
I made some quick changes that implements it the alternative way.

And then I do in my prompt:

```js
Prompt.prototype.close = function() {
  if (this.screen.extraLinesUnderPrompt > 0) {
   utils.down(this.rl, this.screen.extraLinesUnderPrompt);
  }
}
```

This approach makes way less changes to the code. I am in favor of this.

Should also document this behaviour maybe? Nice for prompt authors to know about.

### DIFF
--- a/lib/prompts/base.js
+++ b/lib/prompts/base.js
@@ -7,8 +7,9 @@ var _ = require('lodash');
 var chalk = require('chalk');
 var runAsync = require('run-async');
 var Choices = require('../objects/choices');
+var ScreenManager = require('../utils/screen-manager');
 
-var Prompt = module.exports = function (question, screen, rl, answers) {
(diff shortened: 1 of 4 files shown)


SBoudrias:
This looks like the right approach to me, but we need to implement the default close behavior. I don't want prompt authors to have to write this logic themselves all the time.

mokkabonna:
Sure

mokkabonna:
This ok?

SBoudrias:
Thanks!

foisonocean:
How can i use it?

mokkabonna:
Need to wait for it being released. That's up to @SBoudrias .

SBoudrias:
3.0.2 is out

PR #503, opened 2017-02-22
Restores functionality of clearing screen

mokkabonna:
After the changes you @SBoudrias  introduced here https://github.com/SBoudrias/Inquirer.js/commit/79eebac61d2b1843df77435ca938fe3283ff8f5a and that are in 3.0.2, the clearing of the screen does not work any more.

I located what change caused it. This should restore it. Tested with my autocomplete prompt https://github.com/mokkabonna/inquirer-autocomplete-prompt


SBoudrias:
Thanks for following up in this.

PR #509, opened 2017-02-27
Prevent extra items in list, higher than pageSize

mokkabonna:
Now if you have a list of pageSize +2 items they are all shown without pagination. IMO it should be strictly fixed pagesize.

image

Is there any other reason for the +2? Also should the "9" here https://github.com/SBoudrias/Inquirer.js/pull/509/files#diff-232d5c801450a865eb599f3512b9d8c0L27
actually be pageSize +2? I can't see any change when changing the 9.

Issue originally reported here: https://github.com/mokkabonna/inquirer-autocomplete-prompt/issues/23

### DIFF
--- a/lib/ui/baseUI.js
+++ b/lib/ui/baseUI.js
@@ -40,6 +40,8 @@ UI.prototype.close = function () {
   this.rl.removeListener('SIGINT', this.onForceClose);
   process.removeListener('exit', this.onForceClose);
 
+  this.rl.output.unmute();
+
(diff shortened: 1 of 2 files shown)


SBoudrias:
You're right, the 9 should also be changed. Mind adding this to your PR?

mokkabonna:
Then it should be **pageSize** instead right?

SBoudrias:
Thanks :)

This part is a bit weird because we didn't use to allow to pass a pageSize option at all. So the code was apparently only half fixed :P

Issue #506, opened 2017-02-23
Empty line printed with release 3.0.2


mvila:
Since release 3.0.2, an empty line is printed after the user has completed a simple prompt such as:

```js
inquirer.prompt([
  {
    name: 'color',
    message: 'Color'
  }
]);
```

mvila:
Yes, it is a big issue for me because I need to control exactly what is printed in the console. Is there a way to get rid of this extra line?

mvila:
Thanks @SBoudrias! 😄

PR #511, opened 2017-02-28
Fix for not clearing final line

mokkabonna:
Can you test this @SBoudrias ? This restores the correct functionality without giving an extra line for me, same for you?

Before you had an console.log('\n') there, but gives an extra line.

### DIFF
--- a/lib/ui/baseUI.js
+++ b/lib/ui/baseUI.js
@@ -29,6 +29,7 @@ var UI = module.exports = function (opt) {
 
 UI.prototype.onForceClose = function () {
   this.close();
+  console.log('');
 };
 
 /**


SBoudrias:
I think that fix is fine given it should only impact force close. As far as I understand #506 problem, it was about prompt closed following usual process (answer all question and be done with Inquirer)

But so in your case, the only time the prompt isn't cleaned properly is when you `ctrl-c` it?

mokkabonna:
Yes only on force close. The prompt is cleared until the last line but only horizontally to the position of the cursor position.

SBoudrias:
Okay, I'll release it tonight (in my TZ)

Issue #507, opened 2017-02-23
checkbox: accepting default choices still displays instructions after selection


mysticflute:
I have a checkbox prompt with some of the choices with checked:true.

When you accept these defaults without toggling any of them (space, a, i) it prints the line of the selected choices as expected but that line still contains the instruction text:
(Press <space> to select, <a> to toggle all, <i> to inverse selection)

![screen shot 2017-02-23 at 3 50 30 pm](https://cloud.githubusercontent.com/assets/327116/23278531/dcb08184-f9df-11e6-8d0b-8a96dfb24c09.png)

mysticflute:
actually this also happens if you use the <a> and <i> shortcuts too. So the only way to get rid of that is to use <space> on a specific choice

SBoudrias:
Good catch, that's a bug.

andyford:
I ended up here because I was trying to figure out why the keyboard hints disappeared as soon as you pressed space, a, or i (before pressing Enter to continue). It's actually very confusing to remove hints while a user is still trying to use the prompt.

PR #512, opened 2017-02-28
Checkbox: Do not show usage hints after selection  

anicke:
The usage hints will not be displyed after these events:
* Default choices are selected
* Using the "a" shortcut to select all items.
* Using the "i" shortcut to inverse current selection. 

Addresses #507

### DIFF
--- a/lib/prompts/checkbox.js
+++ b/lib/prompts/checkbox.js
@@ -37,6 +37,7 @@ function Prompt() {
   }
 
   this.pointer = 0;
+  this.firstRender = true;
 
   // Make sure no default is set (so it won't be printed)
   this.opt.default = null;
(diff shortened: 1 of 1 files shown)


lib/prompts/checkbox.js
```diff
@@ -86,7 +88,7 @@ Prompt.prototype.render = function (error) {
   var message = this.getQuestion();
   var bottomContent = '';
 
-  if (!this.spaceKeyPressed) {
+  if (this.hints) {
```
SBoudrias: Instead of tracking the `this.hints` value in many methods, it'd be simpler to think about `firstRender`.

What the code you changed basically means only render the hints on the initial render. So I think that's how we should model the code. Render hints on first render, hide afterwards.

anicke:
@SBoudrias Good idea! I have made a new patch.

SBoudrias:
Awesome thanks a lot!

PR #522, opened 2017-04-06
Password prompt should display no characters

jcready:
Fixes https://github.com/SBoudrias/Inquirer.js/issues/177

### DIFF
--- a/lib/prompts/password.js
+++ b/lib/prompts/password.js
@@ -7,15 +7,6 @@ var chalk = require('chalk');
 var Base = require('./base');
 var observe = require('../utils/events');
 
-function mask(input) {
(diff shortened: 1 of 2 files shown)


SBoudrias:
Hey, can we display a useful message to explain to the user their password won't show up as they type? This might be confusing to a few users who're not used to CLI and unix.

jcready:
@SBoudrias not sure why the codacy check is failing, do you want me to switch to using double-quotes for strings even though single-quotes are used throughout?

SBoudrias:
No worry about codacy, it's not really working out. I'll disable it.

Thanks for the PR, it's merged now. I'll make a release soon.

PR #526, opened 2017-04-21
Update validate fn param docs

sdgluck:
Since #254 it has been possible to receive the answers hash via an option's `validate` function.

This PR adds this information to the docs.

Cheers 👍


SBoudrias:
Thanks!

PR #532, opened 2017-05-16
Use rx-lite-aggregates instead of full rx

ransombriggs:
While profiling our dependencies we noticed that inquirer includes the full `rx` module which adds around `7MB` to the inquirer dependencies, vs `~500K` for `rx-lite-aggregates` + `rx-lite`.  One thing I was not entirely happy with was that this bumps `rx` back from `^4.1.0` to `^4.0.8`, which is the latest `rx-lite-aggregates` version.  I also added an `rx-lite` dependency because otherwise `3.1.2` satisfies the `*` version from `rx-lite-aggregates`.

I did see in a8b2ce91fa88dd1c2aeec97c3cf9c220f032acd2 that the `rx-lite` dependency was swapped out for `rx` which I presume was because the `reduce` functionality was needed looking at the diff.  Are there any other reasons why it was changed to `rx`?
```
-> % du -h -s node_modules/rx-lite-aggregates
104K	node_modules/rx-lite-aggregates
-> % du -h -s node_modules/rx-lite           
464K	node_modules/rx-lite
-> % du -h -s node_modules/rx                
7.1M	node_modules/rx
```

### DIFF
--- a/examples/rx-observable-array.js
+++ b/examples/rx-observable-array.js
@@ -1,5 +1,5 @@
 var inquirer = require('..');
-var Rx = require('rx');
+var Rx = require('rx-lite-aggregates');
 
 var questions = [
   {
--- a/examples/rx-observable-create.js
+++ b/examples/rx-observable-create.js
@@ -1,5 +1,5 @@
 var inquirer = require('..');
-var Rx = require('rx');
+var Rx = require('rx-lite-aggregates');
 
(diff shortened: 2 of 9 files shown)


SBoudrias:
Hey, thanks for sending this out! Huge improvement!

I think you're correct that I changed rx-lite to rx in order to get the reduce operation.

PR #546, opened 2017-06-08
Fix hidden password validation quirk

ngotchac:
When using a hidden password prompt, with a validation, after having raised an error the password was displayed. This PR fixes this.

See `examples/password.js` for an example of this bug.

### DIFF
--- a/examples/password.js
+++ b/examples/password.js
@@ -8,6 +8,18 @@ var inquirer = require('..');
 inquirer.prompt([
   {
     type: 'password',
+    message: 'Enter "Passw0rd"',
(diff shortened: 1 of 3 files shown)


SBoudrias:
Hey, thanks for sending the PR.

I think the solution was a bit overly complex, so I've pushed another simpler fix.

I'll make a patch release today. Thanks for the help though.

ngotchac:
Okay yeah I understand :) However does the error gets cleared with a hidden password input?

SBoudrias:
You're right the error doesn't get cleared.

PR #558, opened 2017-06-21
Allow non-string values as a list default option

jtillmann:
At the moment the list default value only allows Numbers (choice array index) and Strings (choice array values). However, values might as well be anything, not just Strings. (In my case entire database objects with their name as then name of the choice)

This PR allows for any type of default option to be used.

### DIFF
--- a/lib/prompts/list.js
+++ b/lib/prompts/list.js
@@ -39,8 +39,8 @@ function Prompt() {
     this.selected = def;
   }
 
-  // Default being a String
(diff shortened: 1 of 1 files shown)


jtillmann:
The Travis test failed since numbers outside ouf array index range should cause the default value to not be set, but my initial changed allow for those numbers to be looked for as choice values. 

I have changed this, however now numbers or no longer allowed as default values

Issue #293, opened 2015-10-02
Have to send two SIGINT to kill when prompt is open


mbleigh:
If you Ctrl-C out of a command while an inquirer prompt is open, it just adds a couple lines of space and then you have to Ctrl-C again to actually exit. I'd expect the calling program to be the one controlling what happens on SIGINT here.

Is this working as intended? If so, is there a way to be notified when Inquirer has trapped a SIGINT?

SBoudrias:
Inquirer shouldn't trap SIGINT. What's your OS/terminal/node version?

mbleigh:
Interesting...I tried to create a minimal repro but couldn't reproduce the behavior. I thought I had seen the behavior in `bower init` but now can't reproduce that either. Feel free to mark as closed, I must be mistaken about the cause of my issue.

ChristianAlexander:
I ran into this issue recently. Because of Inquirer.js's usage of `readline2`, SIGINT only gets sent to Inquirer, which is fine when Inquirer is in the only running process. 

For me, the problem came up when I had a child process running. Since Inquirer does not propagate the signal, it is never exposed to the main process. This essentially removes any ability to listen for any signals with `process.on`. 

The real question is, should SIGINT only go to Inquirer, and be required a second time for the process, or should Inquirer be responsible and ensure that SIGINT is sent properly?

SBoudrias:
@ChristianAlexander can you link to the Node documentation stating that readlines won't propagate signals to the process?

ChristianAlexander:
After looking into this further, there is no documentation explicitly stating that signals do not propagate. In fact, manually sending a SIGINT (ex: `kill -2 <node process id>`) does work.

The problem is that readline itself listens to the TTY directly, so a 'real' SIGINT isn't produced.
See: https://github.com/nodejs/node/blob/master/lib/readline.js#L689-L700

SBoudrias:
@ChristianAlexander mind opening an issue on Node and linking this issue?

ChristianAlexander:
I'm unsure if this should be considered an issue at the Node level. Perhaps there could be a case where it wouldn't be desirable to send a SIGINT on `ctrl+c` in a project directly dependent on readline.

However, I can not come up with a scenario where any Inquirer-based projects would use `ctrl+c`. Even if such a scenario were to exist, the event wouldn't be exposed to consumers of Inquirer.

[...]

PR #564, opened 2017-07-09
[fix] #293 Exit script when SIGINT signal received

adamreisnz:
This PR properly addresses https://github.com/SBoudrias/Inquirer.js/issues/293 by triggering a `process.exit(0)` on `SIGINT` on the readline.

### DIFF
--- a/lib/ui/baseUI.js
+++ b/lib/ui/baseUI.js
@@ -19,9 +19,9 @@ var UI = module.exports = function (opt) {
 
   // Make sure new prompt start on a newline when closing
   process.on('exit', this.onForceClose);
-  
+
   // Terminate process on SIGINT (which will call process.on('exit') in return)
-  this.rl.on('SIGINT', function() {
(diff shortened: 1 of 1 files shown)


SBoudrias:
Hey @adamreisnz, thanks for sending this. Do you have any documentation of this? I'm really unclear why Node wants us to manually exit the process with a `process.exit()` call rather than handling it automatically.

SBoudrias:
Okay, doesn't seem to be documented very well, but listening on the SIGINT cancel the default handler; which is why we need to the exit the process ourselves.

I'll merge once CI pass.

adamreisnz:
Thanks, yes I followed the link posted in the issue, and after reading some more I stumpled upon an issue in Node repo which still isn't resolved. Long story short, this seemed to be the way to do it, listen for SIGINT on the readline instance like you were already doing, but then send terminate the process because that's not done.

PR #568, opened 2017-07-14
add inquirer-select-line to plugins

adam-golab:
I added new inquirer prompts for selecting index in an array where should be added a new element.


SBoudrias:
Looks cool! Thanks for putting this out :)

PR #569, opened 2017-07-18
Better way to kill process

adamreisnz:
As per https://github.com/SBoudrias/Inquirer.js/issues/293#issuecomment-315944753, this change still keeps the desired behaviour of killing the process on CTRL-C, but it gives parent processes a chance to clean up.


welwood08:
I happened to be skimming through the code looking for something unrelated and noticed there's cleanup code 20 lines below this change that should probably be synced to remove the new listener.

adamreisnz:
Possibly yes, it looked to me like there was some redundancy, but as I didn't author this package nor spent a lot of time analysing the code, I wasn't sure how refactoring it would impact other area's. Probably best for the package owner to have a look.

PR #571, opened 2017-07-28
Added `prefix` and `suffix` custom option

Rawnly:
Added `prefix` and `suffix` options in questions objects


jimthedev:
Any update on this one?

SBoudrias:
Thanks for the PR.

Sorry it took so long to review. I wasn't totally sure this was the approach I'd like to take on the issue, but finally I think that's a fair option. I'll merge and release with next version.

Fynnx1:
Won't `this.opt.prefix` if undefined cause 

`undefined question content`?

SBoudrias:
@Fynnx1 we assigned default values. Are you having this issue?

PR #578, opened 2017-08-19
Adding inquirer-command-prompt to the plugins list

sullof:
@SBoudrias 
I have just published a new version of the `inquirer-command-prompt` plugin, adding an example and a few tests. I think it is ready.


SBoudrias:
Thanks :)

PR #583, opened 2017-08-30
Fix cursor behavior

LitoMore:
This bug can reproduce in [this case](https://github.com/LitoMore/inquirer-chalk-pipe/blob/89502b9c7e6ce194e155455d8c7999f8f0524ae3/index.js#L60).

ANSI escape code should not be counted in the length.

**Expected behavior**

inquirer-chalk-pipe-expected

**Actual behavior**

inquirer-chalk-pipe-expected-behaivor

**How to review this pull reuqest**

```bash
$ git clone -b instant-preview https://github.com/LitoMore/inquirer-chalk-pipe.git
$ cd inquirer-chalk-pipe
$ npm install
$ node example.js
# Then, try `green` or other styles
```


SBoudrias:
Thanks!

PR #589, opened 2017-09-17
The delimiter can now be set with an option in question config

Fynnx1:
```js
[
 {
    name: 'example',
    message: 'What is your name?  ',
    delimiter: chalk.blue('+'),
    validate: value => {
      if (value) {
        return true;
      } else {
        return 'Please answer this question' && false;
      }
    },
  }
]
```

Will result in, 

screen shot 2017-09-17 at 15 40 14

If no delimiter is set, it will default back to the green `?`

screen shot 2017-09-17 at 15 41 15

I added this because I was after a quick and easy way to set the delimiter, I also couldn't find any other way to do it. So a quick read over the source-code and this is the result.

### DIFF
--- a/lib/prompts/base.js
+++ b/lib/prompts/base.js
@@ -8,25 +8,24 @@ var chalk = require('chalk');
 var runAsync = require('run-async');
 var Choices = require('../objects/choices');
 var ScreenManager = require('../utils/screen-manager');
-
-var Prompt = module.exports = function (question, rl, answers) {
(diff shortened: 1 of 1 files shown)


SBoudrias:
Hey, just merged #571 that does the same thing. Hope it'll accomplish the same for your issue with next release.

Fynnx1:
Yep, just noticed. Probably should've read over the PR's... But yes, that will also work too!

PR #592, opened 2017-09-26
Friendly error message when an internal error

nuintun:
The debugging is terrible when I write a plugin!

Before:
tim 20170926222854

After:
tim 20170926222753

### DIFF
--- a/lib/ui/prompt.js
+++ b/lib/ui/prompt.js
@@ -43,7 +43,12 @@ PromptUI.prototype.run = function (questions) {
       return this.answers;
     }.bind(this), {})
     .toPromise(Promise)
-    .then(this.onCompletion.bind(this));
+    .then(this.onCompletion.bind(this))
+    .catch(function (error) {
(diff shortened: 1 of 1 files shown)


lib/ui/prompt.js
```diff
@@ -43,7 +43,12 @@ PromptUI.prototype.run = function (questions) {
       return this.answers;
     }.bind(this), {})
     .toPromise(Promise)
-    .then(this.onCompletion.bind(this));
+    .then(this.onCompletion.bind(this))
+    .catch(function (error) {
+      console.error(error);
+
+      process.exit();
```
SBoudrias: Let's not `process.exit`, this prevents users from handling the exception.

nuintun: Already deleted ```process.exit```.

nuintun:
Closed, ```Promise.catch``` worked now, but it didn't work before!  Very strange...

PR #596, opened 2017-10-05
fix: use lodash modules to allow deduping in consumers

darscan:
Replaces `lodash` with individual `lodash` modules. This reduces the size of `inquirer` (from 8.0M to 3.8M), but more than that allows de-duping of `lodash` modules for consumers.

```
# In an empty dir

$ npm i inquirer
$ du -sh node_modules
8.0M  node_modules

$ trash node_modules

$ npm i ../Inquirer.js
$ du -sh node_modules
3.8M  node_modules
```
I've left `lodash` as a devDependency as a convenience for the unit tests.


darscan:
Seems `codeclimate` is not happy with any of the open PRs :/

darscan:
Hi @SBoudrias Any thoughts on this? I realise that it's probably a bit extreme.. At the same time, reducing the filesize for consumers seems reasonably beneficial.

SBoudrias:
Hey, it is indeed a bit extreme. I think `lodash` is often cached and used within subdependencies.

I don't think the complexity of this change is worth the benefit. Thanks for sending the PR though!

darscan:
Thanks, I agree - thought I'd just raise it for fun/discussion. Cheers :)

PR #601, opened 2017-10-21
Fix extra newline on windows without decremented width.

leak4mk0:
Hello,
On Windows, extra newline is inserted if a newline is inserted at the right end of the terminal.
Terminal width is decremented by 1 to avoid this problem now.
I would like to propose a method by moving the cursor.
Thank you.


SBoudrias:
Hey @leak4mk0, I've been putting aside reviewing this PR for a little while now. I read the code and I don't understand the bug you're having.

Can you post screenshots of the behavior before and after? And maybe an example to reproduce the issue you're having? Without this information, I can't really make an informed decision as to merge this PR and if this really is the proper fix for the issue. Thanks

leak4mk0:
Thank you for your reading my PR.

First, I want to fix slipped cursor on win32.
This problem is occurred by decremented width.
master

So, I tried to remove decrement of console width on win32 (File: screen-manager.js, Line: 113).
But new problem is occurred and I understood why console width is decremented..
On windows, if there is newline at right end of console, extra newline appears.
fix-width

Therefore, I got an idea.
I tried to insert cursor up ("\x1b[1A") after newline, if there is newline at right end.
fix-line-control

SBoudrias:
@leak4mk0 thanks for the explanation. Can you try to make the code a bit more understandable?

Right now this is a bit too confusing and quite unclear how it fix the issue.

I'd suggest assigning more descriptive variable names, stop using or at least removing nested ternary operators, and maybe split the core logics pieces into many functions.

Also, would the bug be fixed on windows if we simply do not insert `\n` at the line end inside `breakLines`?

leak4mk0:
Thank you for your reply.

As you say, this code is not easy for reading, sorry.
I'll rewrite this code more simply.

I tried your way ("do not insert `\n` at the line end") before I tried cursor-up way.
But, duplicate lines appeared.
image

SBoudrias:
@leak4mk0 thanks! BTW I've made some major code upgrade in the repo today, might have created a few conflicts on your side.

leak4mk0:
I'm sorry for being late.
Most of this problem seems to be fixed in new screen manager.
So, I fixed only two points for Windows console.
1. Remove extra right space on Windows. (screen-manager.js: 114)
2. Prevent move cursor if cols equals to 0. (screen-manager.js: 76)
    Because behavior of `cursorForward(0)` looks like one of `cursorForward(1)` on Windows cmd.exe.

SBoudrias:
Thanks a lot :)

PR #602, opened 2017-10-23
Paginator now handles multiple lines

joshkel:
Prior to this, a list with multi-line items works erratically; the current selection might be off the screen, or the list height may vary depending on which items are selected.

As part of this, Paginator now takes an optional ScreenManager instance. Prompts have been updated accordingly.

I noticed this most clearly with yarn's upgrade-interactive feature. To demonstrate:

1. Clone the yarn repo.
2. Within it, run `yarn upgrade-interactive` in a window that's too small to display the resulting table.
3. The current selection is off the top of the screen, and scrolling through the list doesn't work great in general.

The improvement can also be noticed by running [examples/long-list.js](https://github.com/SBoudrias/Inquirer.js/blob/651a569ccf37a8d8dd6ae4f348c414bab3649c19/examples/long-list.js). Without these changes, the height of the list changes depending how long individual list items are.

I have ran long-list and yarn upgrade-interactive (which uses checkbox) and have ran the test suite but have not tested further.

### DIFF
--- a/lib/utils/paginator.js
+++ b/lib/utils/paginator.js
@@ -8,15 +8,24 @@ var chalk = require('chalk');
  * a subset of the choices if the list is too long.
  */
 
-var Paginator = module.exports = function () {
+var Paginator = module.exports = function (screen) {
   this.pointer = 0;
   this.lastIndex = 0;
+  this.screen = screen;
 };
 
 Paginator.prototype.paginate = function (output, active, pageSize) {
   pageSize = pageSize || 7;
   var middleOfList = Math.floor(pageSize / 2);
   var lines = output.split('\n');
+
+  if (this.screen) {
+    lines = this.screen.breakLines(lines);
+    active = _.sum(lines.map(function (lineParts) {
+      return lineParts.length;
+    }).splice(0, active));
+    lines = _.flatten(lines);
+  }
 
   // Make sure there's enough lines to paginate
   if (lines.length <= pageSize) {
--- a/lib/prompts/checkbox.js
+++ b/lib/prompts/checkbox.js
@@ -42,7 +42,7 @@ function Prompt() {
   // Make sure no default is set (so it won't be printed)
   this.opt.default = null;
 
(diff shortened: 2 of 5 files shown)


lib/utils/paginator.js
```diff
@@ -8,16 +8,25 @@ var chalk = require('chalk');
...
+  this.screen = screen;
 };
 
 Paginator.prototype.paginate = function (output, active, pageSize) {
   pageSize = pageSize || 7;
   var middleOfList = Math.floor(pageSize / 2);
   var lines = output.split('\n');
 
+  if (this.screen) {
+    lines = this.screen.breakLines(lines);
```
SBoudrias: Hey, I feel like it'd be pretty simple to make `breakLines` and `normalizedCliWidth` into fully standalone functions.

Then we wouldn't need to keep track of whether or not we received a ScreenManager instance.

joshkel: Thanks for the feedback. I'm happy to make whatever changes will help maintenance.

Since `normalizedCliWidth` needs an `output` value, I assume that that would be passed in as a parameter to `breakLines` and `normalizedCliWidth`?

Where would you like the standalone functions to reside? `utils.js`? Static functions on `ScreenManager`?

SBoudrias: They could be static on `ScreenManager` for now at least.

SBoudrias: The readline instance (e.g. used in `this.rl.output`) is always available on the UI classes as `this.rl`. So yeah, we can pass it in as an argument.

joshkel:
Hey. I'm sorry I didn't follow up on this like you'd asked. Thank you for merging it. Do you want me to open a separate PR to make the functions standalone?

SBoudrias:
@joshkel Oh hey, actually I think it's fine as is. I've played around the code thinking about fixing myself to make the next release; and after all, I thought it wasn't really clearer with standalone functions. So we're good like this :)

Thanks again for sending this PR

PR #607, opened 2017-10-31
Guard list against invalid string defaults

Froren:


lib/prompts/list.js
```diff
@@ -38,7 +38,8 @@ function Prompt() {
   if (_.isNumber(def) && def >= 0 && def < this.opt.choices.realLength) {
     this.selected = def;
   } else if (!_.isNumber(def) && def != null) {
-    this.selected = this.opt.choices.pluck('value').indexOf(def);
+    let index = this.opt.choices.pluck('value').indexOf(def);
```
SBoudrias: Thanks for sending the fix. This code would be cleaner as:

```js
this.selected = Math.max(this.opt.choices.pluck('value').indexOf(def), 0);
```

Mind updating?

Froren: Great catch, I'll use that from now on. Will update it in a while.

Froren: Updated.

PR #609, opened 2017-11-01
Adds string default option for list and rawlist

Froren:
- Feel free to drop this PR if it's not a direction you want to move in.

### DIFF
--- a/lib/prompts/expand.js
+++ b/lib/prompts/expand.js
@@ -226,10 +226,8 @@ Prompt.prototype.generateChoicesString = function (choices, defaultChoice) {
   if (_.isNumber(defaultChoice) && this.opt.choices.getChoice(defaultChoice)) {
     defIndex = defaultChoice;
   } else if (_.isString(defaultChoice)) {
-    let index = _.flatMap(choices.realChoices, function(c) {
-      return c.value;
-    }).indexOf(defaultChoice);
-    defIndex = Math.max(index, 0); 
+    let index = _.findIndex(choices.realChoices, ({value}) => value === defaultChoice);
+    defIndex = (index !== -1 ? index : defIndex);
   }
   var defStr = this.opt.choices.pluck('key');
   this.rawDefault = defStr[defIndex];
--- a/lib/prompts/expand.js
+++ b/lib/prompts/expand.js
@@ -218,13 +218,18 @@ Prompt.prototype.validateChoices = function (choices) {
 /**
  * Generate a string out of the choices keys
  * @param  {Array}  choices
- * @param  {Number} defaultIndex - the choice index to capitalize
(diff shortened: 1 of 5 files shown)


lib/prompts/expand.js
```diff
@@ -218,13 +218,18 @@ Prompt.prototype.validateChoices = function (choices) {
...
   var defIndex = choices.realLength - 1;
-  if (_.isNumber(defaultIndex) && this.opt.choices.getChoice(defaultIndex)) {
-    defIndex = defaultIndex;
+  if (_.isNumber(defaultChoice) && this.opt.choices.getChoice(defaultChoice)) {
+    defIndex = defaultChoice;
+  } else if (_.isString(defaultChoice)) {
+    let index = _.flatMap(choices.realChoices, function(c) {
+      return c.value;
+    }).indexOf(defaultChoice);
+    defIndex = (index !== -1 ? index : index);
```
SBoudrias: Overall looks good, just same comment as the other review about this piece `(index !== -1 ? index : index);`. It's cleaner (less code branches) if we use `Math.max(index, 0)`.

lib/prompts/expand.js
```diff
@@ -218,13 +218,18 @@ Prompt.prototype.validateChoices = function (choices) {
...
  */
-Prompt.prototype.generateChoicesString = function (choices, defaultIndex) {
+Prompt.prototype.generateChoicesString = function (choices, defaultChoice) {
   var defIndex = choices.realLength - 1;
-  if (_.isNumber(defaultIndex) && this.opt.choices.getChoice(defaultIndex)) {
-    defIndex = defaultIndex;
+  if (_.isNumber(defaultChoice) && this.opt.choices.getChoice(defaultChoice)) {
+    defIndex = defaultChoice;
+  } else if (_.isString(defaultChoice)) {
+    let index = _.flatMap(choices.realChoices, function(c) {
```
SBoudrias: Missed that earlier in the PR, but we can do all this in a single loop if we use Lodash `_.findIndex`

`_.findIndex(choices.realChoices, ({value}) => value === defaultChoice)`

(and let's do the same with the rawlist to keep the logic more consistent)

Froren: Sure, that sounds great.

lib/prompts/expand.js
Froren: I realize now that this was a bug, it should be `defIndex = (index !== -1 ? defIndex : index);`

I'll make the change in the commit to use `_.findIndex`, though I don't think we can use Math.max anymore with that, since expand's default is the end of the array.

Froren:
Updated again.

SBoudrias:
Thanks :) I'll cut a new release this weekend most likely

Issue #574, opened 2017-08-04
Wrong license used?


michaelzoidl:
Hi, at first, your project is awesome, i'm using it in several projects and its always a good time when working with it.

But sadly i discovered last week that the project is using a package which uses a package with a [LGPL](https://choosealicense.com/licenses/lgpl-3.0/) license, that means i can not use inqurier any more as long as it has the [`external-editor`](https://www.npmjs.com/package/external-editor) package which uses the [`jschardet`](https://www.npmjs.com/package/jschardet) package, (the `jschardet` package has the LGPL license)

That means btw you can only publish your software as MIT as long as the `external-editor` package is not substantially impare inquirer. :/

Do you think it's possible to replace the external-editor package? Or maybe check out if it's ok to use MIT for this project, i'm not a lawyer, so it's possible i'm wrong.

SBoudrias:
Seems to not be a problem according to this answer https://softwareengineering.stackexchange.com/questions/141847/how-does-using-a-lgpl-gem-affect-my-mit-licensed-application and the details on github when you look at the license: https://github.com/aadsm/jschardet/blob/master/LICENSE

kreig303:
@SBoudrias I hate to be a nag but even though you view it as not being a problem, any commercial organization attempting to use open source software will shy away from GPL and derivatives.  The FSF licenses are dated and vague.

Believe me, your software's usage will drop over time if you continue to use a piece with an LGPL... even if your own usage of `jschardet` is not threatened due to your source distribution.

SBoudrias:
@mrkmg Hey Kevin, any chance you could find an alternative to `jschardet`?

I know most alternative seems to rely on the ICU binary, so it might be a bit complicated to find a proper alternative.

michaelzoidl:
@SBoudrias but you don't have to find an alternative to `jschardet`, since you use the `external-editor` package which uses `jschardet`.
So the solution would be to find an alternative to `external-editor`, plus, all dependencies should be MIT.

mrkmg:
I would love find an alternative. This has come up before. I will put an effort in today to find an MIT package.

mrkmg:
I just launched a new version (2.1.0) which uses https://github.com/runk/node-chardet, licensed under the MIT License.

Also updated tests and verified compatibility from 0.12 to 9.2.0

SBoudrias:
@mrkmg thanks a ton :)

kreig303:
[...]

PR #616, opened 2017-11-27
Bumped version to 4.1.0, bumped external-editor dependency to ^2.1.0 …

kreig303:
…to remove jschardet LGPL license dependency.

Final fix for Issue #574.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "inquirer",
-  "version": "4.0.0",
+  "version": "4.1.0",
   "description":
     "A collection of common interactive command line user interfaces.",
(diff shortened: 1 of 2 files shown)


kreig303:
Ran `npm test` too... first time it failed on
```
1) inquirer.prompt
       hierarchical mode (`when`)
         should run prompt if `when` returns true:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```
but then the second time (et seq.) it ran fine... just fyi?

SBoudrias:
Thanks! I'll make a release tonight.

PR #632, opened 2018-01-10
feat(prompt): progress indicator for filter and validate with optional message

### LINKED ISSUE #631: FR: Show progress indicator on async validate
I have an async validation that takes a while. When the user presses enter, nothing seems to happen, until the validation completes. Is it possible to add some visual indication (for example, changing the question mark into a progress spinner), while the validation is running. The UI feels unresponsive now.

Unfortunately, I can't console.log something like that myself, because if the validation fails, any logging will cause the question to repeat, which is not what I want.

kbrandwijk:
WIP - First proposal for adding progress indicator to filter and validate step, with an optional message to be displayed during these steps. Includes an example. Will update tests and docs after initial review.

Closes #631

### DIFF
--- a/examples/filter-validate-progress.js
+++ b/examples/filter-validate-progress.js
@@ -1,0 +1,41 @@
+/**
+ * Filter and validate progress example
+ */
+
+'use strict';
+var inquirer = require('..');
+
+var questions = [
+  {
+    type: 'input',
+    name: 'first_question',
+    message: 'Question with filtering and validating text',
+    validate: async () => {
+      await new Promise(r => setTimeout(r, 3000));
+      return true;
+    },
+    filter: async answer => {
+      await new Promise(r => setTimeout(r, 3000));
+      return `filtered${answer}`;
+    },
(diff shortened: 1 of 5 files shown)


kbrandwijk:
This functionality could in theory also be moved from `base` to `screen-manager`, so `this.screen.render` could be called with additional flags to either add a spinner to the text or the bottomText. Please let me know what you think.

kbrandwijk:
It seems Codacy and the XO settings don't agree...

benmonro:
any reason this can't be merged? @kbrandwijk

SBoudrias:
@kbrandwijk I removed Codacy as its supports for eslint config is too flakey. It kept starting to mark PR as failing every few months.

I'd love to get this merged. Any chance you could resolve the conflicts and try to finalize the integration?

sam3d:
Don't suppose you have any updates on this @kbrandwijk? This would be incredibly useful to have!

dhensby:
This was merged via #1002 - so this can now be clsoed

PR #636, opened 2018-01-13
Use name as default message

stuartpb:
Following the style of wizards like `npm init` (see [PromZard](https://github.com/npm/promzard) and its precedents).


SBoudrias:
Thanks for sending the PR!

PR #640, opened 2018-01-19
Correct subscription method

### LINKED ISSUE #637: Error using editor at Subscription.dispose
When i use editor and save/exit, i get an exception: `Subscription.dispose is not a function`

Any idea why?

user3587412:
Fixes #637 

`forEach` now returns a promise instead of a disposable subscription (therefore `.dispose()` does not exist)


SBoudrias:
Thanks!

PR #642, opened 2018-01-24
Apply optional transformer to input before accepted

### LINKED ISSUE #359: Is it possible to change the color of the input beyond certain length?
Is it possible to change the color of the input beyond certain length? 

Somthing like
screen shot 2016-04-04 at 11 57 11 am

git-cz

---

May be something like:

``` diff
...
}, {
    type: 'input',
    name: 'subject',
-    message: 'Write a short, imperative tense description of the change:\n'
+    message: 'Write a short, imperative tense description of the change:\n',
+    transformer: function(line) {
+        return line.substring(0, 99) + chalk.red(line.substring(100))
+   }
}, {
....
```

and

[Inquirer.js/lib/prompts/input.js#L55-L70](https://github.com/SBoudrias/Inquirer.js/blob/master/lib/prompts/input.js#L55-L70) can invoke this `transformer function`

``` diff
Prompt.prototype.render = function (error) {
  var bottomContent = '';

    message += chalk.cyan(this.answer);
  } else {
-   message += this.rl.line;
+   message += reference.to.the.transformer(this.rl.line);
  }

  if (error) {
    bottomContent = chalk.red('>> ') + error;
  }

  this.screen.render(message, bottomContent);
};
```

Thanks!

travi:
closes #359

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ Array values can be simple `strings`, or `objects` containing a `name` (to displ
 - **prefix**: (String) Change the default _prefix_ message.
 - **suffix**: (String) Change the default _suffix_ message.
 
-`default`, `choices`(if defined as functions), `validate`, `filter`, `transformer`, and `when` functions can be called asynchronous. Either return a promise or use `this.async()` to get a callback you'll call with the final value.
+`default`, `choices`(if defined as functions), `validate`, `filter`, and `when` functions can be called asynchronously. Either return a promise or use `this.async()` to get a callback you'll call with the final value.
 
 ``` javascript
 {
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ A question object is a `hash` containing question related values:
 Array values can be simple `strings`, or `objects` containing a `name` (to display in list), a `value` (to save in the answers hash) and a `short` (to display after selection) properties. The choices array can also contain [a `Separator`](#separator).
 - **validate**: (Function) Receive the user input and answers hash. Should return `true` if the value is valid, and an error message (`String`) otherwise. If `false` is returned, a default error message is provided.
 - **filter**: (Function) Receive the user input and return the filtered value to be used inside the program. The value returned will be added to the _Answers_ hash.
-- **transformer**: (Function) Receive the user input and return the transformed value to be displayed to the user. The transformation only impacts what is shown while editing. It does not impact the _Answers_ hash.
+- **transformer**: (Function) Receive the user input and return the transformed value to be displayed to the user. The transformation only impacts what is shown while editing. It does not impact the `answers` hash.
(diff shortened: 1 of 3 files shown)


travi:
i added a test for the actual application of the transformer for the input, but i'm not familar enough with your test suite to know if that covers the functionality full in line with your normal approach. is there any additional testing that you want to be added?

if you're good with this approach, i'm happy to follow up and add this to the docs

SBoudrias:
Can you add documentation for this new feature?

travi:
sure thing. if you're good with this approach, i'll try to get this added to the docs tonight

SBoudrias:
Yes, I think the approach and UT are good.

README.md
```diff
@@ -113,12 +113,13 @@ A question object is a `hash` containing question related values:
 Array values can be simple `strings`, or `objects` containing a `name` (to display in list), a `value` (to save in the answers hash) and a `short` (to display after selection) properties. The choices array can also contain [a `Separator`](#separator).
 - **validate**: (Function) Receive the user input and answers hash. Should return `true` if the value is valid, and an error message (`String`) otherwise. If `false` is returned, a default error message is provided.
 - **filter**: (Function) Receive the user input and return the filtered value to be used inside the program. The value returned will be added to the _Answers_ hash.
+- **transformer**: (Function) Receive the user input and return the transformed value to be used inside the program. The value returned will be added to the _Answers_ hash.
```
travi: updating this documentation raises a question for me about the behavior, as it has been added. 

my intent with the transformer is to format what is shown to the user while they are typing into the Input (things like those listed in #359: change color after some # of characters, prevent typing beyond another # of characters, maybe provide an indication of the remaining characters like "15 characters left"). i would expect the formatting to be removed before the text was added to the _Answers_ hash or be used elsewhere in the program. essentially i would only want the formatting to be shown while typing and removed after the input is considered answered.

since i don't have full context about how things are wired together, do i need to modify the implementation at all to have it behave according to those expectations? are there other examples that i missed that behave in that way and i should model the docs description from?

SBoudrias: Yeah this only applies to the view layer. It doesn't modify the actual answer.

As such, this function cannot be used to prevent the user from entering more than X characters. It could only be used to modify the style of the input text.

travi: > As such, this function cannot be used to prevent the user from entering more than X characters. It could only be used to modify the style of the input text.

great clarification. so, my understanding was mostly in line with my goal, but i hadn't considered that angle. it doesn't prevent the overall goal, but very worth clarifying. 

i'll get this description updated

travi:
docs have been updated. let me know if you'd like any adjustments, but i think it should be good to go otherwise.

SBoudrias:
Thanks! I just updated piece of the documentation because the new API isn't actually asynchronous.

travi:
thanks for working through it with me. 

this is a great tool and really appreciate your work on it!

Issue #517, opened 2017-03-21
Auto answering


likerRr:
Is it possible to automatically answer the question? E.g. when user calls cli app with arguments I want to parse them and depending on their values automatically answer a question when it's asksed. 

So I would like to write something like this:

```javascript
const askName = {
  type: 'input',
  name: 'name',
  message: `What's your name?`,
  // auto also can be string/boolean any other valid type
  auto: session => process.argv.name ? process.argv.name : undefined
};
```

So if `auto`'s value is undefined, then the answer is prompted to user otherwise the answer automatically sets to returned from callback value.

It's nice to have this ability rather then don't ask the question at all because user can see the answer that was choosen by his actions (in my case by calling cli app with some arguments).

I tried to achieve such behavior from manually calling `prompt.ui.rl.emit('line', 'hello')` but I can't find sutable callback in inquirer's life cycle where I can do this.

sholladay:
Absolutely. This is a great pattern in [Yeoman](http://yeoman.io/) generators like [generator-seth](https://github.com/sholladay/generator-seth), so I will use that as an example.

Steps:
1. [Get a command line argument](https://github.com/sholladay/generator-seth/blob/acb0f92fb4da5ca03877db90220a25d089028fc9/app/index.js#L60).
2. [Dynamically show/hide a prompt](https://github.com/sholladay/generator-seth/blob/acb0f92fb4da5ca03877db90220a25d089028fc9/app/index.js#L136) based on whether the argument was provided.
3. [Merge the arguments and prompt answers](https://github.com/sholladay/generator-seth/blob/acb0f92fb4da5ca03877db90220a25d089028fc9/app/index.js#L200) with something like `Object.assign()`.

The linked example uses some Yeoman APIs for dealing with arguments, but could be refactored pretty easily. I would use [meow](https://github.com/sindresorhus/meow) for this in a standalone CLI.

This approach is quite flexible. For example, instead of skipping the prompt when the user has provided an argument, you could still show the prompt but make their argument the default value, which would be shown to them and behave kind of like a confirmation screen. To do this, you would pass the argument to `default` instead of `when`.

likerRr:
@sholladay thank you for explanation, but unfortunatelly it's not what I'm looking for. 

[...]

PR #645, opened 2018-02-12
Add tests for all `when` functions returning false

karlhorky:
Failing test to show that mutated `answers` array is discarded when all `when` functions return false.

Ref: #517

### DIFF
--- a/test/specs/inquirer.js
+++ b/test/specs/inquirer.js
@@ -526,6 +526,37 @@ describe('inquirer.prompt', function() {
       });
     });
 
+    it('should still return mutated answers if all `when`s return false', function() {
+      var prompts = [
+        {
+          type: 'confirm',
+          name: 'q1',
+          message: 'message',
+          when: function(answers) {
+            answers.q1 = 'foo';
+            return false;
+          }
+        },
+        {
+          type: 'confirm',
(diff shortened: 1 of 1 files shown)


SBoudrias:
Do you mean if only some `when` functions return false the changes aren't ignored?

karlhorky:
Yep, exactly. It accepts the mutations to the answers object if at least one prompt goes through.

karlhorky:
I thought it maybe could have something to do with this: https://github.com/SBoudrias/Inquirer.js/blob/258064f8b1a4002e3393455918912004748d8bea/lib/ui/prompt.js#L98-L100

SBoudrias:
#665 should've fix this issue :)

Let me know if that's not the case!

karlhorky:
Great, I'll take a look and report back if there are still issues, thanks!

karlhorky:
I've tested the fix from #665 by upgrading to `6.0.0` and the answers are indeed there. Thanks @yujunlong2000!

karlhorky:
I've tested the fix from #665 by upgrading to `6.0.0` and the answers are indeed there. Thanks @yujunlong2000 and @SBoudrias!

PR #652, opened 2018-02-23
Patch observable operators

dwieeb:
Hello, thank you for this wonderful package. 

I've noticed that by simply importing `inquirer` 5.1.0, **519** (~.28s on my computer) modules get added to the module cache.  Here's a quick script to test:

```js
const inquirer = require('inquirer');
console.log(Object.keys(require.cache).length);
```

By inspecting the module cache, I noticed that the majority of the modules being imported by inquirer are unused rxjs operators.

This PR brings the number of imported modules down to **152** (~.15s on my computer) by using rxjs [patching](https://github.com/ReactiveX/rxjs/tree/stable#installation-and-usage) instead of importing the entire set of rxjs functionality.

Please consider this may be a breaking change if dependent modules are expecting inquirer observables to have all operators attached (not sure if inquirer API returns observables anywhere--I don't use observables with inquirer).

### DIFF
--- a/lib/prompts/base.js
+++ b/lib/prompts/base.js
@@ -10,6 +10,12 @@ var runAsync = require('run-async');
 var Choices = require('../objects/choices');
 var ScreenManager = require('../utils/screen-manager');
 
+require('rxjs/add/operator/filter');
+require('rxjs/add/operator/mergeMap');
+require('rxjs/add/operator/share');
+require('rxjs/add/operator/take');
+require('rxjs/add/operator/takeUntil');
+
 class Prompt {
   constructor(question, rl, answers) {
     // Setup instance defaults property
@@ -87,7 +93,7 @@ class Prompt {
     var validate = runAsync(this.opt.validate);
     var filter = runAsync(this.opt.filter);
     var validation = submit
-      .flatMap(value =>
+      .mergeMap(value =>
         filter(value, self.answers).then(
           filteredValue =>
             validate(filteredValue, self.answers).then(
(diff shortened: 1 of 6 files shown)


dwieeb:
Regarding the Codacy checks: I followed the style from existing code. Is the linter configured properly?

SBoudrias:
@dwieeb might be an issue with Codacy. I'll check it later on.

I'm not sure about this branch, I feel like this wouldn't be a very clear change for people using the rx interface. This package has a lot of issues about size though, I feel I get a bug report or similar PR every few months... Not sure what to do about it, I feel like this should be fixed upstream.

dwieeb:
@SBoudrias I don't think Inquirer is _heavy_ (a fresh npm install only adds 32 packages). But when one of those packages reads and parses half a thousand JS files when the base module is required, runtime performance is poor, especially since this package is meant for CLI programs, which have users continually booting up and shutting down the node runtime. 

It looks like the next major version of RxJS [changes imports anyway](https://github.com/ReactiveX/rxjs#installation-and-usage). Perhaps when Inquirer upgrades to RxJS 6 (which will be a breaking change) this PR can be included?

SBoudrias:
> It looks like the next major version of RxJS changes imports anyway. Perhaps when Inquirer upgrades to RxJS 6 (which will be a breaking change) this PR can be included?

Yes, this would be awesome!

SBoudrias:
@dwieeb as I merged the rxjs 6 branch, I'll close this one. But feel free to follow-up if you think we can do more :)

dwieeb:
@SBoudrias Thanks! I do think we can do more to optimize. A first step would be lazily loading rxjs imports. Not sure how welcome you'll be to that! I'll open a no-pressure PR about it eventually.

PR #659, opened 2018-03-24
Pass answers object to transformer functions

### LINKED ISSUE #658: Pass answers object to transformer functions
I have a use case where I'd like to render the current input differently based on the answers to previous questions. This would be possible if transformer functions receive the answers object as a second parameter (currently they get one parameter, the user input). I'm happy to open a PR if this sounds useful, and it should be backwards compatible since the current parameter wouldn't change.

raingerber:
closes #658


SBoudrias:
I'll prepare a release later today.

raingerber:
Thank you!

PR #660, opened 2018-03-26
Add transformer example

LitoMore:
### Changes

- Add `transformer` example for input

[chalk-pipe](https://github.com/LitoMore/chalk-pipe) is a good use case for `transformer`.

Just try some colors like `orange`, `#ff99cc`..


SBoudrias:
Good idea, thanks :)

PR #661, opened 2018-03-26
Add support for transform the answer after accepted

LitoMore:
Similar to `transformer`, the `answerTransformer` will only impacts what is shown while answered. It does not modify the answers hash.

### Usage

```javascript
{
  type: 'input',
  name: 'fav_color',
  message: "What's your favorite color",
  answerTransformer: function(color) {
    return chalkPipe(color)(color);
  }
}
```

### Preview

image


LitoMore:
Regarding the coverage test: Am I missing some test?

SBoudrias:
@LitoMore any chance we could use the same transformer function to also transform what we display once the prompt is answered?

I'd like to limit the amount of new API we add.

LitoMore:
@SBoudrias Good point.

What about this solution?

```javascript
{
  type: 'input',
  name: 'fav_color',
  message: "What's your favorite color",
  transformer: {
    input: function(input, inputs) {},     // This will transform user's input
    answer: function (answer, inputs) {}   // This will transform answer
	// Even more...
  }
}
```

SBoudrias:
What about a flag passed to the transformer function? `isFinal` or something like this

lib/prompts/input.js
```diff
@@ -42,9 +42,14 @@ class InputPrompt extends Base {
     var bottomContent = '';
     var message = this.getQuestion();
     var transformer = this.opt.transformer;
+    var answerTransformer = this.opt.answerTransformer;
 
     if (this.status === 'answered') {
-      message += chalk.cyan(this.answer);
+      if (answerTransformer) {
+        message += answerTransformer(this.answer, this.answers);
```
LitoMore: > What about a flag passed to the transformer function? `isFinal` or something like this (@SBoudrias)

We should transform answer in `answered` case, how can we send the `isFinal` flag to here?

SBoudrias: add a 3rd argument? `{ isFinal: this.status === 'answered' }`

LitoMore:
@SBoudrias Code updated

SBoudrias:
Awesome thanks!

SBoudrias:
Last thing before merging, could you update the README documentation for this feature? Thanks!

LitoMore:
@SBoudrias Docs updated ;)

SBoudrias:
Thanks! :)

Issue #654, opened 2018-03-05
Input for only numbers?


championshuttler:
is there any input type through that only we can input number?

SBoudrias:
There's no such prompt at the moment. The closest would be the normal input prompt with a validation checking user only inputted numbers.

I'm not against adding such a prompt to the core if someone wants to send over a PR.

karlhorky:
Can this be closed now that #663 was merged?

dkirchhof:
Hello, i was looking for the number prompt feature and found this issue first. You should update the documentation and typings, that this feature was merged.

SBoudrias:
Oh good point. Anyone want to send a PR to update the README?

dkirchhof:
Just made a pull request (https://github.com/SBoudrias/Inquirer.js/pull/786).

SBoudrias:
Thanks for the PR!

We can properly close this issue now :)

PR #663, opened 2018-04-05
Adds a NumberPrompt type for convenience (#654)

CaAlden:
Creates a convenient prompt specifically for getting numbers as input from the user. If the user input's something that is not a number, a default value is returned if it exists or `NaN` if no default was specified.

### DIFF
--- a/lib/prompts/number.js
+++ b/lib/prompts/number.js
@@ -1,0 +1,28 @@
+'use strict';
+/**
+ * `input` type prompt
+ */
+
+var Input = require('./input');
+
+/**
+ * Extention of the Input prompt specifically for use with number inputs.
+ */
+
+class NumberPrompt extends Input {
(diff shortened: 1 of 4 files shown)


SBoudrias:
Thanks!

lib/prompts/number.js
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Extention of the Input prompt specifically for use with number inputs.
+ */
+
+class NumberPrompt extends Input {
+  filterInput(input) {
+    if (input && typeof input === 'string') {
+      input = input.trim();
+      // Match a number in the input
+      let numberMatch = input.match(/(^-?\d+|^\d+\.\d*|^\d*\.\d+)(e\d+)?$/);
```
aleclarson: Is this better than `parseFloat` in some way?

PR #665, opened 2018-04-16
when set answer value in when and return false ignore prompt, user can get the value last.

yujunlong2000:
return that mutated answers array when all when functions return false.
Ref: #517 
Ref: #645


SBoudrias:
Thanks!

Issue #635, opened 2018-01-11
Mask password default


kbrandwijk:
Currently, when you specify a default for a password field, it's shown in plain text. I would like either an option, or by default, to mask the default value as well. So either `(default is hidden)` or `(*********)`, depending on the mask setting.

SBoudrias:
Should we just reject default passwords?

kbrandwijk:
Well, I actually have a use case for a default password, where I don't want to skip the question.

SBoudrias:
Isn't it weird though to have a default password and not being able to see what it is?

pofallon:
Maybe an option to specify a "mask" function to control how the default values are displayed?  In my case I'm prompting the user to confirm or overwrite an existing API key (`input` type) and secret (`password` type).  It would be great to control how the defaults (the existing values) are displayed to the user without affecting the actual default values.  I'd like to display just the first few characters of the API key with an ellipses, while I might display the existing API secret as a series of asterisks.

(Sorry if that's partially off-topic, but to me both cases are just a mask used to control how the default value is shown to the user...)

Thanks for creating such a great node module!

SBoudrias:
@pofallon for this case, I'd just edit the `message` value and wouldn't specify any `default`.

I see a lot of workarounds to most issues here. I don't think hiding the default password is the best default behavior for this library or a blocker to implement your own custom behavior around that.

yvele:
Same use case here.. The user can overwrite the default password (found somewhere)

And I don't want the password to appear in the console logs

SBoudrias:
I'll accept a PR on this issue. Seems there's enough community support to push the needle and hide the default password when provided.

menixator:
It can be a good utility feature to have. Currently, I have a testing setup that heavily relies on inquirer. One of the routines involves logging in, and I prefer to have a fallback password in the environment so that I don't have to type it out every time I run it. Although it's not necessary, it would make the password prompt more complete.

SBoudrias:
Merged on master, I'll try to release a major version in the coming days.

PR #668, opened 2018-04-28
Hide default values for passwords

avb100:
Addressing issue #635, this PR displays `[hidden]` as the default value instead of displaying the actual password, when applicable. The message is consistent with text displayed when the password is masked.


SBoudrias:
Thanks!

PR #671, opened 2018-05-05
Added support for cursor positioning via constants.CARET marker

### LINKED ISSUE #669: Using transformer and moving the caret cursor
Hi there :)

I'm using the `transformer` function to visually append hinted data such as:

```javascript
transformer: (input, answer) => {
  return `${a}.mydomain.com`;
}
```

works as expected, but how could I play with the caret cursor so it's positioned where the user is typing:

out

Full example:

```javascript
var inquirer = require('inquirer');

inquirer.prompt({
  name: 'hostname',
  type: 'input',
  message: 'hostname:',
  transformer: function(a,b) {
    return `${a}.mydomain.com`
  }
});
```

Is it possible?

gentunian:
fixes #669

### DIFF
--- a/lib/prompts/input.js
+++ b/lib/prompts/input.js
@@ -6,6 +6,7 @@
 var chalk = require('chalk');
 var Base = require('./base');
 var observe = require('../utils/events');
+const constants = require('../utils/constants');
 
 class InputPrompt extends Base {
   /**
@@ -42,11 +43,16 @@ class InputPrompt extends Base {
     var bottomContent = '';
     var message = this.getQuestion();
     var transformer = this.opt.transformer;
+    var caretOffset;
 
     if (this.status === 'answered') {
       message += chalk.cyan(this.answer);
     } else if (transformer) {
       message += transformer(this.rl.line, this.answers);
+      if (message.indexOf(constants.CARET) !== -1) {
+        caretOffset = message.slice(0, message.indexOf(constants.CARET));
+      }
+      message = message.replace(constants.CARET, '');
     } else {
       message += this.rl.line;
     }
@@ -55,7 +61,7 @@ class InputPrompt extends Base {
       bottomContent = chalk.red('>> ') + error;
     }
 
-    this.screen.render(message, bottomContent);
+    this.screen.render(message, bottomContent, caretOffset);
   }
 
   /**
--- a/lib/utils/constants.js
+++ b/lib/utils/constants.js
@@ -1,0 +1,7 @@
+/**
+ * Constants
+ */
+module.exports = {
+  // Caret string for cursor positioning
+  CARET: 'aW5xdWlyZXItY2FyZXQK'
+};
--- a/lib/inquirer.js
+++ b/lib/inquirer.js
@@ -1,10 +1,18 @@
 'use strict';
+const constants = require('./utils/constants');
+
 /**
  * Inquirer.js
  * A collection of common interactive command line user interfaces.
  */
 
 var inquirer = module.exports;
+
+/**
+ * Caret export
+ */
+
+inquirer.CARET = constants.CARET;
 
 /**
  * Client interfaces
--- a/lib/utils/screen-manager.js
+++ b/lib/utils/screen-manager.js
@@ -22,7 +22,7 @@ class ScreenManager {
     this.rl = rl;
   }
 
-  render(content, bottomContent) {
+  render(content, bottomContent, changedPrompt) {
     this.rl.output.unmute();
     this.clean(this.extraLinesUnderPrompt);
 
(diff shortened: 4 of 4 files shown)


lib/utils/constants.js
```diff
@@ -0,0 +1,7 @@
+/**
+ * Constants
+ */
+module.exports = {
+  // Caret string for cursor positioning
+  CARET: Math.random().toString(36)
```
SBoudrias: Not sure we should use `Math.random()` here. This would break if the system ends up with many version of inquirer installed and some plugins are relying on another version.

Let's hardcode a random string here instead.

lib/prompts/input.js
```diff
@@ -42,11 +43,14 @@ class InputPrompt extends Base {
     var bottomContent = '';
     var message = this.getQuestion();
     var transformer = this.opt.transformer;
+    var caretOffset;
 
     if (this.status === 'answered') {
       message += chalk.cyan(this.answer);
     } else if (transformer) {
       message += transformer(this.rl.line, this.answers);
+      caretOffset = message.slice(0, message.indexOf(constants.CARET));
```
SBoudrias: When not found, it'll do `slide(0, -1)`. That's probably not what we expected. Maybe `message.slice(0, Math.min(0, message.indexOf(constants.CARET)))` will fix this.

I wonder if that's causing the bug you talked about with regards to the caret position being offset by one.

lib/utils/constants.js
gentunian: good to know, not sure if I completely understand the issue with multiple versions.

lib/prompts/input.js
gentunian: yes, the minor issue I found was what you mentioned. I didn't want to add IF statement for solving as it's a bug that could be ignored and I prefer code styling over solving it. But your solution using `Math.min()` is clever enough to maintain the code legibility.

gentunian:
it would be nice to check why the code is failing:

Codacy/PR Quality Review — Not up to standards.

lib/prompts/input.js
gentunian: reviewing this it won't work with `Math.min()` just because min(0, -1) = -1. I'm working on a solution for it.

gentunian:
@SBoudrias up to standards? "Strings must use doublequote. (quotes)"

I'm used to write angular apps and my linting always says that I should use single quotes (').

lib/prompts/input.js
SBoudrias: `Math.max()` would work then; just provide the wrong function as example :P

SBoudrias:
@gentunian don't worry about Codacy, looks like they once again stopped parsing the eslint config properly... I've disabled the check as it didn't provide value and confused contributors.

lib/utils/constants.js
SBoudrias: To explain the issue with multiple versions, it's because with nested modules systems like Node, there's no guarantee there's a single version of Inquirer installed.

So each version get its own instance. So things like randoms values or `instanceof` checks on classes breaks between different instances. And these issues are really hard to debug on the user side as it's hard to see - and just think about. (Had an issue like this when we first implemented the `inquirer.separator` for lists)

Hope this makes a bit a sense :)

lib/prompts/input.js
gentunian: @SBoudrias it won't work with `Math.max()` either because `"foo-bar".slice(0, 0)` will strip the whole string. The only way it would work if with a function like this:
```javascript
message.slice(0, 
  ImaginaryPackage.positive(
    message.indexOf(constants.CARET), message.length)
)
```

That is, a function that will return the positive number of 2 numbers :)

thetutlage:
Looking for same thing :smile:

gentunian:
@SBoudrias I'm committing with `--amend` just to maintain the commit message. Just let me know if you are ok with it :)

@thetutlage great! check the PR and give us feedback of what you think :+1:

SBoudrias:
@gentunian no issues with `--amend`. Just make sure you force the push otherwise the updates won't be seen here.

gentunian:
@SBoudrias are we ok to merge this PR?

SBoudrias:
@gentunian I thought this branch still had the off-by-one error on the cursor? Is this not the case given we're still slicing with the `indexOf()` result?

gentunian:
@SBoudrias it's fixed with an if statement. For some reason, seems that github can't show the commits pushed with `--force`, but you can see the files here: https://github.com/SBoudrias/Inquirer.js/pull/671/files

And the change in discussion: https://github.com/SBoudrias/Inquirer.js/pull/671/files#diff-a44f44aa53608f57cbfbd206ad281c3d

gentunian:
@SBoudrias anything I could help for this to be merged?

gentunian:
hi @SBoudrias can you guide me on how could be merge this change? :)

oscarmarcelo:
Any plans for this PR? 🙂

martinmcwhorter:
Any plans to merge this? Is there something with this PR that you need another pair of eyes to look at?

PR #673, opened 2018-05-07
Updated external-editor

mrkmg:
- Updated external-editor to version 3.0.0

Updating to use new TypeScript re-written version of External Editor.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "chalk": "^2.0.0",
     "cli-cursor": "^2.1.0",
     "cli-width": "^2.0.0",
-    "external-editor": "^2.1.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -42,7 +42,7 @@
     "chalk": "^2.0.0",
     "cli-cursor": "^2.1.0",
     "cli-width": "^2.0.0",
-    "external-editor": "^2.1.0",
+    "external-editor": "3",
```
SBoudrias: Let's keep standard with the other version def `^3.0.0`

mrkmg:
Wow, didnt realize npm would save the package file like that... I'll fix first thing in the am.

PR #674, opened 2018-05-09
Add 'Boolean' as option type to documentation of default

blewisio:
The Boolean type was omitted from the documentation for the `default` property, but it's a valid type when using `confirm`.


SBoudrias:
Thanks! :)

PR #680, opened 2018-05-23
Add inquirer-fuzzy-path to README

adelsz:
A plugin I wrote when I couldn't find a robust directory selection prompt.
Allows fast file and/or directory selection using fuzzy autocomplete.
Might be useful for someone else.


SBoudrias:
Thanks!

PR #687, opened 2018-06-02
Update to monorepo

SBoudrias:
Updated to a monorepo to be able to work on a major rewrite while still maintaining the current version.

I was worried working on a major rewrite on a branch might makes it hard to incorporate fixes/features introduced to the current package while we work on the rewrite branch.

cc @dwieeb


package.json
```diff
@@ -1,84 +1,27 @@
 {
```
dwieeb: You can add `"private": true` if you like

lerna.json
```diff
@@ -0,0 +1,8 @@
+{
+  "lerna": "2.11.0",
+  "packages": [
+    "packages/*"
```
dwieeb: You can use `"packages/**"` to have packages such as `packages/inquirier/` and `packages/@inquirer/core/` if preferred

lerna.json
```diff
@@ -0,0 +1,8 @@
...
+  ],
+  "version": "0.0.0",
```
dwieeb: Will you be using fixed versioning mode? I don't think that will work with having 6.x inquirer in with 7.x/latest

You can use independent versioning mode though: https://github.com/lerna/lerna#independent-mode---independent

dwieeb:
> I was worried working on a major rewrite on a branch might makes it hard to incorporate fixes/features introduced to the current package while we work on the rewrite branch.

This will likely be a challenge in any case, especially with the fresh release of 6.0 😄 

See my concern w/ fixed versioning mode above. Personally I'd opt for the branch strategy to keep fixed versioning.

lerna.json
SBoudrias: Ah I thought I initialized it that way, but looks like I messed this up. Updated.

Issue #108, opened 2014-04-16
Enter empty value for input with default


bajtos:
Consider the following prompt:

``` js
    { name: 'proxy',
      message: 'HTTP proxy',
      default: 'http://proxy' }

// renders
// [?] HTTP Proxy: (http://proxy)
```

It is not possible to submit an empty value now (meaning "do not use any proxy"), because an empty response is always converted to the default value. See [lib/prompts/input.js#L70](https://github.com/SBoudrias/Inquirer.js/blob/e744e57b19b63f77807c21951afabc479b25b0dc/lib/prompts/input.js#L70).

A workaround is to add a special non-empty token (e.g. `none`) and modify the code handling the answers to replace this token with empty string.

IMHO it would be better to have a solution provided by inquirer, to get a consistent user experience across all CLI tools built on this module.

magland:
For anyone coming across this, here's the solution I am using

```
{
  name: 'proxy',
  message: 'HTTP proxy (use . for no proxy)',
  default: defaultProxy || '.' 
}
```

Then check the answer and replace '.' with undefined or empty string.

PR #700, opened 2018-07-11
Allow empty input by removing default if user enters backspace on empty input

danmactough:
~~Strip surrounding quotes from input value; enable empty values~~

As mentioned a loooong time ago in #108, there's currently no way to input an empty value for an input with a default value.

~~But what if we allowed users to put quotes around their input values the way you would when setting a bash variable? And then the value would be set to the unquoted value.~~

~~In other words, when a user enters `"foo"` the value would be set to `foo` (without quotes).~~

~~This feels like expected behavior to me, and it would enable an empty value for an input with a default value, because `""` would become an empty string.~~

~~Is this too magical or unexpected?~~

~~Would it be better to simply treat a value of `""` or `''` as a special case that should be transformed to an empty string?~~

I think _some_ solution would be preferable to the current recommendation (adding an extra Y/N question).

**UPDATED** in line with [comment](https://github.com/SBoudrias/Inquirer.js/pull/700#issuecomment-405063832) to use `backspace` instead of stripping quotes.

### DIFF
--- a/packages/inquirer/lib/prompts/input.js
+++ b/packages/inquirer/lib/prompts/input.js
@@ -97,7 +97,23 @@ class InputPrompt extends Base {
    * When user press a key
    */
 
-  onKeypress() {
+  onKeypress(event) {
+    const { key } = event;
+
+    // Empty the default when a user clears the input
+    // The `this.hasInput` flag is required to properly detect when the user
+    // has pressed the backspace key while the line is already empty, in which
+    // case we reset the default, which allows a user to supply an empty response.
+    if (key.name === 'backspace' && !this.rl.line) {
+      if (this.hasInput) {
+        this.hasInput = undefined;
+      } else {
+        this.opt.default = undefined;
+      }
+    } else if (this.rl.line && !this.hasInput) {
+      this.hasInput = true;
+    }
+
     this.render();
   }
 }
--- a/packages/inquirer/test/specs/prompts/input.js
+++ b/packages/inquirer/test/specs/prompts/input.js
@@ -98,4 +98,47 @@ describe('`input` prompt', function() {
       done();
     }, 200);
   });
+
+  it('should use the default when the user submits empty input', function() {
+    this.fixture.default = 'DEFAULT';
+
+    var prompt = new Input(this.fixture, this.rl);
+    var promise = prompt.run();
+
+    this.rl.emit('line', '');
+
+    return promise.then(answer => {
+      expect(answer).to.equal('DEFAULT');
+    });
+  });
+
+  it('should use the default when the user enters some input then deletes it', function() {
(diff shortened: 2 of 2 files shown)


SBoudrias:
I think stripping the quotes might have side effet for other users who might need them. I guess you could allow spaces and strip them in the filter function.

Or maybe use `backspace` on empty answer to get rid of the default?

danmactough:
Are you suggesting backspace would clear the default and rerender the prompt? That sounds great!

danmactough:
> Yeah, I think this would be more intuitive and more clearly communicate it with the user.
> 
> Want to send a PR?

I've looked into using backspace, and I don't actually see how we could do it correctly.

It's possible to [intercept the backspace here](https://github.com/SBoudrias/Inquirer.js/blob/master/packages/inquirer/lib/prompts/input.js#L101) and remove the default value, but it's not possible to distinguish whether the user has hit backspace on an _already empty line_ (for which we want to clear the default) or instead has hit backspace _to remove the last previously-typed character to leave the line blank_ (which I don't think should trigger clearing the default.

Maybe an example will clarify.

If we change that `onKeypress` method to:

```js
  onKeypress(event) {
    const { key } = event;
    // Empty the default when a user clears the input
    if (key.name === 'backspace' && !this.rl.line) {
      console.log('');
      console.log('key', key);
      console.log('this.rl.line', JSON.stringify(this.rl.line));
      this.opt.default = undefined;
    }

    this.render();
  }
```

```sh
$ node packages/inquirer/examples/input.js
? What's your first name # hit enter
? What's your last name (Doe) # hit backspace
key { sequence: '',
  name: 'backspace',
  ctrl: false,
  meta: false,
  shift: false }
this.rl.line ""
? What's your last name # ^D

$ node packages/inquirer/examples/input.js
? What's your first name # hit enter
? What's your last name (Doe) a # hit a followed by backspace
key { sequence: '',
  name: 'backspace',
  ctrl: false,
  meta: false,
  shift: false }
this.rl.line ""
? What's your last name # ^D
```

SBoudrias:
I see the problem. I think on keypress, we can check the value of this.line. And if the line is empty then we set a flag like `this.isEmpty` or a relevant name. Then if `this.isEmpty === true && key.name === 'backspace'` we clear the default.

danmactough:
@SBoudrias Ok, I updated this to use the backspace key instead of stripping surrounding quotes. I realized that we want to get this behavior even if the user has made no keypress before the backspace, so I implemented that little state tracking as `hasInput` instead of `isEmpty`.

SBoudrias:
Thanks! The latest version looks great :)

danmactough:
Thanks for your thoughtful feedback, Simon.

orgads:
This works for a text field, but doesn't work for password.

PR #709, opened 2018-08-01
feat: pre-defined answers

### LINKED ISSUE #708: Pre-defined answers
I should be able to pass a hash of pre-defined answers to `prompt` that essentially auto-fills the associated prompts.

The pre-defined answers would still be passed to the `filter`, `transformer`, and `validate` functions. <s>Beforehand, they would be coerced to a string as if they were manually typed into the prompt.</s>

WDYT?

aleclarson:
Closes #708
```js
const {prompt} = require('inquirer');

const answers = await prompt(questions, {
  semver: true,      // confirm
  version: '1.0.0',  // input
  license: 2,        // list (choices index)
  libs: [],          // checkbox
});
```

Pre-defined answers are always passed to the `filter`, `transformer`, and `validate` options.

For `list` and `rawlist` prompts, we look for a choice that matches the pre-defined answer, which may be an index, a choice `name`, or a choice `value`. If no match is found, the prompt is used.

For `checkbox` prompts, we set `checked = true` for any matching choices. The pre-defined answer can be an array of strings or a comma-separated string (eg: `"a,b,c"`). If the pre-defined answer `!== ""` and no matching choices are found, the prompt is used.

For `confirm` prompts, the pre-defined answer can be a boolean or a string (eg: `"yes"` or `"y"`).

For `input`, `password`, `number`, and `editor` prompts, the pre-defined answer is coerced to a string.

When a pre-defined answer is rejected, its associated prompt is used.

Pre-defined answers that have no associated question cause a warning to be printed.

Prompt classes may have a `filterBypass` method that transforms the value of pre-defined answers.

### Changes
- make `prompt` accept an answers object as its 2nd argument
- add default behavior to `Prompt#onEnd` and `Prompt#onError`
- throw error if `Prompt#_run` is not overridden
- rename `Prompt#handleSubmitEvents` to `submit`
- hook up `filterInput`, `onEnd`, and `onError` in `Prompt#submit`
- support `filterBypass` method on Prompt objects
- set `done` property of Prompt objects before calling `_run`
- stop passing `done` callback to `Prompt#_run`
- rename `ExpandPrompt#getCurrentValue` to `filterInput`
- rename `onSubmit` methods to `onEnd`
- add `Choices#findIndex` method

### DIFF
--- a/packages/inquirer/lib/inquirer.js
+++ b/packages/inquirer/lib/inquirer.js
@@ -21,11 +21,13 @@ inquirer.ui = {
 
 /**
  * Create a new self-contained prompt module.
+ * @param {Object} opt Readline options
+ * @return {Function}
  */
 inquirer.createPromptModule = function(opt) {
-  var promptModule = function(questions) {
+  var promptModule = function(questions, answers) {
     var ui = new inquirer.ui.Prompt(promptModule.prompts, opt);
-    var promise = ui.run(questions);
+    var promise = ui.run(questions, answers);
 
     // Monkey patch the UI on the promise object so
     // that it remains publicly accessible.
--- a/packages/inquirer/lib/objects/choices.js
+++ b/packages/inquirer/lib/objects/choices.js
@@ -105,6 +105,14 @@ module.exports = class Choices {
     return _.find(this.choices, func);
   }
 
+  findIndex(func) {
+    var length = this.length;
+    for (var i = 0; i < length; i++) {
+      if (func(this.choices[i])) return i;
+    }
+    return -1;
+  }
+
   push() {
     var objs = _.map(arguments, val => new Choice(val));
     this.choices.push.apply(this.choices, objs);
--- a/packages/inquirer/lib/prompts/base.js
+++ b/packages/inquirer/lib/prompts/base.js
@@ -58,20 +58,26 @@ class Prompt {
       this.done = resolve;
 
       var val = this.answers[this.opt.name];
-      if (this.filterBypass) {
-        val = this.filterBypass(val);
-      }
-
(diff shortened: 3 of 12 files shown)


karlhorky:
Thanks for taking a shot at this Alec!

I suppose this doesn't allow for dynamically setting answers during the prompt sequence as I mentioned in https://github.com/SBoudrias/Inquirer.js/issues/517#issuecomment-366711463 ?

Would you consider an alternative API that would allow for this? Perhaps accepting callbacks with the answers object as a parameter, in addition to the static values?

aleclarson:
@karlhorky That's a good idea. I'll see what I can do. Something like this, right?

```js
const answers = await prompt(questions, {
  foo: true,
  bar(answers) {
    if (answers.foo) {
      // Promises are supported
      return fetch(url);
    }
    // Return undefined to trigger the prompt.
  }
});
```

aleclarson:
@karlhorky Okay, give it a try. It works great! :)

cb4750be1470f1df2df98795261d95d10f763334

karlhorky:
Wow, awesome, your example looks great! 🎉

Does this run in the sequence of the prompts? Eg. if the `foo` in your example was user-entered at runtime, does it still go into the block with the `fetch`?

aleclarson:
@karlhorky Yup

aleclarson:
@SBoudrias Any thoughts on this? 😄

arcanis:
What's this PR status?

aleclarson:
@arcanis I'm using this PR without issue. I think the maintainer won't merge it because (1) it has no tests and (b) it's not implemented in the new core.

PR #720, opened 2018-08-27
add keyboard shortcut for rawlist selection

### LINKED ISSUE #699: keyboard shortcut for list selection ?
Hi,
It would be great to add the possibility to select a choice in a list through a keyboard shortcut, for example :
```
$ node list.js
? What do you want to do? (Use arrow keys)
❯ [O]rder a pizza 
  Make a [r]eservation 
  ──────────────
  Ask for opening [h]ours 
  - Contact support (Unavailable at this time)
  Talk to the receptionist 
```
Typing "o" would select the first choice (order a pizza), typing "r" the second choice (make a reservation), and so on.

Could be useful in the case of a long list, no need to use the arrow keys...

phantom0424:


phantom0424:
This is my first time contributing code to the open source library. ^_^

packages/inquirer/lib/prompts/rawlist.js
```diff
@@ -146,6 +150,35 @@ class RawListPrompt extends Base {
...
     this.render();
   }
+
+  /**
+   * When user press up key
+   */
+
+  onUpKey() {
+    if (this.selected) {
+      this.selected--;
```
SBoudrias: We'll need to also account for the selected index going under or over the list length.

packages/inquirer/lib/prompts/rawlist.js
```diff
@@ -146,6 +150,35 @@ class RawListPrompt extends Base {
...
+    } else {
+      this.selected = this.opt.choices.realLength - 1;
+    }
+    this.rl.line = (this.selected + 1).toString();
```
SBoudrias: Is reassigning the `rl.line` necessary? I don't think it'll be in this case.

SBoudrias:
Hey @phantom0424, thanks for the PR. Always happy to welcome new contributors!

packages/inquirer/lib/prompts/rawlist.js
phantom0424: I think the `rl.line` value should correspond to the selected index, otherwise it is not very strange?

phantom0424:
That error job is because it has timed out.

SBoudrias:
@phantom0424 I think you closed this PR by mistake right?

phantom0424:
Yes, I found that I referenced to a mismatched issue #699  .

SBoudrias:
Thanks!

Issue #711, opened 2018-08-02
Keep the answer if validation fails


garyking:
If validation fails on an input, then the answer is erased. Please give us an option to keep the answer even if validation fails, because sometimes just one character needs to be changed in the answer to make it pass validation.

ZYSzys:
Yeah, it seems to be more reasonable indeed.

SBoudrias:
Fixed in #825

PR #725, opened 2018-09-01
[issue #711] Keep the answer if validation fails

dangen-effy:


packages/inquirer/lib/prompts/input.js
```diff
@@ -89,7 +89,9 @@ class InputPrompt extends Base {
     this.done(state.value);
   }
 
-  onError(state) {
+  onError(state) { 
+    this.rl.line += state.value;
```
SBoudrias: Should this just be `=`?

SBoudrias:
Hey, thanks for sending this PR!

The linting is failing on Travis, could you fix that? Once CI pass, I think we'll be good to merge.

packages/inquirer/lib/prompts/input.js
dangen-effy: I'll modify it  as `+=`

Thank you for review.

dangen-effy: Actually, `+=` `=` are doesnt matter. Because `this.rl.cursor` is set to `0` on error.
But for readability using `+=` is more reasonable i think.

SBoudrias: I'm not sure it's more readable, what we want is to keep the answer on the line. Having `+=` is confusing because if the line wasn't cleaned and cursor reset to 0, we'd end up duplicating the answer. Without knowing the internal of how readline works, this would be quite confusing.

austen-herbst:
> Hey, thanks for sending this PR!
> 
> The linting is failing on Travis, could you fix that? Once CI pass, I think we'll be good to merge.

@dangen-effy The reason this is failing Travis is because of a trailing space that was added to line 93 after the ``{``. If you remove that, the ESLint validation will pass.

PR #728, opened 2018-09-14
disable lock files

dwieeb:
Lock files are not super useful for developing libraries. They don't enforce versioning of transitive dependencies when the library is installed and they create unnecessary noise in commit history.

@SBoudrias Let me know your thoughts on this one. It comes down to opinion, but my reasoning is above.


dwieeb:
Looks like Travis runs `npm ci` though.

LitoMore:
BTW, I think we can add `package-lock.json*` and `yarn.lock*` to the `.gitignore` file.

bokub:
From the [npm documentation](https://docs.npmjs.com/files/package-lock.json):

> This file is intended to be committed into source repositories

PR #738, opened 2018-10-04
Add README file to 'inquirer' npm package

bokub:
Right now, the npm `README` of inquirer is empty, so it looks like you're looking at the wrong module

image

I've added two npm scripts:
- Just before publishing, the `README` from `./` is copied to `./packages/inquirer`
- Just after publishing, that file is deleted

That should fix the problem


SBoudrias:
Thanks for scripting the fix! I'll republish when I'm back from vacation.

Issue #769, opened 2019-01-26
In rawlist, the answers go in negative while moving up 


sarthology:
In the type `rawlist`, when the key is used to choose value the answers go in negative, if you up and more than the total choices if you go down.

Is this a bug or I'm missing something?

```
inquirer.prompt([{
            type: 'rawlist',
            name: 'title',
            message: 'Select the story 🤓',
            choices: ['👻','🤓','👩🏻‍💻','👨🏻‍💻','🦄','💻','👾','👽'],
            paginated: true
        }])
        .then(answers => {
            openLink(answers);
        });
```
![](https://media.giphy.com/media/24FMrDc1eOnQmW35de/giphy.gif)

sarthology:
Ohk on it.

LitoMore:
Hi @sarthology We could close this issue if it already solved.

sarthology:
Yeah it's solved from my side.

PR #774, opened 2019-01-30
Rawlist bug #769 fixed

sarthology:
Bug #769 is fixed.  😊


SBoudrias:
Thanks!

I'll try to prepare a release during my evening as time permit.

sarthology:
Wait, there is one more bug. I'm fixing it too.

Issue #575, opened 2017-08-10
For a question of type `checkboxes`, a checkbox item that is both `checked` and `disabled` should be included in the answers


odolha:
In this function:
https://github.com/SBoudrias/Inquirer.js/blob/815690c79cce2de1ff56be0ecebf226f972024a5/lib/prompts/checkbox.js#L130-L137
if a choice is `disabled` it will never be part of the question's answers, even if it is also `checked`.

This is inconsistent with common behavior. Checkboxes can be both checked and disabled.

I'm surprised I didn't find anything related to this because I think it is quite a common use case. The case is when you want to have some choices constrained by previous answers (but you want to make this clear to the user).

BTW: the rendering in such cases is also not very nice (no difference between checked && disabled and only disabled) but this can be worked around because of the message we can provide in `disabled`, e.g. using something like "Required" vs. "Invalid option"

SBoudrias:
Hey, thanks for opening this issue. Looks like a valid use case.

Given you already figured out the solution, would you mind sending a PR? Feel free to also explore UI options.

PR #779, opened 2019-02-04
Allow disabled checkboxes to still be counted in answers if they're checked programatically

jonscottclark:
This PR is related to #575.

A new option for a question is introduced called `includeDisabled`. By default, this option is `false`, to keep the current behaviour as the default (disabled choices are not passed through to answers.

@SBoudrias Let me know if there's anything else I can contribute to this PR; looking forward to your feedback!

For disabled choices that are marked as `{checked: true}`, their radio buttons will be gray instead of green.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,7 @@ A question object is a `hash` containing question related values:
 - **pageSize**: (Number) Change the number of lines that will be rendered when using `list`, `rawList`, `expand` or `checkbox`.
 - **prefix**: (String) Change the default _prefix_ message.
 - **suffix**: (String) Change the default _suffix_ message.
+- **includeDisabled**: (Boolean) In returned answers, include choices that are `{disabled: true}` as well as `{checked: true}`. Defaults: `false`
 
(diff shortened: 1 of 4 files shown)


jonscottclark:
@SBoudrias Working on getting these tests to pass. I was getting errors locally when running tests that prevented me from getting much further - the ones below, if they look familiar to you? I ran `npm install` before running `npm run test`.

```
FAIL  packages/input/test.js
  ● Test suite failed to run

    Cannot find module '@inquirer/core' from 'index.js'

    > 1 | const { createPrompt } = require('@inquirer/core');

FAIL  packages/core/test.js
  ● Test suite failed to run

    Cannot find module 'cli-spinners' from 'index.js'

    > 6 | const spinner = require('cli-spinners').dots;
```

SBoudrias:
You probably need to run yarn lerna bootstrap to get the tests to pass

jonscottclark:
Thanks, @SBoudrias, I will try that.

PR #780, opened 2019-02-06
Updated README.md with a reactive example.

ambiguities:
The example provided may milead readers as it will not work in the latest RxJS, version 6 at the time of writing. The new example works in RxJS 5.


ambiguities:
-.- I should have looked over my detailed message...

Relequestual:
As of now, package.json specifies: `"rxjs": "^7.2.0"`.
The example is still broken as far as I can tell though. -_-

PR #796, opened 2019-04-11
Add number support for choices prompt

### LINKED ISSUE #765: chore: better error message handling for `choices`
right now if `choices` contains a number, e.g. `choices: [ 100, 90, 80 ]` then the following error message appears (which isn't obvious to the user that `choices` can only be a String or Object (per the README)

```sh
(node:52503) UnhandledPromiseRejectionWarning: TypeError: Cannot use 'in' operator to search for 'value' in 100
    at new module.exports (/Users/me/.nvm/versions/node/v8.12.0/lib/node_modules/lass/node_modules/inquirer/lib/objects/choice.js:25:22)
```

LitoMore:
Fix #765


SBoudrias:
Thanks!

PR #801, opened 2019-05-15
Remove paginated option from example

jedmao:
This `paginated` option only exists in this example, but nowhere in the code. It was removed in https://github.com/SBoudrias/Inquirer.js/pull/55.


SBoudrias:
Thanks!

Issue #805, opened 2019-06-14
prompt.ui.close() does not remove keypress listeners


jared-hexagon:
## Issue

When closing existing prompts with this code:

```js
const { prompt } = require('inquirer')
const myPrompt = prompt({ ... })
myPrompt.ui.close()
```

does not unbind keypress events which can cause a `MaxListenersExceededWarning` warning when prompting a bunch of times.

## Solution

In [baseUI.js](https://github.com/SBoudrias/Inquirer.js/blob/master/packages/inquirer/lib/ui/baseUI.js) we probably need to unbind all of the events that we bound. Calling `rl.close()` does not automatically unbind them.

## Repro

Calling `prompt.ui.close()` many times will cause the warning to appear:

https://github.com/jared-hexagon/inquirer-memory-leak-repro

PR #808, opened 2019-06-17
Remove all event listeners on close to avoid memory leak

jared-hexagon:
## Summary

Fix for #805 by removing all event listeners on `prompt.ui.close()`.

## Testing

In the repro (https://github.com/jared-hexagon/inquirer-memory-leak-repro) I `npm link`ed my fork and the `MaxListenersExceededWarning` warning never shows up.

### DIFF
--- a/packages/inquirer/lib/prompts/base.js
+++ b/packages/inquirer/lib/prompts/base.js
@@ -78,6 +78,7 @@ class Prompt {
    */
   close() {
     this.screen.releaseCursor();
+    this.rl.input.removeAllListeners();
   }
 
   /**


SBoudrias:
I'm not too sure about this change.

The problem is that we'll end up removing ALL listeners; even those that might've been setup outside of the Inquirer code base.

I'd prefer to specifically remove the listeners we're adding.

jared-hexagon:
> The problem is that we'll end up removing ALL listeners; even those that might've been setup outside of the Inquirer code base.

It only removes listeners for a single event emitter. If we create the emitter we should be safe to remove listeners for it (which we do in `baseUI.js` on line 15). Otherwise I think it will be a pain in the ass to remove each listener because of the layers going on.

SBoudrias:
Yeah for sure won't be easy... Let's try this out, and if no one raises issues, we should be good to go.

sodatea:
https://github.com/SBoudrias/Inquirer.js/issues/811
So this is causing bugs. Please revert.

chrisdothtml:
In light of #811 I looked into this, and from what I can tell, `this.rl` is shared across all prompts. The readline instance is initially setup in [lib/ui/baseUI.js:15](https://github.com/SBoudrias/Inquirer.js/blob/01c2999b3ab8c35018daf40c78d6969ed79f6acf/packages/inquirer/lib/ui/baseUI.js#L15), and the same instance (`this.rl`) is passed into all subsequent prompt instances created at [lib/ui/prompt.js:85](https://github.com/SBoudrias/Inquirer.js/blob/01c2999b3ab8c35018daf40c78d6969ed79f6acf/packages/inquirer/lib/ui/prompt.js#L85). **So calling `removeAllListeners` removes listeners for all future prompts as well as the instance being closed.**

> I'm not too sure about this change...
> ...
>
> Let's try this out, and if no one raises issues, we should be good to go

@SBoudrias in retrospect, perhaps a bit more scrutiny should be put on changes like this considering this package has 12mil+ weekly downloads

jared-hexagon:
@chrisdothtml 

big_yikes

PR #821, opened 2019-07-07
Clear password field on backspace

### LINKED ISSUE #820: It's not possible to clear a password field with default
Similar to #108 which was solved long ago, a password field cannot be cleared.

orgads:
Fixes #820.


SBoudrias:
Thanks!! I'll prepare a release soon.

PR #835, opened 2019-08-08
close keypress stream when rl close

pgherveou:
In our CLI we capture SIGINT events and cancel any active prompt when this happens.
I noticed that when we do that, Inquirer leak the prompts, 
This happens because `rl` event streams (created when calling `observe(this.rl)`) never complete, thus causing underlying streams not to complete as well.


SBoudrias:
Good find thanks!

pgherveou:
Thanks Simon! Would you mind publishing the patch on npm?

Issue #861, opened 2019-11-20
rawlist: move up and down do not work properly with multi rawlist


hk029:
When you use multiple rawlist, there will be a mess up in the second rawlist. As you can see in the gifs below, when you press `Down` key 3 times, and `Up` key 1 times, the current selection should be the third one, but it turn out to the first one.

Here is the gifs
![](https://p1.music.126.net/xqHm2FPX0SQjvWDCQ8IK7Q==/109951164500081208.gif)

![](https://p1.music.126.net/915lZkKCDBxbh9krCOszZA==/109951164500075471.gif)

Here is my code
```
const { prompt } = require('inquirer');

const questions = [
    {
        type: 'rawlist',
        name: 'application',
        message: 'first question',
        default: 'web',
        choices: ['a', 'b', 'c'],
    },
    {
        type: 'rawlist',
        name: 'prefix',
        message: 'second question',
        default: 'music-',
        choices: ['aa', 'bb', 'cc', 'dd'],
    },
];

async function main() {
    const params = await prompt(questions);
    console.log(params);
}
```

Environment | Info
-- | --
inqquirer | 7.0.0
System | Mac OS X 10.15.1

ovidiu141:
Shouldn't this issue be closed?

PR #863, opened 2019-11-22
fix multi rawList bug (#861)

xiaoxielnino:
The selected choice should depend on `this.selected` value instead on the `this.rl.line` value.
And we render the 'correct' index of the choices on the screen by "increase 1" of the `this.selected` value.

The test of rawList should change correspondingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,11 +2,12 @@
 
 # Inquirer.js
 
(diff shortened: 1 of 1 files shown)


xiaoxielnino:
Although this pull request can pass all the tests. However, the selected index of multi rawlist is NOT correct. So, I Think we should change the tests of rawList  a little bit. Please refer to my first pull request

SBoudrias:
Hey, thanks for sending this PR. It looks like you did identified the bug properly, but your code is removing a few features.

Can you clean up the branch to re-add the features that doesn't need to be removed. And could you also fix the prettier warning on the CI. Thanks!

xiaoxielnino:
Sorry about that removing some features. And I have re-added the features that doesn't need to be removed. And  also fix the prettier warning on the CI. please take a check. Thanks.

PR #865, opened 2019-11-25
fix couldn‘t re input when input an invalid number

lvjiaxuan:
NaN as number type, I can not clear it on common line.So I replace with string type.Any better way?


klgd:
how to fix? i have same problem

lvjiaxuan:
> how to fix? i have same problem

别用number类型，用string。

PR #867, opened 2019-11-26
fix multi rawList bug (#861)

xiaoxielnino:
Sorry about that removing some features. And I have re-added the features that doesn't need to be removed. And also fix the prettier warning on the CI. please take a check. Thanks.

### DIFF
--- a/packages/inquirer/lib/prompts/rawlist.js
+++ b/packages/inquirer/lib/prompts/rawlist.js
@@ -64,14 +64,13 @@ class RawListPrompt extends Base {
     validation.success.forEach(this.onEnd.bind(this));
     validation.error.forEach(this.onError.bind(this));
 
-    events.keypress
-      .pipe(takeUntil(validation.success))
-      .forEach(this.onKeypress.bind(this));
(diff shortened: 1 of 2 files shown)


xiaoxielnino:
I've seen many codecov/patch weird behavior compared to codecov/project which was always properly reporting the coverage as expected.
And I don't know how the pass the codecov/patch, if there is an example for codecov/patch, please help me out.
Unless someone thinks codecov/patch is needed for a good reason, I think keeping codecov/project only will be enough.

xiaoxielnino:
> I've seen many codecov/patch weird behavior compared to codecov/project which was always properly reporting the coverage as expected.
> And I don't know how the pass the codecov/patch, if there is an example for codecov/patch, please help me out.
> Unless someone thinks codecov/patch is needed for a good reason, I think keeping codecov/project only will be enough.

It seems that I should add the unit test to pass the ci. Please have a check. Thanks.

SBoudrias:
Thanks for the bug fix, it appears to be working fine :+1:

PR #869, opened 2019-11-29
Bugfix: Choices.push() breaks index if a disabled item was already in the Choices

bibli-alex:
This fixes a bug in the following scenario:
- a new Choices is initialised, with one of the items being `disabled: true`
- `choices.push('item')` is used
In this situation, the realChoices/realLength of the Choices becomes incorrect, as disabled items weren't being filtered out of the list in the `push` function, only in the `constructor`. 

I've also added the necessary changes to the existing test.


SBoudrias:
Thanks for sending the bug fix!

Issue #767, opened 2019-01-19
Multiple Questions on Windows 10 Hangs


jlkalberer:
```
Windows 10 - Version 1803 (OS Build 17134.523)
yarn - 1.9.4
npm - 6.0.1
node - v10.13.0
```

I dug into this a bit because of https://github.com/Microsoft/AppCenter-SDK-React-Native/issues/471#event-2067227003

Inquirer uses `stdout` to write to the command prompt. It looks like [if this line is run](https://github.com/SBoudrias/Inquirer.js/blob/master/packages/inquirer/lib/ui/baseUI.js#L55), it will cause windows to "hang". 
*Also, it turns out that node 8.15.0 works with this code so it could also just be a regression*

I think what is really going on is that the stream is just ending so we don't see any more output.

Here is a repro of the bug.

```
const readline = require("readline");

let index = 0;
const runQuestion = () => {
  index += 1;
  console.log(`Question ${index} - type something`);

  const rl = readline.createInterface({
    input: process.stdin,
    output: process.stdout
  });

  rl.on("line", line => {
    console.log(`Received: ${line}`);

    rl.output.end();
    rl.pause();
    rl.close();

    runQuestion();
  });
};

runQuestion();
```

With this simple example removing `rl.output.end()` fixes things but the only way to work around it in `inquirer` is to remove the `rl.close()` line.

mshima:
@SBoudrias 
I've tracked back https://github.com/yeoman/generator/issues/1098 to this issue.
It's reported in a lot of places, node.js included (https://github.com/nodejs/node/issues/21210https://github.com/nodejs/node/issues/21210)

Removing `rl.close();` fix the issue, but I don't know the consequences.
Maybe make it conditional for non windows.

mshima:
Found https://github.com/porketta/Inquirer.js/commit/91b4216d67bed610cc941c953a622be79ed34d3d at https://github.com/nodejs/node/issues/21771

Diablohu:
After this fix, there are more problems

1. The process that run _Inquirer.js_ will never end, make the program hang until killed manually
2. The option of type `list` or `checkbox` cannot be selected. When input UP or DOWN, option will not be switched, instead the last input value showed up
![1](https://user-images.githubusercontent.com/1946774/71814561-bf02c680-30b7-11ea-9f28-258824e76013.gif)

mshima:
Seems this have to be reverted.
@Diablohu what is your environment?

[...]

PR #879, opened 2020-01-04
Workaround node.js bug that closes main input and output stream.

mshima:
Fixes https://github.com/SBoudrias/Inquirer.js/issues/767.


SBoudrias:
Ah that's great! Thanks!

PR #884, opened 2020-01-14
Add inquirer-table-prompt plugin

eduardoboucas:
Available at https://github.com/eduardoboucas/inquirer-table-prompt.


SBoudrias:
Thanks!

PR #885, opened 2020-01-22
Handle questions as a plain object {[name]: question}

caub:
I think it's worth to support a questions object, like this:
```js
const { foo, bar } = await inquirer.prompt({
  foo: {
    message: '...',
    default: '...',
  },
  bar: {
    default: '...',
  }
}):
```
it looks better imo, `{ foo, bar }` visually matches the corresponding keys in the question object

and object literals keys are ordered (as long as they're not numeric)

This PR detects such an object, safely normally, since a question should contain a `name` prop, there can't be a conflict with a single question object

### DIFF
--- a/packages/inquirer/lib/ui/prompt.js
+++ b/packages/inquirer/lib/ui/prompt.js
@@ -32,7 +32,12 @@ class PromptUI extends Base {
 
     // Make sure questions is an array.
     if (_.isPlainObject(questions)) {
-      questions = [questions];
+      // It's either an object of questions or a single question
+      questions = Object.values(questions).every(
+        (v) => _.isPlainObject(v) && !_.has(v, 'name')
+      )
+        ? Object.entries(questions).map(([name, question]) => ({ name, ...question }))
+        : [questions];
     }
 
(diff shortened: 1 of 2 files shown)


caub:
@SBoudrias Do you think we could do that?

codinghusi:
Hey, will this be merged one day?

caub:
@codinghusi I gave up, I'm using http://npm.im/prompt

codinghusi:
Oh nice, there's another package for that :)
_(But doesn't have that nice highlighting and stuff 🙃 )
Btw. I'm currently working on my own wrapper for inquirer.js where this feature here will be also included._

May I ask, what kind of problem you had with merging? As I could see, it had only problems with code formatting guidelines.

caub:
We use prompt package because it's light and we don't need fancy highlighting

Ok I fixed the eslint issues, I don't have more time to spend on this sorry, it'd probably need tests

SBoudrias:
Thanks for the PR. I think that's a good idea to reduce boilerplate a little.

PR #891, opened 2020-02-14
Produce error on prompt in non-tty environment.

### LINKED ISSUE #495: prompt() fails silently if stdin isn't a TTY
When run with `node test.js < /dev/null`, this script exits with code 0 and without logging anything:

```javascript
var inquirer = require('inquirer');

inquirer.prompt([{name: 'username', message: 'username'}])
.then(
    function (answers) {
      console.log('Answers:');
      console.log(answers);
  },
  function (err) {
    console.log('Error:');
    console.log(err.stack);
    process.exit(1);
  }
);
```

This can cause some weird problems in CI environments, where scripts succeed without doing anything.

I encountered this with inquirer 3.0.1 on Node.js 6.9.1.

jhorbulyk:
Fixes #495

### DIFF
--- a/packages/inquirer/lib/inquirer.js
+++ b/packages/inquirer/lib/inquirer.js
@@ -27,13 +27,11 @@ inquirer.createPromptModule = function(opt) {
     // Check if prompt is being called in TTY environment
     // If it isn't return a failed promise
     if (!process.stdin.isTTY) {
-      return new Promise(function(resolve, reject) {
-        const nonTtyError = new Error(
-          'Prompts can not be meaningfully rendered in non-TTY environments'
-        );
-        nonTtyError.isTtyError = true;
-        reject(nonTtyError);
-      });
+      const nonTtyError = new Error(
+        'Prompts can not be meaningfully rendered in non-TTY environments'
+      );
+      nonTtyError.isTtyError = true;
+      return Promise.reject(nonTtyError);
     }
 
     var ui = new inquirer.ui.Prompt(promptModule.prompts, opt);
--- a/packages/inquirer/lib/inquirer.js
+++ b/packages/inquirer/lib/inquirer.js
@@ -24,6 +24,18 @@ inquirer.ui = {
  */
 inquirer.createPromptModule = function(opt) {
   var promptModule = function(questions) {
+    // Check if prompt is being called in TTY environment
+    // If it isn't return a failed promise
+    if (!process.stdin.isTTY) {
+      return new Promise(function(resolve, reject) {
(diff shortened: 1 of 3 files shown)


packages/inquirer/lib/inquirer.js
```diff
@@ -24,6 +24,18 @@ inquirer.ui = {
  */
 inquirer.createPromptModule = function(opt) {
   var promptModule = function(questions) {
+    // Check if prompt is being called in TTY environment
+    // If it isn't return a failed promise
+    if (!process.stdin.isTTY) {
+      return new Promise(function(resolve, reject) {
```
SBoudrias: Let's replace with `return Promise.reject(error)`

SBoudrias:
The rest looks good to me, we should be able to merge and release quickly.

jhorbulyk:
@SBoudrias Updated

SBoudrias:
Perfect thanks! I'll make a new release soon.

PetrSnobelt:
Hi, this MR causes the problem when used with `when` property, for example:
```
responses = await inquirer.prompt([{
          name: 'name',
          message: 'give a name to your app',
          default: '@kiwicom/fe-app',
          when: () => !flags.name,
        }])
```
We provide name in flags on CI, but it break our CI with "Error: Prompts can not be meaningfully rendered in non-TTY environments"

jhorbulyk:
@PetrSnobelt Fixed in https://github.com/SBoudrias/Inquirer.js/pull/896

ruyadorno:
Hi @SBoudrias and @jhorbulyk sorry but unfortunately this change introduces a regression since `stdin` can be configured to something else other than just `process.stdin`: 

https://github.com/SBoudrias/Inquirer.js/blob/2c521b70d45073a9a4d50ab3ae9481fd6104c1a4/packages/inquirer/lib/ui/baseUI.js#L70

it completely broke [ipt](https://github.com/ruyadorno/ipt) in your latest patch release since it actually leverages the configurable `stdin` in order to provide the functionality there 😭

dan-barrett:
I'm somewhat confused as to where this issue stands. Looks like it was partly reverted in a future commit?

Can I explicitly configure inquirer to fail in non-tty environments or is that no longer possible? Having hard time finding documentation about it. Thanks!

jhorbulyk:
> I'm somewhat confused as to where this issue stands. 

My understanding is that after the latest merged PRs, inquirer will fail in non-tty environments but only if:
* inquirer hasn't been configured to consume something other than `stdin`
* `when()` functions still cause the prompt to be rendered

PR #896, opened 2020-03-02
Fix bug with non-tty checks and when.

jhorbulyk:
Fixes problem identified here: https://github.com/SBoudrias/Inquirer.js/pull/891#issuecomment-593304435 by moving TTY check to immediately before question is asked.


SBoudrias:
Thanks!

dotansimha:
Is there a way to bypass the new behaviour? 
It breaks for me in https://github.com/dotansimha/graphql-code-generator/pull/3571 
We are running it in Jest test environment, and it seems like `isTTY=false` there (we have a mock for the inputs, and it worked before, but now it's failing because of this new check)

PR #898, opened 2020-03-08
Fixes tty error on custom input streams

ruyadorno:
The implementation which introduces thrown errors when using non-tty environments didn't took into account the fact that a different input stream can be provided via config.

This changeset fixes it by introducing the appropriate checks and tests to ensure no further regressions.

/cc @SBoudrias @jhorbulyk 
ref: #891

### DIFF
--- a/packages/inquirer/lib/ui/prompt.js
+++ b/packages/inquirer/lib/ui/prompt.js
@@ -83,7 +83,10 @@ class PromptUI extends Base {
   fetchAnswer(question) {
     // Check if prompt is being called in TTY environment
     // If it isn't return a failed promise
-    if (!process.stdin.isTTY) {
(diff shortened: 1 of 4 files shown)


mshima:
This PR doesn’t fix yeoman-test regression https://github.com/yeoman/yeoman-test/issues/78.
Should me moved to inside prompt (base) run.
Yeoman-test creates a custom prompt.

mshima:
@ruyadorno take a look at https://github.com/SBoudrias/Inquirer.js/pull/903

ruyadorno:
@mshima 👍 thanks, I prefer your approach of bringing the check up a level to `baseUI.js` and making the new tty error an opt-in thing rather than the default, it's def a breaking change for folks consuming these APIs and should not be in v7

I'll close this PR so we can focus on #903

PR #901, opened 2020-03-08
Add support to pre-filed answers.

mshima:
Add a parameter for a pre-filed answers.

Use case:
Allow to easily skip already answered question, but allowing following questions to use the skipped answers.

### DIFF
--- a/packages/inquirer/lib/ui/prompt.js
+++ b/packages/inquirer/lib/ui/prompt.js
@@ -16,9 +16,13 @@ class PromptUI extends Base {
     this.prompts = prompts;
   }
 
-  run(questions) {
+  run(questions, answers) {
     // Keep global reference to the answers
-    this.answers = {};
+    if (_.isPlainObject(answers)) {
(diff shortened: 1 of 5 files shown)


packages/inquirer/lib/ui/prompt.js
```diff
@@ -16,9 +16,9 @@ class PromptUI extends Base {
     this.prompts = prompts;
   }
 
-  run(questions) {
+  run(questions, answers) {
     // Keep global reference to the answers
-    this.answers = {};
+    this.answers = answers || {};
```
SBoudrias: We'll want to clone the `answers` object to make sure we're not mutating outside values.

mshima: Lodash? Or {...answers} is enough?

mshima:
@SBoudrias i found out that if the first question contains when with function, then the promise is never returned on test cases. Don’t tested on real cases.

Issue #902, opened 2020-03-09
Produce error on prompt in non-tty environment.


mshima:
Issue to track non-tty environment.

Refs:
https://github.com/SBoudrias/Inquirer.js/pull/891
https://github.com/SBoudrias/Inquirer.js/pull/896
https://github.com/SBoudrias/Inquirer.js/pull/898

PR #903, opened 2020-03-09
Revert tty behavior for inquirer 7.x

mshima:
- Move tty check to setupReadlineOptions
- Add skipTTYChecks
- Enable skipTTYChecks by default for 7.x behavior stability.

Fixes https://github.com/SBoudrias/Inquirer.js/issues/902.

### DIFF
--- a/packages/inquirer/lib/ui/baseUI.js
+++ b/packages/inquirer/lib/ui/baseUI.js
@@ -65,13 +65,16 @@ class UI {
 
 function setupReadlineOptions(opt) {
   opt = opt || {};
+  // Inquirer 8.x:
+  // opt.skipTTYChecks = opt.skipTTYChecks === undefined ? opt.input !== undefined : opt.skipTTYChecks;
+  opt.skipTTYChecks = opt.skipTTYChecks === undefined ? true : opt.skipTTYChecks;
 
   // Default `input` to stdin
   var input = opt.input || process.stdin;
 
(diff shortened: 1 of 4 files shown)


packages/inquirer/lib/ui/baseUI.js
```diff
@@ -65,10 +65,23 @@ class UI {
 
 function setupReadlineOptions(opt) {
   opt = opt || {};
+  // Inquirer 8.x:
+  // opt.skipTTYChecks = opt.skipTTYChecks === undefined ? opt.input !== undefined : opt.skipTTYChecks;
```
ruyadorno: personally I wouldn't bother documenting/commenting this, maybe the folks maintaining the package will just rather have `skipTTYChecks=true` as default

mshima: I think this check is valid and maybe deserves a 8.x version together with 7.0.7.

ruyadorno:
@mshima properly sent you the test here: https://github.com/mshima/Inquirer.js/pull/1

packages/inquirer/lib/ui/baseUI.js
ruyadorno: fair enough 👍

SBoudrias:
Thanks to both of you for the help on this issue :D

SBoudrias:
`7.0.7` is out now.

PR #905, opened 2020-03-13
Add `description` option to choices

spalladino:
Adds a `long` option to a choice in a select prompt. If the option is set, the text in `long` for the currently selected option is shown as help at the bottom of the list window. This is particularly useful for giving more info to the user on the option being selected.

Screenshot from 2020-03-13 20-16-30

```js
select({
    message: 'Select a package manager',
    choices: [
      { name: 'npm', value: 'npm', long: 'npm is the most popular package manager' },
      { name: 'yarn', value: 'yarn', long: 'yarn is an awesome package manager' },
      { name: 'jspm', value: 'jspm', disabled: true }
    ]
  });
```


heynext:
it's very useful for me! look forward to be merged.

LitoMore:
Maybe we can have a better name for this option. For example: `description`.

spalladino:
I picked `long` as opposed to the existing `short` option, but I'm fine with changing it. Let me know if you'd prefer `description` or another alternative so I make the change in the code.

SBoudrias:
Hey @spalladino, let's go with `description`.

I feel long makes it sounds like a replacement option for the name; not extra information to display at the same time as the option name.

spalladino:
Done! Sorry for the delay!

LitoMore:
@spalladino Looks good! By the way, can you add test for this?

LitoMore:
ping @spalladino

spalladino:
Sorry for the delay @LitoMore. I looked into this, and found no tests at all for the `select` package, which made me dig a bit deeper and led to a question: what's the difference between the `select` package, and the regular `list` prompt in the `inquirer` package? Is `select` used, or should I have implemented this change in `list` instead?

If it's ok to keep it in `select`, then I should set up a new test suite from scratch. Is there any existing suite you suggest I try to copy for this?

SBoudrias:
Hey @spalladino, the select package is `@inquirer/select`. It is published right now as a beta package since it's part of a major refactor I hope to someday complete.

It's fine to skip the tests for this part - but I'll revert the documentation change since it won't be available on the currently official `inquirer` release.

PR #916, opened 2020-04-28
docs: ✏️ add inquirer-search-list to plugin list

robin-rpr:
added robin-rpr/inquirer-search-list plugin to README.md


SBoudrias:
Thanks!

PR #918, opened 2020-05-02
Fix @inquirer/confirm formatted output value

rbardini:
The `@inquirer/confirm` value is formatted twice, on keypress and once done for display:

https://github.com/SBoudrias/Inquirer.js/blob/32ee96f8ffa96198734c46faf86340bc36878914/packages/confirm/index.js#L13-L14

https://github.com/SBoudrias/Inquirer.js/blob/32ee96f8ffa96198734c46faf86340bc36878914/packages/confirm/index.js#L24-L25

This means the final formatted value is always `yes`, even if the entered value is `no`:

Issue demo

(callback value is correct, as it is not called with the formatted value.)

---

This PR removes the second formatting, thus fixing the issue:

Solution demo

### DIFF
--- a/packages/confirm/index.js
+++ b/packages/confirm/index.js
@@ -22,7 +22,7 @@ module.exports = createPrompt((config, done) => {
   let formattedValue = value;
   let defaultValue = '';
   if (status === 'done') {
-    formattedValue = chalk.cyan(value ? 'yes' : 'no');
+    formattedValue = chalk.cyan(value);
   } else {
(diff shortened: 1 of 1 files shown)


SBoudrias:
Thanks!

rbardini:
No problem, @SBoudrias!

Do you plan to release a new `@inquirer/confirm` version soon? I'd like to use it in a couple projects, but this issue is a blocker for me.

Please let me know if I can help in any way 😉

SBoudrias:
No worries! `7.2.0` is out

PR #923, opened 2020-05-27
Only mport used lodash methods

merceyz:
**What's the problem this PR addresses?**

- Inquirer is currently loading lodash through the main entrypoint which makes it impossible for webpack to treeshake the unused code.
- Yarn is using Inquirer in its cli and, because of this, is shipping a lot of unnecessary code that we don't need in our bundle.

Compared to the other PRs that attempted to tackle this in the past (https://github.com/SBoudrias/Inquirer.js/pulls?q=is%3Apr+is%3Aclosed+lodash+module) this PR doesn't use the per method packages (they're deprecated and didn't help; https://lodash.com/per-method-packages) but instead requires the used methods directly. This will get rid of unused code and potentially help startup times as there is less code to process.

I made the diff as small as possible which is why I construct a `_` object with the used methods. If the project switches to es2015 imports then this can be removed in favour of using https://github.com/lodash/babel-plugin-lodash

**How did you fix it?**

Imported only the used lodash methods

### DIFF
--- a/packages/core/index.js
+++ b/packages/core/index.js
@@ -1,4 +1,6 @@
-const _ = require('lodash');
+const _ = {
+  isFunction: require('lodash/isFunction')
+};
 const readline = require('readline');
 const chalk = require('chalk');
 const MuteStream = require('mute-stream');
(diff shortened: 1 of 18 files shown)


merceyz:
Hey @LitoMore :wave:, when do you think this will be released?

LitoMore:
@merceyz I don't have permission to publish on npm, let me contact @SBoudrias for help.

LitoMore:
@merceyz 7.2.0 released.

Issue #924, opened 2020-06-02
Allow loop list to be optional


adiktofsugar:
Based on comments here: https://github.com/SBoudrias/Inquirer.js/issues/26
And also personal experience.
Suggestion is question has `loop: false` as an option.

PR #925, opened 2020-06-02
feat: add loop: false for list type prompts

adiktofsugar:
Fixes: #924

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,7 @@ A question object is a `hash` containing question related values:
 - **prefix**: (String) Change the default _prefix_ message.
 - **suffix**: (String) Change the default _suffix_ message.
 - **askAnswered**: (Boolean) Force to prompt the question if the answer already exists.
+- **loop**: (Boolean) Enable list looping. Defaults: `true`
 
 `default`, `choices`(if defined as functions), `validate`, `filter` and `when` functions can be called asynchronously. Either return a promise or use `this.async()` to get a callback you'll call with the final value.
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -210,9 +210,11 @@ Separator instances have a property `type` equal to `separator`. This should all
 
 #### List - `{type: 'list'}`
 
-Take `type`, `name`, `message`, `choices`[, `default`, `filter`] properties. (Note that
+Take `type`, `name`, `message`, `choices`[, `default`, `filter`, `loop`] properties. (Note that
 default must be the choice `index` in the array or a choice `value`)
 
+`loop:false` will disable the default looping behavior.
```
LitoMore, line 216: You should add put this documentation to [Question](https://github.com/SBoudrias/Inquirer.js#question) section.

README.md
```diff
@@ -222,6 +224,8 @@ default must be the choice `index` in the array or a choice `value`)
 Take `type`, `name`, `message`, `choices`[, `default`, `filter`] properties. (Note that
 default must be the choice `index` in the array)
 
+`loop:false` will disable the default looping behavior.
```
LitoMore, line 227: Ditto.

README.md
adiktofsugar: moved

SBoudrias:
Thanks @adiktofsugar

sqren:
Hey @adiktofsugar 

Thanks for adding this. Just noticed one odd thing: for lists with more items than the `pageSize`, even with `loop:false` the list is still displayed as a loop (the last item "50" is above the first item "1"). Would be great if the loop effect could be disabled entirely:

backport bug2

PR #533, opened 2017-05-17
pass current answers hash to filter

imyelo:
resolve https://github.com/SBoudrias/Inquirer.js/issues/428 😉

PR #936, opened 2020-06-29
add documentation for changes made to add answers hash to filter in #533

FrederickEngelhardt:
Adds documentation for filter answers hash support which was added in #533


SBoudrias:
👌 great thanks!

PR #939, opened 2020-07-04
fix: paginator works with loop:false

adiktofsugar:
Follow up to https://github.com/SBoudrias/Inquirer.js/pull/925

I'm not sure if I should have changed the Paginator in packages/core. It seems like a copy, and it seems like this repo is moving toward a monorepo style, but I think the one in inquirer/utils that I've changed will fix the actual problem so I'll leave it to the maintainers to copy these changes in, unless you want me to do it.

### DIFF
--- a/packages/inquirer/lib/utils/paginator.js
+++ b/packages/inquirer/lib/utils/paginator.js
@@ -11,10 +11,11 @@ var chalk = require('chalk');
  */
 
 class Paginator {
-  constructor(screen, isInfinite) {
+  constructor(screen, options = {}) {
+    const { isInfinite = true } = options;
     this.lastIndex = 0;
     this.screen = screen;
-    this.isInfinite = isInfinite === undefined ? true : isInfinite;
+    this.isInfinite = isInfinite;
   }
 
   paginate(output, active, pageSize) {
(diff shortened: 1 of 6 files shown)


adiktofsugar:
This is what it looks like. The first prompt is with loop:false. The second is without.
inquirer-fixed

sqren:
Beautiful!

packages/inquirer/lib/utils/paginator.js
```diff
@@ -7,20 +7,18 @@ var _ = {
...
 /**
- * The paginator keeps track of a pointer index in a list and returns
- * a subset of the choices if the list is too long.
+ * The paginator returns a subset of the choices if the list is too long.
  */
 
 class Paginator {
-  constructor(screen) {
-    this.pointer = 0;
+  constructor(screen, isInfinite) {
```
SBoudrias, line 14: Let's pass this as option `{ isInfinite }`. Just a boolean argument won't convey the meaning.

packages/inquirer/lib/utils/paginator.js
```diff
@@ -33,7 +31,21 @@ class Paginator {
     if (lines.length <= pageSize) {
       return output;
     }
+    var getLinesFn = this.isInfinite ? this.getInfiniteLines : this.getFiniteLines;
+    var visibleLines = getLinesFn.call(this, lines, active, pageSize);
```
SBoudrias, line 35:
```js
var visibleLines = this.isInfinite
  ? this.getInfiniteLines(lines, active, pageSize)
  : this.getFiniteLines(lines, active, pageSize);
```

packages/inquirer/lib/utils/paginator.js
SBoudrias: Also, I believe all major versions on node supports defaults parameter `{ isInfinite = true }` and we can remove the type check default later in the code.

SBoudrias:
Thanks! 🙏

Issue #944, opened 2020-07-13
Allow loop in checkbox prompts to be optional


PabloPerezRodriguez:
Checkbox prompts don't allow you to disable looping when you reach the end of the list. But list and rawlist prompts allow you.

I'm asking for the same as #925 (allow loop list to be optional) but for checkbox prompts.

PR #945, opened 2020-07-13
feat: add loop: false for checkbox type prompts

PabloPerezRodriguez:
I spent quite a long time figuring out why `loop: false` didn't work in checkbox prompts.
Turns out it wasn't implemented.

Here's my go at it. Fixes: #944


SBoudrias:
Thanks!

PR #947, opened 2020-07-17
Fix: cleanup when the prompt fails.

mshima:


SBoudrias:
Ah great catch! Thanks for the fix :)

PR #948, opened 2020-07-21
Update index.js

7006:
rewrite using String.prototype.repeat available in NodeJs since v4.0.0

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat


SBoudrias:
Thanks!

PR #950, opened 2020-08-10
fix: readline.close causes freeze in Windows

chengzhuo5:
#767


chengzhuo5:
Node ver 14.6 has fixed this problem.

PR #954, opened 2020-08-30
feat: add file selection example

swseverance:
* Adds an example of how a user would be prompted to search through directories and select files

image

### DIFF
--- a/packages/inquirer/examples/file-selection.js
+++ b/packages/inquirer/examples/file-selection.js
@@ -1,0 +1,79 @@
+const fs = require("fs");
+const path = require("path");
+const inquirer = require("..");
+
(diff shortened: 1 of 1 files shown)


packages/inquirer/examples/file-selection.js
```diff
@@ -0,0 +1,79 @@
...
+    .prompt([
+      {
+        type: "list",
+        message: "Select files",
+        name: "selection",
+        choices,
+        default: () => lastFileSelected
+      },
+    ])
+    .then(({ selection }) => {
```
SBoudrias, line 57: Let's use await/async in the demos. Not doing it lead people to misunderstand that it is possible.

PR #975, opened 2020-10-22
Added an optional question value 'helper' and 'pageHelper'

JezerM:
I added an optional question value for custom help message on **list type**.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,10 +8,6 @@
 [![FOSSA Status](https://app.fossa.com/api/projects/git%2Bgithub.com%2FSBoudrias%2FInquirer.js.svg?type=shield)](https://app.fossa.com/projects/git%2Bgithub.com%2FSBoudrias%2FInquirer.js?ref=badge_shield)
 
 A collection of common interactive command line user interfaces.
-## Changes made by me: JezerM
-All these changes are made for a "better" user experience, specially to custom those default messages to other idioms.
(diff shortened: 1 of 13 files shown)


LitoMore:
Should document this chagne.

abhijit-hota:
Hello. A user here.

Can you update the PR to include another list-specific option to change the end text, too?
image

Maybe an option called `listOpts` can take an object having these two parameters because we wouldn't want tot crowd the main options too much.

JezerM:
> Hello. A user here.
> 
> Can you update the PR to include another list-specific option to change the end text, too?
> image
> 
> Maybe an option called `listOpts` can take an object having these two parameters because we wouldn't want to crowd the main options too much.

Actually, there is no specific List option to this end text, it refers to its Paginator util. Anyways, I made that change and now you could change it, check my latest commit :D

abhijit-hota:
@JezerM That's great! Waiting for this to get reviewed.

LitoMore:
You should think more about the documentation.

Please write the unit test for your code.

JezerM:
> You should think more about the documentation.
> 
> 
> 
> Please write the unit test for your code.

I've never worked with testing before, but I'll try it~

JezerM:
> You should think more about the documentation.
> 
> 
> 
> Please write the unit test for your code.

I tried to make the unit test, and went well (I guess).

PR #995, opened 2021-02-22
[docs] clarification about prompt method. 

knoid:
I had to look into the source code to find out that providing already answered questions was as easy as filling them in the 2nd argument. That's why I had to do this PR to help save time to future users. I saw it even has tests so it looks like a very thought-out feature.

Hope this helps.
Great library!


SBoudrias:
Thanks! That'll probably help other people indeed!

PR #1002, opened 2021-03-10
feat(prompt): progress indicator for filter and validate with optiona…

### LINKED ISSUE #631: FR: Show progress indicator on async validate

dhensby:
I've rebased the PR in #632 to try and help get this over the line.

Please let me know if there's anything else that needs to be done.

Closes #631


dhensby:
OK - I've refactored/rebased this since there have been quite a few layout changes in terms of file locations, etc.

I'm not sure how best to test these changes or increase the coverage - so any pointers on that would be appreciated. 🙏

SBoudrias (approved):

SBoudrias (approved):

SBoudrias:
Thanks! This is a great feature, I'm very happy to get this in :)

PR #1006, opened 2021-04-08
Expand: check for duplicates in a case-insensitive way

Lutetium-Vanadium:
When running, the expand prompt checks for keys in an case-insensitive manner. However, it did not do this in `validateChoices`.

Previously, if the following diff was applied to `packages/inquirer/examples/expand.js` it would not throw an error during validation.
```diff
           name: 'Abort',
           value: 'abort',
         },
+        {
+          key: 'X',
+          name: 'Abort2',
+          value: 'abort2',
+        },
       ],
     },
   ])
```

While running, if the duplicate key `x` is selected, the prompt would look like this:
prompt with duplicate

And it would return the first duplicate it encountered:

output after selecting the duplicate

This patch fixes it so that a duplicate key error is thrown during validation.


SBoudrias:
Thanks! Good fix :)

Issue #929, opened 2020-06-12
rawlist displays 'value' instead of 'short' after selection


jns:
Using Inquirer v7.1.0

Summary:
The rawlist type displays the 'value' instead of the 'short' after a user makes selection.

Steps to reproduce:
1) create a choices array of objects with properties `name`, `value`, and `short`
2) Add choices to a 'rawlist' type prompt
3) Run and make a selection.  The 'value' property is displayed after selection instead of 'short'

Note: 
This does not occur for a 'list' type.

idbartosz:
Bump, same with `7.3.3`.

CeroOscura:
Bump, the same issue I've been seeing in 8.0.0

Issue #889, opened 2020-02-12
Rawlist prints the value of the selected choice on answer instead of name or short


schl3ck:
Rawlist prints the value of the selected choice on answer instead of the properties `name` or `short`.

I digged into it, where this probably comes from and I think that it is here:
https://github.com/SBoudrias/Inquirer.js/blob/6a883e64dc238fc0dcc33ac77f6abcd4ca4dff5a/packages/inquirer/lib/prompts/rawlist.js#L61

The function that is bound here, returns what the name suggests, but as this also defines the value that is displayed to the user, it displayes `value` of the selected choice instead of its `name` or `short`.

ovidiu141:
This can be closed

PR #1013, opened 2021-05-06
fix: rawlist displays 'value' instead of 'short' after selection

Shuki-L:
Addresses 
https://github.com/SBoudrias/Inquirer.js/issues/929
https://github.com/SBoudrias/Inquirer.js/issues/889


SBoudrias:
Thanks!

PR #1020, opened 2021-06-15
fix: retain number prompt default

tjallingt:
The number prompt wants to use the default value if the input was invalid, however after a keypress (and if the default value is truthy) the default gets cleared. This is done so the default value is not rendered in the prompt anymore, see: https://github.com/SBoudrias/Inquirer.js/pull/719

This PR changes the logic so that instead of clearing the default value it sets a new state called "touched". We can then use this state to hide the default value from the rendered prompt.

Updated the relevant test so it would have failed before.


SBoudrias (approved):

SBoudrias:
Thanks!

PR #1026, opened 2021-07-03
fixes #1025 rawlist stops handling arrow keys after bad input

### LINKED ISSUE #1025: rawlist stops handling arrow keys after bad input
To reproduce:

1. Run the rawlist example.
2. Type `blah` and press `enter`.
3. Try moving with the arrow keys.

owings1:


SBoudrias (approved):

SBoudrias:
Thanks! Good catch

PR #1027, opened 2021-07-20
do not display password on async filter/validation, fixes #1022

### LINKED ISSUE #1022: masked/hidden password is shown when validate function performs long async operation
Steps to reproduce:

1. clone Inquirer.js repo
2. go to Inquirer.js\packages\inquirer\examples\password.js
3. change requireLetterAndNumber  method to look like:
```
const requireLetterAndNumber = async (value) => {
   async function timeout() {
      const promise = new Promise(resolve => {
          setTimeout(() => {
              resolve();
          }, 3000);
      });
      return promise;
  }
  await timeout();

  if (/\w/.test(value) && /\d/.test(value)) {
    return true;
  }

  return 'Password need to have at least a letter and a number';
}; 
```
4. run the example and get that masked /hidden password is visible while validation method is running

owings1:
The issue starts in [base.js line 100](https://github.com/SBoudrias/Inquirer.js/blob/9bee59b9/packages/inquirer/lib/prompts/base.js#L100):

```javascript
const validation = submit.pipe(
  flatMap((value) => {
    this.startSpinner(value, this.opt.filteringText);
    return asyncFilter(value, self.answers).then(
      (filteredValue) => {
        this.startSpinner(filteredValue, this.opt.validatingText);
```

My solution was to modify `startSpinner()` to call a new method `getSpinningValue()`, which defaults to a passthrough.

```javascript
startSpinner(value, bottomContent) {
  value = this.getSpinningValue(value);
  // ...
```
```javascript
getSpinningValue(value) {
  return value;
}
```

Then I modified `PasswordPrompt` to override that method to return the same as when it renders normally. I moved some code in `render()` to a new method `getMaskedValue()` to keep it dry.

The test case is committed before the fix, so you can checkout commit `6881259b` to see it fail. Or there is a line you can uncomment in the test case that should make it fail:

```javascript
/* This test should fail if you uncomment this line: */
// password.getSpinningValue = value => value;
```

I hoped to avoid adding a method to the base `Prompt` class, but couldn't think of a cleaner way. Maybe it is similar to the `transformer` feature on `InputPrompt`, so perhaps there is a better name than `getSpinningValue()`, or a way to combine masking, transforming, and spinning into something more general.

### DIFF
--- a/packages/inquirer/lib/prompts/base.js
+++ b/packages/inquirer/lib/prompts/base.js
@@ -130,7 +130,7 @@ class Prompt {
   }
 
   startSpinner(value, bottomContent) {
-    value = this.spinningValue(value);
+    value = this.getSpinningValue(value);
     // If the question will spin, cut off the prefix (for layout purposes)
     const content = bottomContent
       ? this.getQuestion() + value
@@ -143,9 +143,9 @@ class Prompt {
    * Allow override, e.g. for password prompts
    * See: https://github.com/SBoudrias/Inquirer.js/issues/1022
    *
-   * @return (String) value to display while spinning
+   * @return {String} value to display while spinning
    */
-  spinningValue(value) {
+  getSpinningValue(value) {
     return value;
   }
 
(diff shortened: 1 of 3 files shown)


JNK:
Would love to see this change merged... 💯

ianquigley-sap:
Any chance this PR can be merged?  Thanks :).

SBoudrias (approved):
Thanks for sending this PR :)

PR #1036, opened 2021-08-19
Notes on defaults improvements

iatenine:
Had a lot of trouble understanding the wording on notes regarding "default" around list, rawlist and expand. Tried my best to make them easy as possible to understand while retaining the original meaning. Also added small tweaks that will hopefully be viewed as formatting improvements


SBoudrias:
Thanks, I agree this is clearer 👏

Issue #1042, opened 2021-09-13
rawlist throws unhandable rejection


SanariSan:
Node: `14.17.4`
Language: `pure JS (no transpilation)` / `TS (tsc into ES2020)` (doesn't matter)

To be more accurate, rawlist throws multiple promise rejections if pressing "Enter" i.e. accepting after entering non-existing index. So, when validate error appears just press "Enter".

Moreover, you can't catch these rejections with .catch();
Also, custom validate function is not executing with type: "rawlist", can't evade it this way.

Sample code
```
await inquirer
	.prompt({
		type: "rawlist",
		name: "list",
		message: "sample list question",
		choices: ["first", "second"],
	})
	.catch((e) => {
		console.log("ERROR");
		console.log(e);
	});
```

Reproducing the error
[WEBM showing error](https://s1.webmshare.com/dPq4L.webm)

Full terminal log from sample

```
? sample list question 
  1) first
  2) second
  Answer: (node:109979) UnhandledPromiseRejectionWarning: AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:
>> Please enter a valid index
  assert(_.isNumber(selector))

    at Choices.getChoice (/home/me/Code/e/node_modules/inquirer/lib/objects/choices.js:63:5)
    at RawListPrompt.getCurrentValue (/home/me/Code/e/node_modules/inquirer/lib/prompts/rawlist.js:131:37)
    at /home/me/Code/e/node_modules/rxjs/dist/cjs/internal/operators/map.js:10:37
    at OperatorSubscriber._this._next (/home/me/Code/e/node_modules/rxjs/dist/cjs/internal/operators/OperatorSubscriber.js:28:21)
    at OperatorSubscriber.Subscriber.next (/home/me/Code/e/node_modules/rxjs/dist/cjs/internal/Subscriber.js:72:18)
    at Interface.handler (/home/me/Code/e/node_modules/rxjs/dist/cjs/internal/observable/fromEvent.js:59:31)
    at Interface.emit (events.js:400:28)
    at Interface._onLine (readline.js:364:10)
    at Interface._line (readline.js:700:8)
    at Interface._ttyWrite (readline.js:1045:14)
(Use `node --trace-warnings ...` to show where the warning was created)
[...]

PR #1045, opened 2021-09-14
fix of #1042 rawlist throw

SanariSan:
Fix of #1042

At the moment of pressing Enter key while Error being shown, `index` (1) [which](https://github.com/SBoudrias/Inquirer.js/blob/8e0b3d4c76bcf5c970936fec2561d3cc7964abb5/packages/inquirer/lib/prompts/rawlist.js#L122) comes as an argument equals to empty string.
At the same time `this.selected` equals to `undefined` (2) [here](https://github.com/SBoudrias/Inquirer.js/blob/8e0b3d4c76bcf5c970936fec2561d3cc7964abb5/packages/inquirer/lib/prompts/rawlist.js#L126).
```
  getCurrentValue(index) { // (1)
    if (index == null) {
      index = this.rawDefault;
    } else if (index === '') {
      index = this.selected; // (2)
    } else {
      index -= 1;
    }

    const choice = this.opt.choices.getChoice(index); // (4)
    return choice ? choice.value : null;
  }
```
This happens because (3) [here](https://github.com/SBoudrias/Inquirer.js/blob/8e0b3d4c76bcf5c970936fec2561d3cc7964abb5/packages/inquirer/lib/prompts/rawlist.js#L160) it gets `undefined` value every time non-existing index inserted, so at the time Error message shown, this value persists, which leads to assertion throw (4) [here](https://github.com/SBoudrias/Inquirer.js/blob/8e0b3d4c76bcf5c970936fec2561d3cc7964abb5/packages/inquirer/lib/prompts/rawlist.js#L131) -> (5) [here](https://github.com/SBoudrias/Inquirer.js/blob/8e0b3d4c76bcf5c970936fec2561d3cc7964abb5/packages/inquirer/lib/objects/choices.js#L63).
```
  onKeypress() {
    const index = this.rl.line.length ? Number(this.rl.line) - 1 : 0;


    if (this.opt.choices.getChoice(index)) {
      this.selected = index;
    } else {
      this.selected = undefined; // (3)
    }
    this.render();
  }
```
```
  getChoice(selector) {
    assert(_.isNumber(selector)); // (5)
    return this.realChoices[selector];
  }
```
To prevent this from happening I decided to re-assign (!) `this.selected` value if needed by checking with nullish coalescing operator.
This way, if user presses Enter without entering non-existing index beforehand (i.e. sending empty string), he gets the first (default) option selected, as it was before the change.
However, if Error been shown, `this.selected` will get `-1` value, which is number (so assert pass + no negative effect) and allows user to press Enter as much as he would like to (nothing happens, expected behavior; alternative to `-1` could be assigning `this.rawDefault`, then second Enter press just chooses first option available). 
At the same time this allows to get back to selection by just pressing arrows or entering number.
```
getCurrentValue(index) {
    if (index == null) {
      index = this.rawDefault;
    } else if (index === '') {
      this.selected = this.selected ?? -1; // (!)
      index = this.selected;
    } else {
      index -= 1;
    }

    const choice = this.opt.choices.getChoice(index);
    return choice ? choice.value : null;
  }
```

### DIFF
--- a/packages/inquirer/lib/prompts/rawlist.js
+++ b/packages/inquirer/lib/prompts/rawlist.js
@@ -123,6 +123,7 @@ class RawListPrompt extends Base {
     if (index == null) {
       index = this.rawDefault;
     } else if (index === '') {
+      this.selected = this.selected ?? -1;
       index = this.selected;
     } else {
       index -= 1;


SBoudrias (approved):

SBoudrias:
Thanks!

hingsir:
Node v10 don't support optional chaining `this.selected = this.selected ?? -1`

yiyu-liao:
same error , my Node Version  is v12.22.1

Issue #1046, opened 2021-09-15
Unexpected token '??'


stephenwoosley:
Looks like on line 126 of `packages/inquirer/lib/prompts/rawlist.js`, there is a ternary with an accidental added ?

zhouxiongking:
i am also getting this trouble。
![image](https://user-images.githubusercontent.com/15647000/133362863-bcb98e8d-342d-4954-9192-9fc7d9205bdc.png)

sxxhlx:
same problem ，this `??` token is not supported below node@14.0.0

Alexixyc:
HELP!!!!!!!!!!!!!!!  Please fix ！！！
![image](https://user-images.githubusercontent.com/17401583/133363773-1b575ec4-924a-4af8-a76e-32df61cb97d5.png)

stephenwoosley:
After looking into it more, looks like the nullish coalescing operator (??) was introduced in ES2020. I'm using node v12 so I'm sure that's the problem, but when I rolled back to inquirer@8.1.3 all was well.

stephenwoosley:
> same problem ，this `??` token is not supported below node@14.0.0

@sxxhlx probably want to roll back to 8.1.3?

yiyu-liao:
save error

hotoo:
+1

B2D1:
![image](https://user-images.githubusercontent.com/36991862/133369500-c7c2d429-a7ae-42e7-920e-9afea7374270.png)
 
my node version is below v14!

ViniSiva:
Is this issue resolved. we are still facing it.  Do we need to downgrade the version ?

enclairfarron:
breaking change

jlipps:
Yes, this was a breaking change in Node engine support on a patch release. recommended to pin to previous version until something like #1047 is merged and published.

splurgeop:
Same issue for us 
this is breaking our build.
When we upgraded to the latest version. This broke our pipeline

/build/node_modules/inquirer/lib/prompts/rawlist.js:126
this.selected = this.selected ?? -1;
^
SyntaxError: Unexpected token '?'
at wrapSafe (internal/modules/cjs/loader.js:915:16)
at Module._compile (internal/modules/cjs/loader.js:963:27)
at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
at Module.load (internal/modules/cjs/loader.js:863:32)
[...]

PR #1047, opened 2021-09-15
replace nullish coalescing operator with ternary

tys1128:
#1046

### DIFF
--- a/packages/inquirer/lib/prompts/rawlist.js
+++ b/packages/inquirer/lib/prompts/rawlist.js
@@ -123,7 +123,7 @@ class RawListPrompt extends Base {
     if (index == null) {
       index = this.rawDefault;
     } else if (index === '') {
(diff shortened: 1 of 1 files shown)


fengmk2 (approved):

fengmk2:
+1

SanariSan:
https://github.com/SBoudrias/Inquirer.js/issues/1046#issuecomment-919898281

alan-agius4 (approved):

SanariSan:
@alan-agius4 this pull request replacing nullish coalescing check with just `null` check, whilst initial point was to check for `undefined` value.

Also null value can't get here, because of if statement few line above 

#1048

alan-agius4:
`== null` also check for undefined values.

`undefined == null // true`

SBoudrias:
Other fix was merged. Thanks for the help!

PR #1055, opened 2021-10-05
feat: checkbox hints now stay on screen until enter is pressed

### LINKED ISSUE #1031: Instructions/hints aren't clear enough for checkbox prompts.
Hi, we've had a few users complain about the instructions on the checkbox prompt disappearing after they press space to make a selection, especially if they want to make more than one selection because they can't remember exactly what the hint said and what they're supposed to do next.

What do you think of keeping the instructions at https://github.com/SBoudrias/Inquirer.js/blob/master/packages/inquirer/lib/prompts/checkbox.js#L95-L104 on screen until the user moves past the checkbox part and adding a `, and <enter> to proceed` or something of the sort?

RodEsp:
This keeps the checkbox hint/instructions on screen and adds `, and <enter> to proceed` to the hint.

It also allows linting and testing to occur on Windows.

Example:
prompthints

Fixes #1031

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,4 +1,5 @@
 {
+  "endOfLine": "auto",
   "singleQuote": true,
   "printWidth": 90
 }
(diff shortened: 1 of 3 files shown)


RodEsp:
Woohoo! Thanks!

SBoudrias:
Released under 8.2.0

PR #1066, opened 2021-11-17
fix: Remove dependency of lodash isString

jimmywarting:
...My goal: totally remove bloated lodash dependency one tiny bit at the time


SBoudrias:
Make sense to do so, thanks for helping out clearing that out.

PR #1068, opened 2021-11-18
Remove more lodash

jimmywarting:


packages/core/lib/utils.js
```diff
@@ -1,22 +1,20 @@
...
  * @param {number} width
  * @return {string}
  */
 exports.breakLines = (content, width) => {
   const regex = new RegExp('(?:(?:\\033[[0-9;]*m)*.?){1,' + width + '}', 'g');
-  return _.flatten(
-    content.split('\n').map((line) => {
+  return content
+    .split('\n')
+    .map((line) => {
```
SBoudrias, line 12: This can be a [`flatMap`](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap)

packages/inquirer/lib/ui/baseUI.js
```diff
@@ -85,11 +81,13 @@ function setupReadlineOptions(opt) {
...
   const output = ms;
 
+  const { input: _, output: __, ...rest } = opt;
+
   return {
     terminal: true,
     input,
     output,
-    ..._.omit(opt, ['input', 'output']),
+    ...rest,
```
SBoudrias, line 90: Could also

```js
return {
    terminal: true,
    ...opt,
    input,
    output,
};
```

And skip any logic of `omit` completely.

packages/core/lib/utils.js
jimmywarting: i looked at it earlier and i though so too, didn't know why i didn't do it later

SBoudrias:
Thanks!

PR #1069, opened 2021-11-19
lodash almost gone

jimmywarting:
There are 3 remaining files after this gets merged.

They are slightly more annoying and harder to refactor out cuz of things like `Lodash 'where' clause` in `_.get` for example.
i would have use optional chaining...

Will leave the rest to you to decide what to do with it:

- [ ] packages/inquirer/lib/ui/prompt.js
- [ ] packages/inquirer/lib/prompts/base.js
- [ ] packages/inquirer/lib/objects/choices.js
- [ ] npm uninstall lodash (from all packages)

### DIFF
--- a/packages/core/lib/Paginator.js
+++ b/packages/core/lib/Paginator.js
@@ -1,8 +1,5 @@
 'use strict';
 
-const _ = {
-  flatten: require('lodash/flatten'),
-};
 const chalk = require('chalk');
 const cliWidth = require('cli-width');
 const { breakLines } = require('./utils');
@@ -45,7 +42,7 @@ class Paginator {
     this.lastIndex = active;
 
     // Duplicate the lines so it give an infinite list look
-    const infinite = _.flatten([lines, lines, lines]);
+    const infinite = [lines, lines, lines].flat();
     const topIndex = Math.max(0, active + lines.length - this.pointer);
 
     const section = infinite.splice(topIndex, pageSize).join('\n');
--- a/packages/inquirer/lib/ui/bottom-bar.js
+++ b/packages/inquirer/lib/ui/bottom-bar.js
@@ -6,9 +6,6 @@
 const through = require('through');
 const Base = require('./baseUI');
 const rlUtils = require('../utils/readline');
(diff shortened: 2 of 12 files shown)


SBoudrias (approved):

SBoudrias:
Ah yeah, that get/set in the prompt file will be hard to re-implement... I'd keep those there for simplicity; the new version (if it ever gets to stable 😅) would get rid of this since it's standalone prompt packages.

I think the other instances can be replaced with similar logic. Not sure the `_.isPlainObject` checks are really necessary; truthy/simple type check could be enough.

jimmywarting:
> Not sure the _.isPlainObject checks are really necessary; truthy/simple type check could be enough.

Yep, thought about it too. but wasn't 100% sure what gets passed in there. I'm new to this project...
back in my mind it goes: "hmm, 🤔 they must have some very specific reason to use a more adv isPlainObject and not just using a simple thrutly, `typeof x === 'object'` or `_.isObject` better not to touch it then". Null is an object after all...

https://github.com/SBoudrias/Inquirer.js/search?q=lodash

jimmywarting:
also here is a crazy idea while doing all this work, i found this:

https://github.com/SBoudrias/Inquirer.js/blob/3a796d8f54ab548f5c7c4cba217384e473d410b2/packages/inquirer/lib/objects/choices.js#L94-L122

Why not just go with: 
```js
class Choices extends Array {
  push(...args) {
    super.push(...args.map((val) => new Choice(val))
  }
}
```

Then you can have all array methods (iterators, .at(), .values(), .entries()) and so on :P 
totally okey now when we have class/extend

SBoudrias:
Ah yes, that was initially implemented before we were able to extend native classes (and even before `class` keyword 😅 ).

For the `isPlainObject` check, pretty sure it was used because of how unreliable `typeof foo === 'object'` is. Obv this code was written a long time ago, so I could've forgotten some specific cases calling for `isPlainObject` vs `isObject`... But I think the UTs would catch blatant error that could cause.

PR #1077, opened 2022-01-06
fix: bump engines field to Node.js>=12

erezrokah:
Hello 👋 

I noticed the engines field doesn't match the latest release https://github.com/SBoudrias/Inquirer.js/releases/tag/inquirer%408.0.0


SBoudrias:
Oh right, thanks!

PR #1084, opened 2022-02-11
Adding help information execution from "Nodemon"

umbrashia:
Many developers wants hot reload execution using "Nodemon"  
and  ``` "stdin": false ```` is very useful, without stdin property  "list" option not worked properly


umbrashia:
i updated all suggestion please review it and also share your valuable feedback :)

SBoudrias (approved):

PR #1087, opened 2022-02-21
chore: add example for a simple regex match on an input

lukeocodes:
I hope you don't mind. I thought examples/ could do with a few more basic prompts in there. This is the first of a few I was thinking of. I will PR the others when I get time.

### DIFF
--- a/packages/inquirer/examples/regex-validate-input.js
+++ b/packages/inquirer/examples/regex-validate-input.js
@@ -1,0 +1,26 @@
+/**
+ * Filter and validate progress example
+ */
+
+'use strict';
+const inquirer = require('..');
+
+/* eslint-disable no-promise-executor-return */
+const questions = [
+  {
+    type: 'input',
+    name: 'api_key',
+    message: 'Please enter a valid API key.',
+    validate(input) {
+      return Promise.resolve().then(() => {
+        if (!!input.match(/([a-f0-9]{40})/g)) {
+          return true;
+        } else throw "Please provide a valid API key.";
+      });
+    },
+  },
+];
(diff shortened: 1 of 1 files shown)


packages/inquirer/examples/regex-validate-input.js
```diff
@@ -0,0 +1,26 @@
...
+
+/* eslint-disable no-promise-executor-return */
+const questions = [
+  {
+    type: 'input',
+    name: 'api_key',
+    message: 'Please enter a valid API key.',
+    validate(input) {
+      return Promise.resolve().then(() => {
+        if (!!input.match(/([a-f0-9]{40})/g)) {
```
SBoudrias, line 16:
```suggestion
        if (/([a-f0-9]{40})/g.test(input)) {
```

Getting the type we want directly instead

SBoudrias:
Any reason to return a Promise in the example since the effective check is synchronous?

lukeocodes:
> Any reason to return a Promise in the example since the effective check is synchronous?

It seemed to suggest it preferred a promise as a response! I will change it

SBoudrias:
Yes and no, maybe examples are often async to be more easily extendable. Not against keeping the promise there, I was just wondering the rationale.

lukeocodes:
I think it is easier to understand without the promise so I've switched it.

lukeocodes:
Can you approve the workflow and i'll resolve the linting from the latest version. Thanks!

SBoudrias (approved):

SBoudrias:
🎉 thanks!

PR #1089, opened 2022-03-01
fix input of "rawlist"

RobInNYC:
This fixes a bug in the input processing of "rawlist". To reproduce the bug:

- Run a rawlist configuration
- Use the arrow keys to select an item
- Then use the number keys to enter a specific item, and press enter
- Get an "invalid option" error

Because the arrow keys record the selected index on the input line, when the user enters a specific index, an ultimately invalid index is processed by the code.

This works around that by not using the input line to record the selected index when using arrow keys. Instead it records the selected index in a private variable.


SBoudrias:
Thanks!

PR #1100, opened 2022-04-03
Add inquirer-interrupted-prompt plugin into readme

lnquy065:
Available at https://github.com/lnquy065/inquirer-interrupted-prompt
Turning any existing inquirer and its plugin prompts into prompts that can be interrupted with a custom key.


SBoudrias (approved):

SBoudrias:
Thanks!

PR #1103, opened 2022-04-18
Support negative float numbers

Felipe-BP:
I added a support for negative float numbers into number prompt by adjusting the regex. Also I added the unit test for this case.


SBoudrias:
Makes sense 😄 thanks!

PR #1104, opened 2022-04-22
Mention inquirer-tree-prompt in README

insightfuls:
This is pretty self-explanatory. I made a plugin which users might like to know about.


SBoudrias:
👏🏻 thanks!

PR #1106, opened 2022-04-24
fix: port breakLines to wrap-ansi

### LINKED ISSUE #304: Use wrap-ansi to force wrap lines
Should simplify the code by a great amount https://github.com/chalk/wrap-ansi

### LINKED ISSUE #1014: Using `terminal-link` causes question text to get broken into two lines
Hi everybody..

So I tried to use the [`terminal-link`](https://www.npmjs.com/package/terminal-link) package to create a link in one of the questions in my wizard, and what happens is the text gets broken into two lines (the red arrow):

image

If I do a regular `console.log`, everything seems to be printed correctly (green arrow).

Not sure if this is related to Inquirer.js or not.

I searched the issues and found there was an issue with a child dependency before, but it seems that was resolved and it didn't have much to do with this issue.

Anyone have any ideas on how to debug this maybe?

Thank you!

---
`terminal-link` version: `2.1.1`
` Inquirer.js` version: `8.0.0`

starpit:
This PR addresses incorrect line wrapping with `terminal-link`. It does so by employing the `wrap-ansi` npm.

Fixes #1014
Fixes #304

To the reviewers: I wasn't sure whether we should pass the [hard](https://github.com/chalk/wrap-ansi#hard) option to `wrap-ansi`:

> By default the wrap is soft, meaning long words may extend past the column width. Setting this to true will make it hard wrap at the column width.

If inquirer needs hard wrapping (in which case I assume wrap-ansi may split long words at arbitrary points in order to avoid overflow), we can easily add that to this PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,8 +44,5 @@
   "jest": {
     "coverageDirectory": "./coverage/",
     "collectCoverage": true
-  },
-  "dependencies": {
-    "wrap-ansi": "^7.0.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -45,5 +45,7 @@
     "coverageDirectory": "./coverage/",
     "collectCoverage": true
   },
-  "dependencies": {}
+  "dependencies": {
+    "wrap-ansi": "^7.0.0"
+  }
 }
--- a/packages/core/lib/utils.js
+++ b/packages/core/lib/utils.js
@@ -1,3 +1,5 @@
+const wrapAnsi = require('wrap-ansi');
+
 /**
  * Force line returns at specific width. This function is ANSI code friendly and it'll
  * ignore invisible codes during width calculation.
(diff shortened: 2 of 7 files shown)


SBoudrias:
Thanks for sending a PR; this looks very promising.

We'll need the `hard` option as terminal line returns are unfortunately messing up cursor position detection; which is why we implemented the `breakLines` feature in the first place.

package.json
```diff
@@ -45,5 +45,7 @@
     "coverageDirectory": "./coverage/",
     "collectCoverage": true
   },
-  "dependencies": {}
```
SBoudrias, line 48: This dependency must be added inside `packages/inquirer/package.json` (that's because this project is a monorepo)

SBoudrias:
Would you mind also addressing the new screen manager located at `/packages/core/lib/screen-manager.js`?

starpit:
Thanks for the great feedback. I have made the requested changes. Let me know if you need me to squash the commits. Not sure how of your preference, so i left them unsquashed for now.

package.json
starpit: thanks. i also added it to packages/core/package.json

SBoudrias (approved):
LGTM!

One extra thing if you have some time and don't mind, it'd be great to include a demo with terminal-link. I think it'll be hard to write an automated test; so at least a demo would help validate and caught regression when doing manual checks pre-releases.

starpit:
Sure, i'll add an example. I'll have to use terminal@2, since terminal@latest is ESM-only.

starpit:
## Expected line breaks (with this PR)
terminal-link correct line breaks

## Buggy line breaks (prior to this PR)
terminal-link bad line breaks

SBoudrias:
Awesome! Thanks a ton for following through with everything 🎉

PR #1108, opened 2022-04-27
fix: breakLines regression versus paginated content

starpit:
[This PR](https://github.com/SBoudrias/Inquirer.js/pull/1106) updated `breakLines` to use `wrap-ansi`. However, in doing so, we broke the paginator. It assumes that the return value will be of the same length as the original array, but that broken lines will be subarrays.


SBoudrias (approved):
Oh interesting... 🤦🏻 not exactly the most straightforward interface

PR #1109, opened 2022-05-01
Minimize lodash load

starpit:
This PR replaces the use of the monolithic `lodash` dependence with the much lighter node modules that offer selective functionality.

It also removes lodash from packages/core/package.json. I don't see any use of it there, so it seems that we can elmininate it entirely from that package.json?

I realize that the goal is to switch over to packages/core and away from packages/inquirer, but this seems like a simple enough way to help tide us over a bit longer?

A full lodash install is 4.9M:
```
1.6M	node_modules/lodash/fp
4.9M	node_modules/lodash
```

Whereas the selective lodash modules add up to 340k:
```
 60K	node_modules/lodash.clone
 40K	node_modules/lodash.set
 76K	node_modules/lodash.map
 76K	node_modules/lodash.filter
 36K	node_modules/lodash.get
 16K	node_modules/lodash.isplainobject
 32K	node_modules/lodash.defaults
```


SBoudrias:
Hi, thanks for sending a PR. But that's being discussed many times, and the chances of lodash being installed/cached on a user computer are higher than each of those sub-modules. And so we think for most users, the current version is preferable. We've already updated to only load the JS required from lodash to reduce import time cost; I think this optimization is enough - and it's also easier to maintain the dependencies.

starpit:
Respectfully disagree, but thanks for the consideration.

PR #1112, opened 2022-05-13
WIP Setup typescript

SBoudrias:
Starting a migration to TS file by file.

Happy to receive some help wrapping up some errors or migrating more files. Just send PR targeting this branch for now.


SBoudrias:
I fixed the CI & demos properly - the esm migration probably wasn't ran and it wasn't working for standard Node.js - just compiled "fake" esm on the UTs. Anyhow, it should be easier to test PR against this branch now.

I had rebase/squashed to remove the noise from the heavy divergence on master (too many commits to resolves conflict from). So if you cloned this one locally, make sure you reset to the new HEAD.

SBoudrias:
Went through and finished typing `@inquirer/core` since there was too much interdependencies between all files.

I'd like to migrate a few prompts to validate the core types are working as expected.

PR #1113, opened 2022-05-14
Paginator.js migrated to Paginator.ts

TDP17:
index.d.ts was added for the library **cli-width** which does not provide types (no @types/cli-width package present either). 
Not sure if this is the way to go.

Looking for feedback on this, so I can start migrating other files too. Thanks.

PS - While tinkering with files with export, building with tsc packages/core/src/lib/key.ts and running node packages/core/src/lib/key.js throws 
`exports.__esModule = true;`
`ReferenceError: exports is not defined`
Any idea on how I should go about tackling this (or ignoring it)?

### DIFF
--- a/packages/core/src/lib/Paginator.ts
+++ b/packages/core/src/lib/Paginator.ts
@@ -1,3 +1,4 @@
+import * as readline from 'readline';
 import chalk from 'chalk';
 import cliWidth from 'cli-width';
 import { breakLines } from './utils';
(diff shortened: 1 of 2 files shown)


SBoudrias:
> `ReferenceError: exports is not defined`

Looks like it tries to compile to commonjs instead of esm modules. Might be a bundle misconfiguration.

The branch looks good, thanks! For the missing module types, I think that's correct. Maybe one day we send PRs to add type defs to DefitinivelyTyped - but those aren't used much. I'm fine them being `any` while we migrate.

Issue #1101, opened 2022-04-13
Refactor to TypeScript


Exortions:
Refactoring the codebase to TypeScript in my opinion would have a few advantages:

  1. Much easier to understand how things work for new/aspiring contributors
  2. Because TypeScript compiles to minified JavaScript, the bundle size for users would be smaller, and run times would be faster.
  3. Overall, it's just a good idea as Types add a lot of functionality to JavaScript.

Of course, I understand if TypeScript isn't your thing, but if it is, I'd be happy to make a PR for at least some (if not most) of the packages.

SBoudrias:
I'm happy to review PRs starting on the new packages. I wanted to add TS, just doesn't have the time.

denik1981:
With a progressive strategy and a nice roadmap, this could be done.

SBoudrias:
The first stab is launched over https://github.com/SBoudrias/Inquirer.js/pull/1112 - there's a fair amount of work, but I explained which files to focus on there. We can do them one by one!

Exortions:
Yep, you should create a new branch for TypeScript.

Exortions:
Nevermind, looks like you already did 😅

Exortions:
@SBoudrias I'll be working on packages/core/src/lib/screen-manager.js.

TDP17:
@SBoudrias I'd like to help migrate the other packages. Can I work on @inquirer/password and @inquirer/select and others if they are free as well?

I'll take @inquirer/checkbox as a reference

SBoudrias:
@TDP17 that'd be great! I'll leave you those for a week, feel free to send an incomplete PR if you get stuck somewhere; that way I'll know the progress and can help take the branch over if I want to wrap this up sooner 😄

mokkabonna:
If I want to convert my plugin to typescript https://github.com/mokkabonna/inquirer-autocomplete-prompt is the approach here the way to go? Import from @inquirer/core and use the createprompt?

```ts
type SelectConfig = AsyncPromptConfig & {
  choices: { value: string; name?: string; description?: string; disabled?: boolean }[];
  pageSize?: number;
};

export default createPrompt<string, SelectConfig>((config, done) => {
```

SBoudrias:
Hey @mokkabonna, yes that'd be the ideal. One thing to note is that afterward your custom prompt will be usable standalone (which I think overall is better.)

[...]

PR #1130, opened 2022-06-20
Migrate @inquirer/input & @inquirer/password to TS

TDP17:
Broken PR with reference to https://github.com/SBoudrias/Inquirer.js/issues/1101#issuecomment-1160453317 :p

A few questions/issues I had -
- In index.ts, I've created a PasswordConfig, should this be a base type for all configs and used via Partial<Config> for all packages configs?
- On https://www.npmjs.com/package/@inquirer/password Mask is a boolean but the correct type seems to be string 
- Could not test since `Could not find a declaration file for module '@inquirer/input'`. Any workarounds for this? From what I understand I might need to fix ts errors in all packages before I can run tests


SBoudrias:
> Mask is a boolean but the correct type seems to be string

That's correct, it's the string of the character used to mask the characters. I forgot the exact details, but maybe `false` is also a valid value that'll just not mask the character and let the user write a blank password.

> Could not test since Could not find a declaration file for module '@inquirer/input'. Any workarounds for this? From what I understand I might need to fix ts errors in all packages before I can run tests

Ahhh! I didn't think about this, but the password prompt is a special case since it wraps the input prompt 🤦🏻 You'll need to migrate `input` first for this to work.

packages/password/src/index.ts
```diff
@@ -1,7 +1,18 @@
 import input from '@inquirer/input';
 import chalk from 'chalk';
 
-export default (config, ...args) => {
+type PasswordConfig = {
```
SBoudrias, line 4: To answer the other question from your post, I think this should become something like

```ts
type PasswordConfig = InputConfig & { ... };
```

Once the input prompt is migrated, we should export (or extract if possible) its config type and reuse it here since the password prompt is only a superset.

TDP17: Alright will do, would you like me to close this request and make a new one in a day or two?

TDP17:
Before I close this I had one issue which might concern the core packages

This is my type for inputconfig
```
type InputConfig = {
  message: string | Promise<string> | (() => Promise<string>);
  default?: string;
  transformer?: (value: string, { isFinal }: { isFinal: boolean }) => string;
  validate?: (value: string) => (boolean | string | Promise<string | boolean>);
};
```

Which leads to message types being incompatible
```
Type 'InputConfig' does not satisfy the constraint 'AsyncPromptConfig'.
  Types of property 'validate' are incompatible.
    Type '((value: string) => string | boolean | Promise<string | boolean>) | undefined' is not assignable to type '(() => boolean) | undefined'.
      Type '(value: string) => string | boolean | Promise<string | boolean>' is not assignable to type '() => boolean'.
```

**I have pushed this change on this PR itself for reference**, unsure what to do about this. Am I doing something wrong? From what I understand validate should be ` (string) => (boolean | string | Promise<string | boolean>)`

PS - 
I even tried setting config type directly with 
```
type InputConfig = AsyncPromptConfig & {
  message: string | (() => string | Promise<string>);
  default?: string;
  transformer?: (value: string, { isFinal }: { isFinal: boolean }) => string;
  validate?: (value: string) => (boolean | string | Promise<string | boolean>);
}
``` 
even with `Omit<AsyncPromptConfig>`, either the demo fails or ts errors

packages/password/src/index.ts
SBoudrias: Oh you can just update this branch I think. But up to you

SBoudrias:
Oh, seems like there might be a bug in the core type? I think you can modify `() => boolean` in the core.

TDP17:
@SBoudrias pinging for review, I think I've wrapped up most of core and password. 

While running yarn test I ran into, maybe this is the reason checks were running for 4 hours without resolution? (https://github.com/SBoudrias/Inquirer.js/pull/1130/checks)
image
I don't think this has anything to do with my changes. Might be something wrong with my local machine? lmk pl

Also I'm unsure on this ` const [errorMsg, setError] = useState<any>(null);`

Thanks :)

TDP17:
> Hum, this looks like eslint not realizing it's running for Node type environment and ESM... What command did you run exactly? Did you make sure to run `yarn install` at the root first?

I did, I'm running `yarn test` on root. I'll start afresh locally and check

TDP17:
Any updates on this?

Once you approve this I can move on to the other packages and hopefully port them all over the weekend on this PR itself

SBoudrias:
Hey @TDP17, sorry I assume it wasn't ready since CI failed. Looks like the UTs aren't running 🤔 There might be some syntax error throwing off jest

TDP17:
Hey thanks for the quick reply, it seems CI has been broken from commit b7a8faf1ef1c5085f1f85f4a119ea8cd26e236e4 in the same way. I originally thought it was due to `// @ts-expect-error broken type check from library` but not having this just throws an error. Let me know how to proceed and/or wait for the resolution of the issue

SBoudrias:
@TDP17 I'll try to take a look next weekend. I won't have time to dig before then :(

SBoudrias:
The branch was running fine locally, I'll merge and we can figure out Github action later 🤔

PR #1133, opened 2022-07-02
@inquirer/rawlist TS migration

TDP17:
#1101 continuation

Some other things to be noted

- stdio in password is now optional

Correct me if I'm wrong but I don't fully understand the difference in the old inquirer and the new separate packages and why do they differ in options available - for example filter doesn't seem to be supported on any prompt types. Is the new core a WIP?

### DIFF
--- a/packages/rawlist/package.json
+++ b/packages/rawlist/package.json
@@ -3,7 +3,11 @@
   "type": "module",
   "version": "0.0.4-alpha.0",
   "description": "Inquirer rawlist prompt",
-  "main": "index.js",
+  "main": "dist/index.js",
+  "typings": "dist/index.d.ts",
+  "files": [
+    "dist/"
+  ],
   "repository": "SBoudrias/Inquirer.js",
   "keywords": [
     "cli",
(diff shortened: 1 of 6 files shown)


TDP17:
I was thinking of adding the other packages in this PR itself, or would you like different ones?

packages/rawlist/package.json
```diff
@@ -22,6 +26,9 @@
     "@inquirer/core": "^0.0.21-alpha.0",
     "chalk": "^5.0.1"
   },
+  "scripts": {
+    "tsc": "tsc"
+  },
```
LitoMore, line 31: This is unnecessary. We can use `npx tsc` intead.

SBoudrias: This is for lerna, it allows to run it easily on all packages

SBoudrias (approved):

SBoudrias:
Thanks!!

I prefer one PR per prompt; it'll be easier to review/merge as we go too. All in one PR will be harder, especially if there's some back and forth on some trickier implementation 😄

PR #1141, opened 2022-07-19
fix: help tip never render in checkbox type

zonemeen:
If the initial value of showHelpTip is set to false, help tip never render in checkbox type.


SBoudrias:
Thanks!

PR #1143, opened 2022-07-20
fix(checkbox): the help tip should disappear when the enter key is triggered

zonemeen:
Hi, according to the previous version of `Inquirer.js`, the help tip should disappear when the enter key is triggered, not when the space key is triggered.


zonemeen:
I understand, anyway also thank you for your reply, then I'll close the PR.

PR #1146, opened 2022-07-22
fix: message renders repeatedly when text spans lines

zonemeen:
When the text is too long and spans lines, all keys that can cause interaction will cause the message to be displayed repeatedly.
The bottom is a reproduction of the bug:

1.checkbox(version: `@inquirer/checkbox@0.0.23-alpha.0`):
checkbox-repeat-render

2.select(version: `@inquirer/select@0.0.22-alpha.0`):
select-repeat-render

The rest also have this bug, I didn't record it as a gif.

System Info:
NodeJS 16.10.0
windows 10


SBoudrias (approved):
Thanks!

PR #1148, opened 2022-07-26
feat(select): the help tip at first render

zonemeen:
The list type in `inquirer@9.0.2` has a help-tip message at first render, while in `@inquirer/select@0.0.22-alpha.0` there is no.

1.The list type in `inquirer@9.0.2`:
list

2.The select type in `@inquirer/select@0.0.22-alpha.0`:
select

### DIFF
--- a/packages/select/src/index.ts
+++ b/packages/select/src/index.ts
@@ -23,6 +23,7 @@ type SelectConfig = AsyncPromptConfig & {
 export default createPrompt<string, SelectConfig>((config, done) => {
   const [status, setStatus] = useState('pending');
   const [cursorPosition, setCursorPos] = useState(0);
+  const [firstRender, setFirstRender] = useState(true);
   const { choices, pageSize = 7 } = config;
   const paginator = useRef(new Paginator()).current;
   const prefix = usePrefix();
@@ -56,7 +57,12 @@ export default createPrompt<string, SelectConfig>((config, done) => {
     }
   });
 
-  const message = chalk.bold(config.message);
+  let message = chalk.bold(config.message);
+
+  if (firstRender) {
+    message += chalk.dim('(Use arrow keys)');
+    setFirstRender(false);
+  }
 
   if (status === 'done') {
     const choice = choices[cursorPosition]!;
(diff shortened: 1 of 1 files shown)


packages/select/src/index.ts
```diff
@@ -23,6 +23,7 @@ type SelectConfig = AsyncPromptConfig & {
 export default createPrompt<string, SelectConfig>((config, done) => {
   const [status, setStatus] = useState('pending');
   const [cursorPosition, setCursorPos] = useState(0);
+  const [firstRender, setFirstRender] = useState(true);
```
SBoudrias, line 26: We should `useRef` for this one so updating it to false doesn't trigger a re-renders.

zonemeen: Thanks, I'll resubmit a PR.

zonemeen: Would it be better if I used `useEffect` to execute only on the first render?

SBoudrias: Oh, I wouldn't since that wouldn't work in React; I'm trying to keep the API consistent, but I see hooks like `useEffect` aren't behaving exactly in the same way now 😅

zonemeen: Since I'm not very familiar with hooks, I can't realize this feature by simply using `useRef` here, do you have any other good solution? Thanks 😄

SBoudrias: No worries, ref are the right solution... But there's a bug in them and they're not referentially equal as I was hoping for... 🤦🏻 I debugged it earlier this week, but didn't have enough time to find the root cause yet.

zonemeen: Thanks for your work.😄

SBoudrias:
Hey! Finally got some time to track down the bug! So it was a bit deep, but quite important, so really nice thing that is fixed now. `useRef` now works as expected, and I've merged this PR and my adjustment manually on the `master` branch.

Thanks for kicking this off - and surfacing a major bug!

zonemeen:
Thanks also for the work you do.

Issue #1149, opened 2022-08-03
[Feature request] launch editor prompt without waiting for <enter>


dellamina:
I'm currently developing an application where I ask the user for a filename (text) and then for the content (editor) but as far as I know there is no way to remove the prompt for the user to press <enter> before launching the editor.
I understand why it works that way but there are cases like in git commit where it makes more sense to just launch the editor without waiting for user input.

SBoudrias:
I'm happy to review a PR adding an option to skip waiting for the enter key.

PR #1150, opened 2022-08-04
feat(editor): add flag waitUserInput

dellamina:
As requested in #1149, editor prompt always requires the user to press enter before launching the system editor, this pr adds a flag `waitUserInput` (default: `true`) that if set to `false` skips the user confirmation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,7 @@ A question object is a `hash` containing question related values:
 - **suffix**: (String) Change the default _suffix_ message.
 - **askAnswered**: (Boolean) Force to prompt the question if the answer already exists.
 - **loop**: (Boolean) Enable list looping. Defaults: `true`
+- **waitUserInput**: (Boolean) Flag to enable/disable wait for user input before opening system editor - Defaults: `true`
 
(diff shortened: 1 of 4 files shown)


SBoudrias (approved):

SBoudrias:
Thanks!!

If you have some extra time, would you mind adding the flag to the new version of the editor prompt too? This would help me make sure we don't loose the feature once we ship the rewrite officially! The new code is over here: https://github.com/SBoudrias/Inquirer.js/blob/master/packages/editor/src/index.ts

dellamina:
Sure, I actually have already done that but I wasnt able to find the test suite for the new version so I pushed the changes only for the js version. Can you give me a hint on how to test the new version?

SBoudrias:
Ah yes, I didn't setup great testing infra for the new prompts; I only tested the core library so far. You can skip UTs for now; but if you're interested, you could take some time looking into how we could set that up... There might be some inspiration in this test here https://github.com/SBoudrias/Inquirer.js/blob/master/packages/core/test.ts - Passing a test stream and retrieving rendered info from there. I was planning to extract a simple reusable test interface in a package too to make it easier.

PR #1154, opened 2022-08-11
feat(checkbox): complement the help tip message

zonemeen:
Complement the default help tip message like the old version, as follows:

old-version


SBoudrias (approved):

SBoudrias:
Thanks!!

Issue #62, opened 2013-09-02
Event listeners warning on multiple prompts


mpal9000:
Hello, thanks for this nice module!

When running multiple prompts, I get this warning:
"(node) warning: possible EventEmitter memory leak detected. 11 listeners added. Use emitter.setMaxListeners() to increase limit."

Tests:

``` javascript
  var inquirer = require('inquirer');
  var prompt1 = function (callback) {
    inquirer.prompt([{
      type: 'list',
      name: 'choice',
      message: 'Select:',
      choices: ['one', 'two']
    }], function () {
      callback(callback);
    });
  };
  var prompt2 = function () {
    inquirer.prompt([{
      type: 'list',
      name: 'choice',
      message: 'Select:',
      choices: ['three', 'four']
    }], function () {});
  };
  var prompt3 = function () {
    inquirer.prompt([{
      type: 'list',
      name: 'choice',
      message: 'Select:',
      choices: ['five', 'six']
    }], function () {});
  };


  // Test 1
  prompt1(prompt1);

  // Test 2
  for (var i = 1; i < 11; i += 1) { prompt2(); }

  // Test 3
  for (var i = 1; i < 11; i += 1) { if (i % 2 === 0) { prompt2(); } else { prompt3(); } }

```

Luke-Robinett:
I've got a mainMenu function that presents the first level of options. Based on the user's choice, they are presented with the relevant submenu (stored in a different function). Some of those submenu options again have a set of related options or input prompts, and so on. I want mainMenu to repeat until the user chooses the "Quit" rawlist option. To achieve this, I recursively call mainMenu() after the switch statement in mainMenu. Unfortunately I know this is killing the stack and causing the error being discussed here. FYI all my functions are async so not using the .then notation. What is the correct way to implement a repeating, multinested menu system in Inquirer? Thanks. Attached my server JS for reference.
[server.txt](https://github.com/SBoudrias/Inquirer.js/files/4131263/server.txt)

vltansky:
[...]

PR #1160, opened 2022-08-29
Call stream end to trigger cleanup

intrnl:
Getting a similar issue to #62 for the new Inquirer.
Calling `end()` like before seems to fix it.


SBoudrias:
Thanks for sending the PR! 🙏🏻

PR #1163, opened 2022-09-28
fix: unable to install dependencies

zonemeen:
If don‘t add "private": true, won't be able to install the dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,4 +1,5 @@
 {
+  "private": true,
   "type": "module",
   "devDependencies": {
     "@jest/globals": "^29.0.3",


SBoudrias (approved):

SBoudrias:
Thanks!

LitoMore:
@SBoudrias What is this commit https://github.com/SBoudrias/Inquirer.js/commit/7f27ea6b8c7e0bc28dcb538d62352b7515a511f3 to fix for?

I noticed this but not sure if you are trying to fix another issue.

SBoudrias:
Oh I wanted to make `@inquirer/test` private initially; but saw it didn't work with Lerna, and it broke the repo state being able to release. This seemed to block lerna initially; but then it didn't make sense... I thought it was due to an update.

PR #1166, opened 2022-09-28
fix(checkbox): press up or down key will cause an infinite loop while

zonemeen:
### Reproduction

Press `up` or `down` key will cause an infinite loop while when all selections are disabled and also cause the command line to get stuck.

GIF 2022-9-28 19-15-51

Reproduction address: https://stackblitz.com/edit/node-8cbusm?file=index.js


SBoudrias:
Interesting... What should be the behaviour when everything is disabled? Doing nothing doesn't feel like the best behaviour

PR #1171, opened 2022-10-02
feat(packages/editor): add flag waitUserInput

dellamina:
Same change as in #1150 but for rewritten `packages/editor`.

Not sure how you feel about it but I needed to expose the InquirerReadLine object to the prompt to be able to launch the external editor without requiring user input.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -91,6 +91,18 @@ export function useRef<Value>(val: Value): { current: Value } {
   return useState({ current: val })[0];
 }
 
+export function useReadline(
+  userHandler: (rl: InquirerReadline) => void
+) {
+  const rl = sessionRl;
+
+  if (!rl) {
+    throw new Error('useReadline must be used within a prompt');
+  }
+
+  userHandler(rl);
+}
+
 export type AsyncPromptConfig = {
   message: string | Promise<string> | (() => Promise<string>);
   validate?: (value: string) => boolean | string | Promise<string | boolean>;
--- a/packages/editor/src/index.ts
+++ b/packages/editor/src/index.ts
@@ -7,17 +7,50 @@ import {
   usePrefix,
   isEnterKey,
   AsyncPromptConfig,
+  InquirerReadline,
+  useEffect,
+  useReadline,
 } from '@inquirer/core';
 
 type EditorConfig = AsyncPromptConfig & {
   default?: string;
   postfix?: string;
+  waitUserInput?: boolean;
 };
 
 export default createPrompt<string, EditorConfig>((config, done) => {
   const [status, setStatus] = useState<string>('pending');
   const [value, setValue] = useState<string>(config.default || '');
   const [errorMsg, setError] = useState<string | undefined>(undefined);
+  const [firstRender, setFirstRender] = useState<boolean>(true);
+
+  const startExternalEditor = (rl: InquirerReadline) => {
+    rl.pause();
+    editAsync(
+      value,
+      async (error, answer) => {
+        rl.resume();
+        if (error) {
+          setError(error.toString());
+        } else {
+          setStatus('loading');
+          const isValid = await config.validate(answer);
+          if (isValid === true) {
+            setError(undefined);
+            setStatus('done');
+            done(answer);
+          } else {
+            setValue(answer);
+            setError(isValid || 'You must provide a valid value');
+            setStatus('pending');
+          }
+        }
+      },
+      {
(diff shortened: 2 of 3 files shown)


packages/editor/src/index.ts
```diff
@@ -61,7 +61,7 @@ export default createPrompt<string, EditorConfig>((config, done, rl) => {
     }
   });
 
-  if (!(config.waitUserInput ?? true) && firstRender && rl != undefined) {
+  if (!(config.waitUserInput ?? true) && firstRender && rl !== undefined) {
```
SBoudrias, line 64:
```suggestion
  if (config.waitUserInput != null && firstRender && rl !== undefined) {
```

packages/core/src/index.ts
```diff
@@ -155,7 +156,7 @@ export function createPrompt<Value, Config extends AsyncPromptConfig>(
         index = 0;
         handleChange = () => workLoop();
 
-        const nextView = view(resolvedConfig, done);
+        const nextView = view(resolvedConfig, done, sessionRl);
```
SBoudrias, line 159: I think I'd add a `useReadline()` hook instead 🤔

packages/editor/src/index.ts
```diff
@@ -7,53 +7,66 @@ import {
...
-      );
+      startExternalEditor(rl);
     }
   });
 
+  if (!(config.waitUserInput ?? true) && firstRender && rl !== undefined) {
+    startExternalEditor(rl);
+    firstRender = false;
+    return '';
+  }
```
SBoudrias, line 68: Let's put this into a `useEffect()` handler instead. It doesn't change anything now, but I think that'll be more in line with potential future optimizations.

packages/editor/src/index.ts
```diff
@@ -7,53 +7,66 @@ import {
...
 } from '@inquirer/core';
 
 type EditorConfig = AsyncPromptConfig & {
   default?: string;
   postfix?: string;
+  waitUserInput?: boolean;
 };
 
-export default createPrompt<string, EditorConfig>((config, done) => {
+let firstRender = true;
```
SBoudrias, line 19: We need to use a ref for this; see `useRef()` usage

packages/editor/src/index.ts
dellamina: What about the code below so the user can also pass `true` (default) and the user is still required to press `<enter>`.

```ts
if (config.waitUserInput != null && config.waitUserInput === false && firstRender && rl !== undefined) {
```

Thanks also for the other suggestions/changes, I will look into those and try to fix them.

SBoudrias: Also works, but since it's the default, maybe we could just type as `waitUserInput?: false`. No big difference either way; just don't like the double conditions `!(a ?? B)` (or other similar forms)

packages/core/src/index.ts
```diff
@@ -91,6 +91,18 @@ export function useRef<Value>(val: Value): { current: Value } {
...
+export function useReadline(
+  userHandler: (rl: InquirerReadline) => void
+) {
+  const rl = sessionRl;
+
+  if (!rl) {
+    throw new Error('useReadline must be used within a prompt');
+  }
+
+  userHandler(rl);
```
SBoudrias, line 103: Let's just have this hook return the readline; then in the prompt it'll just be:

```js
const rl = useReadline();

useEffect(() => {
  if (!waitForUseInput) {
    startExternalEditor();
  }
}, []);
```

packages/editor/src/index.ts
```diff
@@ -26,34 +59,21 @@ export default createPrompt<string, EditorConfig>((config, done) => {
...
-        {
-          postfix: config.postfix || '.txt',
-        }
-      );
+      startExternalEditor(rl);
     }
   });
 
+  if (config.waitUserInput === false && firstRender) {
+    useEffect(() => {
```
SBoudrias, line 67: By React hooks rules, they must be invoked outside conditional structure. So the `if` should be moved inside the `useEffect` callback.

GFargo:
This PR has been hanging out for awhile, and being able to jump directly into the editor would be an amazingly helpful piece of functionality to have for the `editor` component.  

Any opposition to either updating this PR with the required changes to merge, or me creating a fresh PR with the outlined updates from @SBoudrias ?

SBoudrias:
Thanks for the help @dellamina. I made the modifications and finally decided to go with a slightly different solution. It's now in the codebase.

PR #1172, opened 2022-10-05
feat: customize `select` and `checkbox` disabled hints

zonemeen:
The feature is add the customized disabled hint about select and checkbox.

### DIFF
--- a/packages/checkbox/src/index.ts
+++ b/packages/checkbox/src/index.ts
@@ -18,7 +18,7 @@ import ansiEscapes from 'ansi-escapes';
 export type Choice<Value> = {
   name?: string;
   value: Value;
-  disabled?: boolean;
+  disabled?: boolean | string;
 };
 
 type Config<Value> = {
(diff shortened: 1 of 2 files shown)


SBoudrias:
Hey @zonemeen thanks for sending the PR.

I'd like to have a solution to allow overwriting any text inside inquirer; the problem is this one handles a single/one-off case, and wouldn't scale to where I want to take the system.

Ideally I'd like to add a locale/translation option to the context argument (the second one.) But there's a few complications which I'm not sure how to handle. For example, there's a lot of colorized, or split strings in the rendering functions; and I'm not sure what a good interface for those would be 🤔 

Naively, I'm currently thinking of creating a `Locale` class inside `@inquirer/core` that'd `stripAnsi` from the string we want to retrieve; that'd handle the color problem. For the rest, I think we might be able to refactor some code to use full strings and removing string pieces concatenation; or maybe we should use string template (but I'm less certain that's desirable as I want to keep the new packages as small as possible).

zonemeen:
Hi @SBoudrias , I'm really sorry, I don't have any good ideas for now, when I have time to look at the whole project structure and code, and then give my suggestions or solutions 😩

Issue #1159, opened 2022-08-19
Provide multi-module builds


SBoudrias:
If someone want to help, I'd like to configure TS to produce ESM & common.js modules (if possible 🤔)

The pure ESM output has been a pain for many users, so I'd like to reduce the amount of support we need to do, and allow people to stay on latest.

zonemeen:
I recommend trying to use [tsup](https://github.com/egoist/tsup) to package both ESM and common.js modules.

SBoudrias:
I tried this out yesterday and there's a major issue I didn't consider... The problem is a lot of packages on npm do not provide cjs builds anymore. So even if I produce esm & cjs modules, they'll fail due to dependencies...

You can see the tentative here https://github.com/SBoudrias/Inquirer.js/tree/multi-module-repo-setup (run `yarn ts-node ./tools/setup-packages.ts` and `yarn lerna run tsc`)

ehaynes99:
"a lot of packages" are all by the same guy who insists, despite overwhelming evidence to the contrary, that projects no longer need to export commonjs. We'd all be better off finding alternatives for any of them used in node.

Following his advice, you could downgrade versions:

```
"globby": "^11.1.0",
"chalk": "^4.1.2",
"ansi-escapes": "^4.3.2",
"wrap-ansi": "^7.0.0",
```

Additionally, since your projects depend on each other, rather than the order from the glob pattern, your `setup-packages` script should run them in topological order. Otherwise, a lib that depends on `core` would see that core doesn't have a commonjs export (yet) and fail. At a glance, I think this would be it:
```typescript
const paths = [
  'packages/type/package.json',
  'packages/testing/package.json',
  'packages/core/package.json',
  'packages/input/package.json',
  'packages/confirm/package.json',
  'packages/checkbox/package.json',
  'packages/editor/package.json',
  'packages/expand/package.json',
  'packages/password/package.json',
  'packages/rawlist/package.json',
  'packages/select/package.json',
];
```

LitoMore:
@ehaynes99 I don't think it's a good solution.

We are using TypeScript, we will lose some TypeScript definitions from our dependencies and it's hard to maintain. Then we have to write a bunch of tools, workflows, declarations, and unuseful codes to support CommonJS. **Now maintainers feel pain.**

LitoMore:
@ehaynes99 Could you tell why you insist on using CommonJS rather than ESM?

Maybe we can find out another solution to accomplish your requirement.

[...]

PR #1185, opened 2023-01-15
Build script to setup ESM/CJS module exports

gfortaine:
fix https://github.com/SBoudrias/Inquirer.js/issues/1159

### DIFF
--- a/packages/inquirer/lib-cjs/inquirer-default.js
+++ b/packages/inquirer/lib-cjs/inquirer-default.js
@@ -1,0 +1,3 @@
+const inquirer = require("./inquirer.js").default;
+
+module.exports = inquirer;
--- a/packages/inquirer/lib/prompts/checkbox.js
+++ b/packages/inquirer/lib/prompts/checkbox.js
@@ -2,9 +2,9 @@
  * `list` type prompt
  */
 
-import chalk from 'chalk';
-import cliCursor from 'cli-cursor';
-import figures from 'figures';
+import chalk from '#chalk';
+import cliCursor from '#cli-cursor';
+import figures from '#figures';
 import { map, takeUntil } from 'rxjs';
 import Base from './base.js';
 import observe from '../utils/events.js';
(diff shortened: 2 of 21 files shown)


packages/inquirer/lib-cjs/inquirer-default.js
```diff
@@ -0,0 +1,3 @@
+const inquirer = require("./inquirer.js").default;
```
SBoudrias, line 1: This file sneaked in the diff (prob added/commited before the `.gitignore` change)

SBoudrias: Oh I see this is desired. Let's update the `.gitignore` so this file is included not ignored by Git.

packages/inquirer/lib/prompts/checkbox.js
```diff
@@ -2,9 +2,9 @@
  * `list` type prompt
  */
 
-import chalk from 'chalk';
-import cliCursor from 'cli-cursor';
-import figures from 'figures';
+import chalk from '#chalk';
```
SBoudrias, line 5: Is there eslint rules to enforce to use the `#` imports if the dependency is found in the package.json import map. That would help enforce it for future collaborators.

SBoudrias: We'll also need to fix the `node/no-missing-import` eslint error before merging.

packages/inquirer/lib-cjs/inquirer-default.js
SBoudrias, line 1: I'd recommend adding this file to the eslint ignore. It errors on CI now since it's using common.js syntax - and wouldn't compile as-is.

SBoudrias:
Thanks a ton for that awesome PR! 🙏🏻 Let me know if you need help to resolve any issues

italojs:
tahnks for this PR, i'm watiting so much for it :)

PR #1201, opened 2023-03-16
feat(select): customize the display text for disabled text

zonemeen:
I have a project that needs to use a custom disabled text for the `Select` type. As shown below

image

### DIFF
--- a/packages/select/src/index.mts
+++ b/packages/select/src/index.mts
@@ -17,7 +17,12 @@ import figures from 'figures';
 import ansiEscapes from 'ansi-escapes';
 
 type SelectConfig = AsyncPromptConfig & {
-  choices: { value: string; name?: string; description?: string; disabled?: boolean }[];
+  choices: {
+    value: string;
+    name?: string;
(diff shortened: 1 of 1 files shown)


packages/select/src/index.mts
```diff
@@ -79,7 +84,9 @@ export default createPrompt<string, SelectConfig>((config, done) => {
     .map(({ name, value, disabled }, index) => {
       const line = name || value;
       if (disabled) {
-        return chalk.dim(`- ${line} (disabled)`);
+        return chalk.dim(
+          `- ${line} (${typeof disabled === 'string' ? disabled : 'disabled'})`
```
SBoudrias, line 88:
```suggestion
          `- ${line} ${typeof disabled === 'string' ? disabled : '(disabled)'}`
```

I think that might be preferable so that we also allow customizing the `()` at the same time.

SBoudrias (approved):

SBoudrias:
Thanks for the PR!

PR #1203, opened 2023-03-22
docs(select): update usage description

zonemeen:


SBoudrias (approved):
Ah great, thanks for following up with the doc!

PR #1204, opened 2023-03-25
fix broken links

henryhu712:


SBoudrias:
Ohh thanks! Missed those 😄

PR #1209, opened 2023-04-09
Fixes a tiny error in the readme.

TheShiningBoots:
The readme contains the lines:
```sh
npm install --save inquirer@^8.0.0
```
which returns an error when executed, due to the `^`, this pr gets rid of the '^' and make it work like how its supposed to.

```
Running npm install --save inquirer@^8.0.0 returns with error
no matches found: inquirer@^8.0.0
```

Note: this is my first PR so there may be some errors or me not following best practices


SBoudrias:
Hi, I just tried this command myself and it works. The `^` character means the latest minor version of the specified major. This should resolve to `8.2.5` today.

Do you have an `.npmrc` or a proxy on npm APIs that would be non-standard? A version should be resolved.

Issue #1207, opened 2023-04-08
Feature request: Support for transformer function for confirm type


remisture:
Could you please add support for transforming the output for confirm types?

```javascript
await inquirer.prompt({
	type: 'confirm',
	name: 'proceed',
	message: 'Continue?',
	default: true,
	transformer: (answer) => (answer ? '👍' : '👎'),
});
```

hjain5164:
@SBoudrias I would like to work on this feature. Can I go ahead?

hjain5164:
I created a [PR](https://github.com/SBoudrias/Inquirer.js/pull/1212) for this issue. @SBoudrias please review and let me know if there are any modifications required.

hjain5164:
@SBoudrias I think this issue can be closed now since the fix is made in the PR #1212

remisture:
Thanks, @hjain5164 :)

PR #1212, opened 2023-04-16
added transform feature for confirm prompts

hjain5164:
PR for the [Feature request: Support for transformer function for confirm type #1207](https://github.com/SBoudrias/Inquirer.js/issues/1207)

The following changes are made:
- Updated the pizza.js file in the examples of confirm prompt.
- Added the transform default function to the Base class to extend this feature to other prompts if and when required. By default, it will return the value as is.
- Updated confirm.js file to allow support of transforming the output, if the user has provided a transformation function.
- Added a test case for this feature as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,7 +269,7 @@ Choices whose property `disabled` is truthy will be unselectable. If `disabled` 
 
 #### Confirm - `{type: 'confirm'}`
 
-Take `type`, `name`, `message`, `transformer`, [`default`] properties. `default` is expected to be a boolean if used.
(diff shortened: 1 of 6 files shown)


SBoudrias:
Hey @hjain5164 thanks for sending a PR!

Would you mind updating the documentation as well?

hjain5164:
@SBoudrias Sure, I will update the documentation as well.

hjain5164:
Hey, @SBoudrias made the changes in the documentation and refactored the code a little bit for better readability and scalability. Let me know if there are any more changes required.

SBoudrias:
Thanks!

PR #1219, opened 2023-05-04
Fix CommonJS file content

### LINKED ISSUE #1218: The dist files for CommonJS are not correct
CleanShot 2023-05-04 at 15 10 58@2x

As you can see here https://www.jsdelivr.com/package/npm/@inquirer/core?tab=files&path=dist%2Fcjs, it runs `require("./lib/screen-manager.mjs")` for importing modules.

This might be related to:

- https://github.com/microsoft/TypeScript/issues/49462

Or we could consider other tools. For example https://github.com/egoist/tsup.

### How to reproduce

```js
// Use CommonJS, and run this line
const {input} = require('@inquirer/prompts');
```

LitoMore:
Fixes #1218 

This is a quick fix for CommonJS files. We could consider using other tools in the future.

### DIFF
--- a/tools/rename-ext.mts
+++ b/tools/rename-ext.mts
@@ -1,5 +1,5 @@
 import path from 'node:path';
-import fs from 'node:fs';
+import fs from 'node:fs/promises';
 import { globby } from 'globby';
 import * as url from 'node:url';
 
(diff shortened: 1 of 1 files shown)


SBoudrias (approved):
Ugh, TS and ESM is so painful 🤦🏻

PR #1220, opened 2023-05-04
Support Node.js 20

LitoMore:
The `yarn test` script will fail on Node.js 20. This is because:

- https://github.com/TypeStrong/ts-node/issues/1997

I changed the command to `node --loader ts-node/esm` for running scripts.

I also added `--no-warnings=ExperimentalWarning` to suppress those experimental warnings.

### Changes

- Add Node.js 20 to GitHub Actions
- Change `fix-ext` to `node --no-warnings=ExperimentalWarning --loader ts-node/esm ../../tools/rename-ext.mts`

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,6 +7,7 @@ jobs:
       fail-fast: false
       matrix:
         node-version:
+          - 20
           - 18
           - 16
           - 14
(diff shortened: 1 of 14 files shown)


SBoudrias (approved):
Thanks!

PR #1222, opened 2023-05-04
Allow async transformer function for input

knpwrs:


SBoudrias:
Hey @knpwrs thanks for the PR!

Render functions aren't async, so they can't rely on `await`. I also think this would be a pretty bug UX issue. If the transform is not synchronous, the typing will feel very laggy.

Do you have more details about what you're trying to achieve?

knpwrs:
Ah, my apologies. I didn't actually test this code, I just took a quick peek at the surrounding context and thought I saw an `async` so I could just `await`.

What I was looking to achieve is to run the result through https://github.com/ranisalt/node-argon2. Now that I think about it I think the proper solution is thus:

```ts
const password = await argon2.hash(await password({ message: 'Password:' }), { type: argon2.argon2id });
```

I didn't even modify the right prompt! Talk about firing from the hip.

Apologies for the PR noise!

PR #1228, opened 2023-05-08
Comment skipped tests out

LitoMore:
To fix the failing CI https://github.com/SBoudrias/Inquirer.js/actions/runs/4910589207.

I'm also checking why the test `prevent returning promises from useEffect hook` fails.

### How to reproduce the failing CI

```sh
# project root
CI=true yarn test
```

The documentation of this behavior can be found at https://vitest.dev/guide/cli.html#options, the `--allowOnly` option.

The `--allowOnly` will be set to `false` when in CI.

### DIFF
--- a/packages/core/core.test.mts
+++ b/packages/core/core.test.mts
@@ -188,52 +188,52 @@ describe('createPrompt()', () => {
 });
 
 describe('Error handling', () => {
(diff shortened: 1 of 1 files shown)


SBoudrias:
Fixed it properly with that commit.

PR #1229, opened 2023-05-09
fix: checkbox ts type prompt error

zonemeen:
The checkbox ts type prompt error.


zonemeen:
No error is reported when using it, it's the typescript prompt in my Webstorm editor that reports an error.

SBoudrias:
Hum weird, it works fine in VSCode for me. Maybe check you're using the workspace typescript version?

I'll close this PR since the solution change the expected behaviour, and it doesn't impact end users/CI.

Issue #1230, opened 2023-05-10
Does `select` in @inquirer/prompts support separators?


tohagan:
The old list feature included a special seperator do you know if that is supported or planned?

SBoudrias:
Hi! This is a feature I haven't ported yet; but I'm planning to.

If you (or another community member) want to send a PR to introduce this go ahead! I'll help review and merging ASAP. The new prompt uses the same paginator component, so it shouldn't be too hard to port over 🤞🏻

SBoudrias:
Hey, this is now added back to the `select` and `checkbox` prompts.

(@zonemeen there's still rawlist where we need to backport the feature if you're willing to help; I'm disconnecting for today! Cheers!)

tohagan:
Oh that's AWESOME!!! Thank you very much!

zonemeen:
Sorry, I didn't notice this message.😂 And whether to add the corresponding test?

SBoudrias:
Yeah unit test is something I want to add to all prompts eventually. With my time for now, I'm still priorizing backward compatibility issues as they come up.

PR #1231, opened 2023-05-12
feat: support separator in select prompt

zonemeen:
Initial solution on https://github.com/SBoudrias/Inquirer.js/issues/1230

If this solution works, I will add it to the rest of the prompts.

### DIFF
--- a/packages/core/src/index.mts
+++ b/packages/core/src/index.mts
@@ -109,11 +109,6 @@ export type ResolvedPromptConfig = {
   validate: (value: string) => boolean | string | Promise<string | boolean>;
 };
 
-export type SeparatorType = {
-  type?: string;
-  separator?: string;
-};
-
 export function createPrompt<Value, Config extends AsyncPromptConfig>(
   view: (
     config: Config & ResolvedPromptConfig,
--- a/packages/core/src/index.mts
+++ b/packages/core/src/index.mts
@@ -7,6 +7,7 @@ import { getPromptConfig } from './lib/options.mjs';
 export { usePrefix } from './lib/prefix.mjs';
 export * from './lib/key.mjs';
 export * from './lib/Paginator.mjs';
+export * from './lib/Separator.mjs';
 
 export type InquirerReadline = readline.ReadLine & {
   output: MuteStream;
@@ -106,6 +107,11 @@ export type AsyncPromptConfig = {
 export type ResolvedPromptConfig = {
   message: string;
   validate: (value: string) => boolean | string | Promise<string | boolean>;
+};
+
+export type SeparatorType = {
+  type?: string;
+  separator?: string;
 };
 
 export function createPrompt<Value, Config extends AsyncPromptConfig>(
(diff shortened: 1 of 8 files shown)


packages/core/src/index.mts
```diff
@@ -108,6 +109,11 @@ export type ResolvedPromptConfig = {
   validate: (value: string) => boolean | string | Promise<string | boolean>;
 };
 
+export type SeparatorType = {
+  type?: string;
```
SBoudrias, line 113: I think this must be `type: 'separator'` since it cannot/shouldn't be anything else.

SBoudrias, line 113: I think this must be `type: 'separator'` since it cannot/shouldn't be anything else.

```suggestion
  type: 'separator';
```

packages/core/src/lib/Separator.mts
```diff
@@ -0,0 +1,17 @@
...
+import figures from 'figures';
+
+/**
+ * Separator object
+ * Used to space/separate choices group
+ */
+
+export class Separator {
+  separator = '';
+  type: string;
```
SBoudrias, line 11:
```suggestion
  separator = chalk.dim(new Array(15).join(figures.line));
  type = 'separator';
```

packages/core/src/lib/Separator.mts
```diff
@@ -0,0 +1,17 @@
...
+
+  constructor(line: string) {
+    this.type = 'separator';
+    this.separator = chalk.dim(line || new Array(15).join(figures.line));
```
SBoudrias, line 15:
```suggestion
  constructor(separator?: string) {
    if (separator) {
        this.separator = separator;
    }
```

packages/select/src/index.mts
```diff
@@ -49,7 +53,11 @@ export default createPrompt<string, SelectConfig>((config, done) => {
       const offset = isUpKey(key) ? -1 : 1;
       let selectedOption;
 
-      while (!selectedOption || selectedOption.disabled) {
+      while (
+        !selectedOption ||
+        selectedOption.disabled ||
+        selectedOption.type === 'separator'
```
SBoudrias, line 59: This logic is reused ~3 times. Let's extract to a utility `isSelectableChoice()`

packages/select/src/index.mts
```diff
@@ -81,14 +93,18 @@ export default createPrompt<string, SelectConfig>((config, done) => {
...
+    .map(({ name, value, disabled, type, separator }, index) => {
       const line = name || value;
       if (disabled) {
         return chalk.dim(
           `- ${line} ${typeof disabled === 'string' ? disabled : '(disabled)'}`
         );
       }
 
+      if (type === 'separator') {
+        return ` ${separator}`;
```
SBoudrias, line 105: Do we need/should we force a whitespace prefix? (genuinely asking if that's what most users would expect; I'm not sure)

SBoudrias:
I went ahead addressing my PR comments. Hoping to get this in today!

Thanks for sending the PR and doing most of the work, it definitively helped!

zonemeen:
Nice work! I'm also happy to be involved in boosting Inquirer.js.

PR #1235, opened 2023-05-14
fix: checkbox readme error

zonemeen:
image


SBoudrias:
Whoops! Thanks for spotting 😂

PR #1238, opened 2023-05-27
prompts: Expose all core functions

LitoMore:
This exposes all core functions to `@inqurier/prompts`.

I believe some people will use our built-in prompts and their self-made prompts at the same time.

In this case, allowing them access to our `core` functions with `inquirer/prompts` is better.

In the future, we can provide and expose more utility functions in this way.


SBoudrias (approved):

SBoudrias:
Good call, I think that's aligned with the goals of the new version.

PR #1240, opened 2023-06-01
Fix (Confirm): Set every other string fallback to default value

### LINKED ISSUE #1239: inquirer/confirm accepts more than just 'Y' and 'n' answers.
When using `confirm`, inquirer will implicitly choose 'n' if the answer does not start with a 'Y' or the answer is not either 'Y' or 'n' explicitly.

For example:
`Yno' -- Implicit Yes
'asdf' -- Implicit No
'nY' -- Implicit No
'Y123' -- Implicit Yes

This bug was initially found in the Create-t3-app cli, which uses Inquirer as a dependency.

LitoMore:
Resolves #1239

I will research how other confirm prompts like `apt` and `ssh` work before marking it as ready.

### Before

|Input|TrueConfirm output|FalseConfirm output|
|:--|:--|:--|
|foo bar|false|false|

### After

|Input|TrueConfirm output|FalseConfirm output|
|:--|:--|:--|
|foo bar|true|false|

### DIFF
--- a/packages/inquirer/test/specs/prompts/confirm.test.js
+++ b/packages/inquirer/test/specs/prompts/confirm.test.js
@@ -75,6 +75,16 @@ describe('`confirm` prompt', () => {
       rl.emit('line', 'Yes');
     }));
 
+  it("should parse 'N' value to boolean false", () =>
+    new Promise((done) => {
+      confirm.run().then((answer) => {
+        expect(answer).toEqual(false);
+        done();
+      });
+
+      rl.emit('line', 'N');
+    }));
+
   it("should parse 'No' value to boolean false", () =>
     new Promise((done) => {
       confirm.run().then((answer) => {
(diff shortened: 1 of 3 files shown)


SBoudrias:
Code LGTM; let me know once you're done with researching state of the art.

packages/inquirer/test/specs/prompts/confirm.test.js
```diff
@@ -85,9 +95,35 @@ describe('`confirm` prompt', () => {
...
+      fixture.default = true;
+      const trueConfirm = new Confirm(fixture, rl);
+
+      trueConfirm.run().then((answer) => {
+        expect(answer).toEqual(true);
+        done();
+      });
+
+      rl.emit('line', 'bla bla foo');
+    }));
```
SBoudrias, line 119: First, I'm happy to merge as-is since those are very old tests. But just FYI, that pattern you see in the test files is just a codemod to move to vitest; not a preference haha.

You could remove the Promise wrapper in favor of

```suggestion
  it('should parse every other string value to default (true)', async () => {
    fixture.default = true;
    const trueConfirm = new Confirm(fixture, rl);

    const answer = trueConfirm.run();

    rl.emit('line', 'bla bla foo');
    await expect(answer).resolves.toEqual(true);
  });
```

LitoMore: OK. I will change them to async/await in another PR.

LitoMore:
### `apt upgrade`

|Input|(Y/n)|(y/N)|
|:--|:--|:--|
|yesterday|yes|yes|
|yellow|yes|yes|
|nope|no|no|
|nintendo|no|no|
|unicorn|yes|no|

### `ssh`

The `ssh` prompts a `(yes/no)`, users must type an exact string `yes` or `no`.

PR #1248, opened 2023-06-07
Fix regex and documentation issues with confirm prompt

uncenter:
### 1. Regex always returning true
Not sure how this issue hasn't been raised yet, but the regex in the confirm prompt is matching *any* input because of the question mark.

**With question mark (current):**
Input of the word no incorrectly evaluates to true

With the question mark, the regex matches the beginning of a line and the pattern `y|yes` 0 or 1 times. Since there will always be the beginning of a line and the `y|yes` pattern is optional, it is always `true`.

**Without question mark (fixed):**
Input of the word no correctly evaluates to false

With*out* the question mark, the regex matches the beginning of a line followed by the **required** pattern `y|yes`, which properly returns the right value.

### 2. Bad example in documentation
Additionally, the message in the example - `"Enter your name"` - makes no sense for a confirm prompt. I just went for `"Continue?"` instead.

### DIFF
--- a/packages/confirm/README.md
+++ b/packages/confirm/README.md
@@ -17,7 +17,7 @@ yarn add @inquirer/confirm
 ```js
 import confirm from '@inquirer/confirm';
 
-const answer = await confirm({ message: 'Enter your name' });
+const answer = await confirm({ message: 'Continue?' });
 ```
 
 ## Options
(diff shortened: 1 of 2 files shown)


SBoudrias (approved):
Oh god, really need to add unit tests to all prompts ASAP 🤦🏻 

Thanks for sending a PR!

uncenter:
> Oh god, really need to add unit tests to all prompts ASAP 🤦🏻
> 
> Thanks for sending a PR!

I would be happy to write a few tests (unless you want to write them yourself)! If I do end up writing some, are the tests for `@inquirer/input` good examples of what we want?

SBoudrias:
@uncenter that'd be awesome yes! The [input prompt tests](https://github.com/SBoudrias/Inquirer.js/blob/master/packages/input/input.test.mts) are a good example indeed. Let me know if you get to it, I don't think I'll have much time myself in the next ~2 weeks.

PR #1265, opened 2023-07-16
Added error callback for when we fail in onEnd

mastrzyz:


SBoudrias:
Hi @mastrzyz, thanks for sending a PR.

Could you help fixing the CI issues? And instead of an examples, adding a unit test would be more resilient - so it'd be awesome if you have time to add that in!

mastrzyz:
Hey @SBoudrias  , fixed CI and added a UT!, let me know if this is fine or you would like one for each prompt.

PR #1268, opened 2023-07-21
chore: replace `through` with `@ljharb/through`

### LINKED ISSUE #1266: update to a maintained `through`?
Hi! You depend on `through`, an excellent package that unfortunately has not been maintained since 2018. I've forked it as https://npmjs.com/@ljharb/through so that I can maintain it (with @dominictarr's blessing), and am using it in https://npmjs.com/tape.

Would you be interested in a PR that switches this project to using my fork instead of the original? It's effectively identical, but it'd ensure any vulnerabilities will be addressed in the future.

ljharb:
Fixes #1266

`through` is still pulled in via some dev deps, but at least the direct dep is now maintained :-)

### DIFF
--- a/packages/inquirer/lib/ui/bottom-bar.js
+++ b/packages/inquirer/lib/ui/bottom-bar.js
@@ -2,7 +2,7 @@
  * Sticky bottom bar user interface
  */
 
-import through from 'through';
+import through from '@ljharb/through';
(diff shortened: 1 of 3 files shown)


LitoMore (approved):
@ljharb Thank you!

SBoudrias (approved):

PR #1269, opened 2023-07-22
Feat (select): Export types

tuler:
Export types so it's more reusable

### DIFF
--- a/packages/select/src/index.mts
+++ b/packages/select/src/index.mts
@@ -17,7 +17,7 @@ import chalk from 'chalk';
 import figures from 'figures';
 import ansiEscapes from 'ansi-escapes';
 
-type Choice<Value> = {
+export type Choice<Value> = {
   value: Value;
   name?: string;
   description?: string;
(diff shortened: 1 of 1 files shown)


SBoudrias:
Hey! Thanks for sending this PR. I'm not really looking forward to exporting the types because it creates a larger API surface we must manage and keep backwards compatible. I'd be happy to expose some (thinking of utilities to extract types) - but I'd need example use cases to design them!

Would you mind expanding on your use case? (Code/pseudo code would be super helpful!)

tuler:
IHMO when you export a type you don't add much to the API surface.
A program written for `@inquirer/select:1.2` already have to conform to its API, despite the fact the type is exported.
It's true you are more free to rename the type itself, but if you rename a field, add a required field, or remove a field, you are already breaking the API.

The use case is when you write typescript auxiliary code that must conform to the API. It's more elegant to be typed IMHO.

SBoudrias:
I hope this doesn't come up as condescending, but I've maintained large OSS projects that for a while now and I'm not sure you gathered the nuances of choosing the exposed API.

Also, I'm not saying I wouldn't address your concern; it is 100% valid and I want to provide a solution. But there's other approaches I want to explore first since I think they'll work better for Inquirer.

So I'll relaunch my original follow-up:

> Would you mind expanding on your use case? (Code/pseudo code would be super helpful!)

PR #1271, opened 2023-07-25
chore: directly import only needed lodash modules

lukecotter:
Changed  `import _ from 'lodash';`
to
```
import filter from 'lodash/filter.js';
import map from 'lodash/map.js';
```

This reduces the bundle size and will help bundlers with tree shaking.

### DIFF
--- a/packages/inquirer/lib/objects/choices.js
+++ b/packages/inquirer/lib/objects/choices.js
@@ -1,5 +1,6 @@
 import assert from 'node:assert';
-import _ from 'lodash';
+import filter from 'lodash/filter.js';
+import map from 'lodash/map.js';
 
 import Separator from './separator.js';
 import Choice from './choice.js';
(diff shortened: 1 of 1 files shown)


SBoudrias:
Hi, there's a few things I'd recommend taking into consideration:

1. If you care about bundle size, I strongly recommend moving to the new Inquirer release (and package). It is dramatically smaller (thanks to removing rx.js for the most part). See [release announcement here](https://github.com/SBoudrias/Inquirer.js/discussions/1214).
2. I have nothing against getting this change in; but realistically, pushing that on libraries will be very hard. So projects should have some webpack plugins to rewrite lodash imports (likely to be too many transient-dependencies to handle; and anyway, a regression in a single dependency will bring you back to where you were...)
3. I'd love if you could also send a follow up PR including an eslint rule enforcing the change you've made to make sure we don't have regression on Inquirer at least!

Thanks for the PR, I'll cut a new v9 release soon.

lukecotter:
The eslint rule is a good idea. I will add it to my list.
I will definitely look at switching to the new Inquirer release thanks for pointing it out.

PR #1272, opened 2023-07-26
Type: password config type

sjx1995:
In the [password config](https://github.com/SBoudrias/Inquirer.js/tree/master/packages/password),  the transformer and default attributes should not be included. 

The original code inherited the type of the input config, I used Omit to remove these two attributes so that users can receive more user-friendly and accurate prompts.


SBoudrias (approved):
Great catch, thanks for sending a PR!

PR #1279, opened 2023-08-08
Add script to validate the version for publishing

### LINKED ISSUE #1278: Wrong version released as latest
image

LitoMore:
Resolves #1278

I used the `.mjs` for this script since we don't need to export it to our users, and it doesn't import any `.mts` files. It's faster and easier to execute.

I only added the `prepublishOnly` to the `inquirer` package. Please let me know if you want to add this script to all packages.

BTW, we can consider migrating our other scripts from `.mts` to `.mjs` since TypeScript is not a necessary runtime.


SBoudrias:
Did you test it against Lerna? What happen if the publish is rejected by the script?

Lerna has a tendency to get stuck if something fails while it runs.

LitoMore:
Ah, it's tough. The `lerna publish` always creates a tag before publishing even if it fails.

I think the solution to avoid this is add another that wraps the `lerna publish` command. For example:

```js
// package.json
{
	"scripts": {
		"publish": "node ./validate-version.mjs && lerna publish"
	}
}
```

And use `yarn publish` instead.

SBoudrias:
@LitoMore Good idea, I like this solution. Thanks!

We'll need to port the script to the https://github.com/SBoudrias/Inquirer.js/tree/v8.0-legacy branch too if we want to prevent v8 fix releases to go out as `latest`.

LitoMore:
I just realized it's impossible to check the version selected after the `lerna publish` prompt step.

I also checked Lerna's API, there is no way to add some scripts after the `lerna publish`.

Closing for now.

PR #1280, opened 2023-08-08
Migrate `.mts` scripts to `.mjs`

LitoMore:
Related to https://github.com/SBoudrias/Inquirer.js/pull/1279#issue-1840862449.

Here is the build time difference:

|Before|After|
|:-:|:-:|
|||

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,10 +73,9 @@
     "integration/*"
   ],
   "scripts": {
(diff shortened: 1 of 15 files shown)


SBoudrias (approved):
LGTM! Thanks for catching that. I agree TS isn't critical here and not worth the perf cost.

PR #1285, opened 2023-08-19
Refactor (Core): Move from scoped variables to node built-in AsyncLocalStorage

SBoudrias:
Rationale: State management through the scoped variables did (and could again, or might still do so in narrow cases) cause leaks in context around async operations and cleanup phases. I think using AsyncLocalStorage might improve this caveats by ensuring a stable context.

### DIFF
--- a/packages/core/core.test.mts
+++ b/packages/core/core.test.mts
@@ -381,23 +381,6 @@ describe('Error handling', () => {
     await expect(answer).rejects.toThrowError('Error in render function');
   });
 
-  it('Prevent trying to run 2 prompts at once.', async () => {
-    const Prompt = () => '';
-    const prompt = createPrompt(Prompt);
-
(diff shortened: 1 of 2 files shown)


packages/core/core.test.mts
```diff
@@ -381,23 +381,6 @@ describe('Error handling', () => {
     await expect(answer).rejects.toThrowError('Error in render function');
   });
 
-  it('Prevent trying to run 2 prompts at once.', async () => {
```
SBoudrias, line 384: Side effect, it becomes theoretically possible to run 2 inquirer prompts at once.

This is unlikely to have much benefits but:
1. Could be theoretically possible to run prompts on different streams (if Node Readline allow it - it used to have singleton conflicts)
2. When running tests today, if a test fail Inquirer doesn't cleanup. This won't happen anymore 🎉

PR #1289, opened 2023-08-28
Feat: Add support for `loop: false` in `checkbox` and `select`

artfuldev:
In this PR, we add support for `loop: false` in checkbox and select prompts.

In addition, we also add navigation and speed dial as separate hooks which can be used along with pagination to compose up custom prompts. Because they are composable, they support use cases of navigation and speed dial without pagination.

In the revised pagination, we also render only a maximum of pageSize elements at any time. We also account for the render taking up multiple lines for a single item, retaining the cursor position as expected. This results in 2 benefits:

1. It's more performant (as we don't render every line - we only render a page of items)
2. It doesn't cause issues where the cursor position is greater than 1, but a previous item (not the first, or index 0) has several lines of output causing the cursor position (line) to not even be visible.


packages/core/src/index.mts
```diff
@@ -181,41 +182,43 @@ export function usePagination(
...
-  const infinite = [lines, lines, lines].flat();
-  const topIndex = Math.max(0, active + lines.length - state.current.pointer);
-
-  const section = infinite.splice(topIndex, pageSize).join('\n');
-  return section + '\n' + chalk.dim('(Move up and down to reveal more choices)');
+  const [lastActive, setLastActive] = useState(active);
+  useEffect(() => {
+    setLastActive(active);
+  }, [active, setLastActive]);
+
```
artfuldev, line 201: Just trying to separate state and make it more idiomatic. Please let me know if there are any concerns.

packages/core/src/index.mts
```diff
@@ -181,41 +182,43 @@ export function usePagination(
...
+  const [position, setPosition] = useState(0);
+  useEffect(() => {
+    setPosition(
+      (loop ? infinite : finite)({
+        active: { current: active, previous: lastActive },
+        total: lines.length,
+        pageSize,
+      })(position),
+    );
+  }, [loop, active, pageSize, setPosition, position, lastActive, lines.length]);
```
artfuldev, line 211: I'm trying to share the idea that there is a specific position at which we want to render the active item hence the rename from `pointer` to `position`. Would you agree?

packages/core/src/lib/position.mts
```diff
@@ -0,0 +1,49 @@
...
+  active: Change<number>;
+  total: number;
+  pageSize: number;
+};
+
+/**
+ * Given information about a page, decides the next position at which the active
+ * item should be rendered in the page.
+ */
+type PositionReducer = (info: PageInfo) => (pointer: number) => number;
```
artfuldev, line 16: If `set` from `useState` someday (I hope soon) takes a function to do atomic updates instead of values, this will be even more useful

packages/core/src/lib/position.mts
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Creates the next position for the active item considering an infinitely
+ * looping list of items to be rendered on the page.
+ */
+export const infinite: PositionReducer =
+  ({ active, total, pageSize }) =>
+  (pointer) =>
+    total <= pageSize
+      ? active.current
```
artfuldev, line 41: I initially used a finite call here but changed my mind. Please let me know what you think.

packages/core/src/lib/utils.mts
```diff
@@ -16,3 +16,18 @@ export const breakLines = (content: string, width: number): string =>
...
+/**
+ * Creates a 0-based index out of an integer, wrapping around if necessary.
+ */
+const index = (max: number) => (value: number) => ((value % max) + max) % max;
+
+/**
+ * Rotates an array of items by an integer number of positions.
+ */
+export const rotate =
+  (count: number) =>
```
artfuldev, line 29: Can be positive, negative, larger than the size of the array in either direction.

packages/select/select.test.mts
```diff
@@ -152,6 +152,34 @@ describe('select prompt', () => {
...
+
+    expect(getScreen()).toMatchInlineSnapshot(`
+      "? Select a number (Use arrow keys)
+      ❯ 1
+        2
+      (Move up and down to reveal more choices)"
+    `);
+
+    events.keypress('up');
+    events.keypress('up');
```
artfuldev, line 171: keypresses are presently handled by the select component itself, causing this test to fail.

packages/core/src/index.mts
```diff
@@ -181,41 +182,43 @@ export function usePagination(
...
+
+  // Rotate lines such that the active index is at the specified position
+  return rotate(active - position)(lines)
+    .slice(0, pageSize)
+    .concat(
+      lines.length <= pageSize
+        ? []
+        : [chalk.dim('(Move up and down to reveal more choices)')],
+    )
+    .join('\n');
```
artfuldev, line 221: All existing tests (which support looping) pass presently, so this logic is tested.

packages/core/src/lib/position.mts
```diff
@@ -0,0 +1,49 @@
...
+export const finite: PositionReducer =
+  ({ pageSize, total, active }) =>
+  () => {
+    const middle = Math.floor(pageSize / 2);
+    return total <= pageSize || active.current < middle
+      ? active.current
+      : active.current >= total - middle
+      ? active.current + pageSize - total
+      : middle;
+  };
```
artfuldev, line 31: For example,
for page size `5`
and items
`[A, B, C, D, E, F, G, H, I , J]`
will create corresponding positions
`[0, 1, 2, 2, 2, 2, 2, 2, 3, 4]`

packages/core/src/index.mts
```diff
@@ -181,41 +182,43 @@ export function usePagination(
...
+    );
+  }, [loop, active, pageSize, setPosition, position, lastActive, lines.length]);
+
+  // Rotate lines such that the active index is at the specified position
+  return rotate(active - position)(lines)
+    .slice(0, pageSize)
+    .concat(
+      lines.length <= pageSize
+        ? []
+        : [chalk.dim('(Move up and down to reveal more choices)')],
```
artfuldev, line 219: When not looping, do we need a different message?

For example, `(Move down...` and `(Move up...` depending on which direction reveals more options?

SBoudrias: Maybe we could rephrase to something like:

> Use arrow keys to reveal more choices

So that up/down doesn't cause confusion on specific cases while still being applicable 🤔

packages/core/src/index.mts
SBoudrias: Setting state from an effect is often seen as an anti-pattern in React since it'll cause extra re-renders. This will also be true for Inquirer; and somewhat more so since frequent re-renders definitively can cause the prompt to flicker.

I didn't dig deep in the code yet; but if you need to keep a state between renders, I'd recommend `useRef` instead.

SBoudrias:
> However, I've encountered a challenge: the current paginator primarily operates on string outputs and doesn't handle key presses directly.

This was also true with the class based Inquirer `Paginator`; did you read the old code to see how it worked? Is it re-implemented in each prompt?

packages/core/src/lib/position.mts
```diff
@@ -0,0 +1,49 @@
...
+ */
+type PositionReducer = (info: PageInfo) => (pointer: number) => number;
+
+/**
+ * Creates the next position for the active item considering a finite list of
+ * items to be rendered on a page.
+ */
+export const finite: PositionReducer =
+  ({ pageSize, total, active }) =>
+  () => {
```
SBoudrias, line 24: More of a `nit`, but I don't see the value of most of those curried functions (with their current usage inside the hook.)

It feels like unnecessary complexity VS passing 2 arguments to the function.

SBoudrias:
Thanks for sending a first draft! FYI, I won't have time to make another review pass for the next few days - should be able to do one more ~next weekend at the latest.

packages/core/src/index.mts
artfuldev: I've move this back to refs, thanks!

packages/core/src/lib/position.mts
artfuldev: I agree, I was more thinking about `setPointer(current => next)` usage at some point, when being able to do
`setPointer(infinite(args))` might be useful. I'll bring this forward after a few more iterations and you can see if you like it. Otherwise we'll undo the curried version and make it simpler.

packages/core/src/index.mts
artfuldev: Sure, we could try that.

artfuldev:
> Thanks for sending a first draft! FYI, I won't have time to make another review pass for the next few days - should be able to do one more ~next weekend at the latest.

No issues, thank you so much for your feedback and input! I myself will probably need some time to make further changes, but it's great to know you're available for inputs.

artfuldev:
Hi @SBoudrias!

I'd like some feedback before I wrap up. Presently we're trying to use the pagination hook to render only a minimal set of items (some of which may eventually be rendered, mostly because an item's render may result in multiple lines). I hope the test specifications for `lines.test.mts` makes it clear the behavior we want. If you can validate that, I can proceed with the implementation (I want to make sure you are okay with a behavior before implementing it - as you may have different opinions or ideas that will help shape the actual implementation). Please let me know how I can help.

packages/checkbox/checkbox.test.mts
```diff
@@ -34,7 +34,7 @@ describe('checkbox prompt', () => {
        ◯ 5
        ◯ 6
        ◯ 7
-      (Move up and down to reveal more choices)"
+      (Use arrow keys to reveal more choices)"
```
SBoudrias, line 37: I made this change on the master branch to reduce the amount of diff noise generated on your PR. Do you mind re-merging master on top of your changes?

packages/select/src/choice.mts
```diff
@@ -0,0 +1,14 @@
+import { Separator } from '@inquirer/core';
+
+export type Choice<Value> = {
```
SBoudrias, line 3: I do appreciate trying to clean some code, but it'll be much easier for me to review a small diff.

I'm not against reviewing some refactors; but those should go into separate PRs. When adding a new feature, it helps a lot to keep the diff minimal and surgical to lower the review burden.

packages/core/src/index.mts
```diff
@@ -36,7 +35,7 @@ type HookStore = {
 
 const hookStorage = new AsyncLocalStorage<HookStore>();
 
-const context = {
+export const context = {
```
SBoudrias, line 38: I specifically do not want to expose the hooks internals outside the core package. I think you're exporting it to reuse in the new utils files you created; but the side effect is exposing it through the npm package too.

If we need to do this in this PR, let's do a refactor only that'll extract the hooks code to a separate file first.

artfuldev: Oops, I didn't expect to get this part merged in - this branch is more WIP. I'm aware of the side effect.

That's an excellent suggestion; I'd like to open that PR separately right away. It would be nice to rebase with main once that's merged in.

packages/select/src/choice.mts
artfuldev: I apologize for the cognitive overhead.

packages/checkbox/checkbox.test.mts
artfuldev: That's great, thank you so much! I'll do that.

artfuldev:
Hi @SBoudrias !

Thank you so much for your review comments. I apologize for the several changes that are also WIP. I appreciate the feedback and will incorporate those.

I wish to highlight that I was looking for feedback around the behaviour of the lines being rendered when a single page has items that may be rendered across several lines. It'd be great if you can take a look at that and share your thoughts on the expected behavior. Please let me know if the part isn't clear and I can try to add more information. The specific expected behaviors are those expressed in `lines.test.mts`. This has to do with multi-line renders given a CLI width of items, where the expected cursor position is sometimes not enough to render the items accurately. This might be an edge case we don't want to worry about; or you might have certain opinions so I think it's best to have your supervision and feedback on this specific part first.

I'll clean up the PR and incorporate your suggestions and come back with a PR that's ready for review (this is still draft). Thanks again for your review and pointers!

packages/core/src/index.mts
artfuldev: I've opened a separate PR for this in #1292 - I will rebase my changes once that gets merged in.

artfuldev:
I'll be working on incorporating your feedback while waiting for comments on the expected behaviour in `lines.test.mts`. Thanks!

packages/core/src/lib/pagination/lines.mts
```diff
@@ -0,0 +1,43 @@
+import { rotate, splitLines } from '../utils.mjs';
+import { Paged } from './types.mjs';
+
+type Inputs<T> = {
+  items: readonly T[];
+  width: number;
+  render: (paged: Paged<T>) => string;
+  active: number;
+  position: number;
```
SBoudrias, line 9: What is the difference between those 2 parameters?

packages/core/src/lib/pagination/lines.mts
```diff
@@ -0,0 +1,43 @@
+import { rotate, splitLines } from '../utils.mjs';
+import { Paged } from './types.mjs';
```
SBoudrias, line 2: I don't find this name particularly descriptive. I think this is an area around which I'd try to simplify the implementation (the data structure passed around feels large.)

packages/core/src/lib/pagination/lines.mts
```diff
@@ -0,0 +1,43 @@
...
+  active,
+  position,
+  pageSize,
+}: Inputs<T>): string[] => {
+  const split = splitLines(width);
+
+  const indexed = items.map((item, index) => ({ item, index }));
+  const slice = rotate(active - position)(indexed).slice(0, pageSize);
+  const previous = slice
+    .filter((_, i) => i < position)
```
SBoudrias, line 32: nit: We should be able to `.slice()` this array (same a bit lower in the file)

artfuldev: Of course - it's simpler and clearer/idiomatic

packages/core/src/lib/pagination/lines.mts
artfuldev: `active` is the item in the list we need to point at; `position` is the position at which we expect the active item to be rendered. for example, for single-line renders of alphabets, active 2 position 3 is:

```
  Z
  A
  B
> C
  D
```

and active 2 position 1 is:

```
  B
> C
  D
  E
  F
```

for page size of 5.

artfuldev: I can add documentation to the parameters on the type, if that works.

packages/core/src/lib/pagination/lines.mts
artfuldev: Just to be sure - are you concerned that `Paged<T>` is large? I'm a little unsure as it's made of 3 properties, one of them being optional. I understand that the name needs work; I'll put more thought into it and come up with a revision.

artfuldev:
So @SBoudrias I take it you're in general in agreement with the specifications in the `lines.test.mts` file?

artfuldev:
@SBoudrias Are we committing `.yarn/cache` intentionally?

artfuldev:
I don't think my changes in particular are causing the build to break - I need some help 😅

SBoudrias:
> I take it you're in general in agreement with the specifications in the `lines.test.mts` file?

At a high level, yes. But I don't think this is the signal you're looking for.

As in, sure the function seems to work fine and do what it's supposed to do. But is that the right approach/architecture? I'm not sure. I feel we're adding a lot of complexity; and ultimately I would prefer to not add new hooks (though I didn't dive deeper in those.)

If we can dumb down some of the implementation (the currying for example seem more clever than readable), we'll see better where we stand as of now.

> Are we committing `.yarn/cache` intentionally?

Yes, this is keep dependencies locally so they're not reinstalled on CI. It's a mode from yarn 2. Not a huge fan of the extra weight stored in the repo, but works good enough and CI/clones are fairly fast since we skip further installs.

> I need some help

Yeah super weird issue that Typescript started to catch type issues inside vitest... I didn't dig into that yet, but might be from some import.

artfuldev:
Well, take your time and see how you feel about the added complexities and what changes we can make.

I only hope to contribute in a meaningful way and I have no intention of trying to change your opinion on any patterns/structure. I'm happy to rework the PR in any way that suits your goals/vision/preferences.

We have added 2 hooks (in addition to reworking pagination):
1. `useScroll`
2. `useSpeedDial`
these are both well defined behaviours which are optional, and can work irrespective of pagination being used, including in custom prompts. I also have suggestions for things like speed dial and how indexing can work, but didn't wish to bring too many ideas into this PR.

I would rework `breakLines` for 2 reasons:
1. it overcommits a join('\n') into a string[] making it harder to reuse the outputs or compose, since we need to split on its outputs again if we need to operate on the array. it's simpler to return an array and recombine it in consuming code or expose another function which composes the breaking and the joining.
2. its ordering of the arguments makes it harder to partially apply arguments - we ideally want to break different content strings with the same width, and not the same content across different widths.

If you need me to simplify the PR so you can take another look, I'm happy to do that. I can also leave it as it is and let you take a look as and when possible and share your refined thoughts.

packages/core/src/lib/pagination/lines.mts
```diff
@@ -0,0 +1,79 @@
...
+  layoutIndex = requested - 1;
+  while (lineNumber >= 0 && layoutIndex >= 0) {
+    for (const line of split(render(layoutsInPage[layoutIndex])).reverse()) {
+      page[lineNumber--] = line;
+      if (lineNumber < 0) break;
+    }
+    layoutIndex--;
+  }
+
+  return page.slice(0, pageSize).filter((line) => typeof line === 'string');
```
artfuldev, line 78: We iteratively render the lines in the page, only rendering items when necessary. for example, if a single active item can fill the entire page with its lines, no more items will be rendered. Given we have tests, we're able to prove this works. Regardless, this is a pretty simple algorithm. We first render the lines of the active item at the expected position. Then, we fill the next and previous lines iteratively.

packages/checkbox/src/choice.mts
```diff
@@ -0,0 +1,22 @@
+import { Separator } from '@inquirer/core';
```
SBoudrias, line 1: Let's go back on that split and keep a single file per prompt.

packages/core/src/lib/types.mts
```diff
@@ -0,0 +1,12 @@
...
+export type Action<T> = UnaryF<T, void>;
+
+export type Activatable<T> = {
+  active: T;
+  setActive: Action<T>;
+};
+
+export type HasSeveralOrdered<T> = {
+  items: readonly T[];
+};
```
SBoudrias, line 12: Let's inline those types as much as possible (even if it leads to some duplication; not "DRY")

packages/core/src/lib/utils.mts
```diff
@@ -16,3 +16,18 @@ export const breakLines = (content: string, width: number): string =>
...
+
+/**
+ * Creates a 0-based index out of an integer, wrapping around if necessary.
+ */
+export const index = (max: number) => (value: number) => ((value % max) + max) % max;
+
+/**
+ * Rotates an array of items by an integer number of positions.
+ */
+export const rotate =
```
SBoudrias, line 28: Let's simplify those 2 functions by removing the currying. Let's make them a function with multiple arguments.

packages/core/src/lib/pagination/position.mts
```diff
@@ -0,0 +1,45 @@
...
+ * Creates the next position for the active item considering a finite list of
+ * items to be rendered on a page.
+ */
+export const finite: PositionReducer = ({ pageSize, total, active }) => {
+  const middle = Math.floor(pageSize / 2);
+  return total <= pageSize || active.current < middle
+    ? active.current
+    : active.current >= total - middle
+    ? active.current + pageSize - total
+    : middle;
```
SBoudrias, line 28: Let's rewrite that with `if`/`else` instead of nested ternary. (same comment for `infinite`)

packages/checkbox/src/index.mts
```diff
@@ -4,154 +4,86 @@ import {
...
     const [status, setStatus] = useState('pending');
-    const [choices, setChoices] = useState<Array<Separator | Choice<Value>>>(() =>
-      config.choices.map((choice) => ({ ...choice })),
+    const [items, setItems] = useState<ReadonlyArray<Item<Value>>>(
+      choices.map((choice) => ({ ...choice })),
     );
-    const [cursorPosition, setCursorPosition] = useState(0);
     const [showHelpTip, setShowHelpTip] = useState(true);
+    const message = chalk.bold(config.message);
+    const { contents, active, setActive } = usePagination<Item<Value>>({
```
SBoudrias, line 43: Can we bring back `useState` at the component level?

And can we go back on the introduction of `useSpeedDial` & `useScroll` (e.g keeping the keypress handler in each prompt.) Those 2 hooks are adding a fairly large diff, and I'm not convinced it's a positive change for the codebase. It's DRYing stuff, but introducing complexity that doesn't feel necessary.

SBoudrias (commented):
Hey, I been punting on reviewing this week since I felt like I'd run out of time. I don't want to make you wait too long, so I'm making a first pass today where I'm focusing on simplifying the code change and reducing the diff.

I want this PR to go down in size; and we can review at a later point if we want to introduce more refactor or not to the codebase (though from the onset I'm not convinced it's necessary.) If so, that would need to be separate PRs.

Hopefully that's a good first bit to make some progress on really implementing minimal changes to the codebase in order to add the `loop` option.

Let me know if you have questions or need help. For the TS issue, I suspect as you simplify the implementation & typing it might go away.

artfuldev:
> Hey, I been punting on reviewing this week since I felt like I'd run out of time. I don't want to make you wait too long, so I'm making a first pass today where I'm focusing on simplifying the code change and reducing the diff.
> 
> I want this PR to go down in size; and we can review at a later point if we want to introduce more refactor or not to the codebase (though from the onset I'm not convinced it's necessary.) If so, that would need to be separate PRs.
> 
> Hopefully that's a good first bit to make some progress on really implementing minimal changes to the codebase in order to add the `loop` option.
> 
> Let me know if you have questions or need help. For the TS issue, I suspect as you simplify the implementation & typing it might go away.

Thank you so much for a first pass! If it makes things any easier for you, I'm in no hurry and can wait for as long as required for you to review the PR at your own pace.

artfuldev:
I _think_ I've done everything. Hopefully worth a second pass. There are some minor rewrites and renames - I'm happy to undo them. But hopefully they're not as taxing and can be accommodated. Happy to have your feedback again and rework the PR as necessary.

PS: The TS errors remain.

artfuldev:
I would like to add more tests for edge-cases around rendering lines:
for example:

only 2 items, page size 5, first item is 4 lines second item is 1 line, loop false

should render 

```
  1 line 1
  1 line 2
  1 line 3
  1 line 4
> 2 line 1
```

Let me know if you think these are valuable or excessive.

packages/checkbox/src/index.mts
```diff
@@ -24,134 +26,111 @@ export type Choice<Value> = {
...
-
+    const { prefix = usePrefix(), instructions, pageSize, loop = true, choices } = config;
     const [status, setStatus] = useState('pending');
-    const [choices, setChoices] = useState<Array<Separator | Choice<Value>>>(() =>
-      config.choices.map((choice) => ({ ...choice })),
+    const [items, setItems] = useState<ReadonlyArray<Item<Value>>>(
+      choices.map((choice) => ({ ...choice })),
     );
-    const [cursorPosition, setCursorPosition] = useState(0);
+    const [active, setActive] = useState<number>(() => {
```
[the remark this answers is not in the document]
SBoudrias: Could you open 2 separates PRs (checkbox and select) that only include the renaming & refactors (the util functions extraction)?

We can get that merged in, and get this PR more focused on the looping logic.

packages/core/src/lib/utils.mts
```diff
@@ -16,3 +16,20 @@ export const breakLines = (content: string, width: number): string =>
         .map((str) => str.trimEnd()),
     )
     .join('\n');
+
+/**
+ * Creates a 0-based index out of an integer, wrapping around if necessary.
+ * @param {number} max The maximum count
+ * @param {number} value The value to convert to index
+ */
+export const index = (max: number, value: number) => ((value % max) + max) % max;
```
SBoudrias, line 25: I'd prefer not exposing an extra (and small) helper from `@inquirer/core`. Let's inline and duplicate the utility in both prompts.

packages/select/src/index.mts
```diff
@@ -11,113 +11,101 @@ import {
...
   disabled?: boolean | string;
   type?: never;
 };
 
+type Item<Value> = Separator | Choice<Value>;
+
+const selectable = <Value,>(item: Item<Value>): item is Choice<Value> =>
+  !Separator.isSeparator(item) && !item.disabled;
+
+const render = <Value,>({ item, active }: Layout<Item<Value>>) => {
```
SBoudrias, line 35: Similar to `index`, I'd prefer to not expose extra utility types from `@inquirer/core`. Would types be inferred properly without typing it?

Otherwise, let's just define locally `item` & `active` at the cost of cuplication.

packages/checkbox/src/index.mts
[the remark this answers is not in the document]
artfuldev: I've opened #1299 for checkbox. I'll open one for select soon.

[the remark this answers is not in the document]
artfuldev: The changes here should be simpler once those are merged in. Feel free to review those PRs fresh; we can pause on this PR until those are to your satisfaction.

packages/checkbox/src/index.mts
```diff
@@ -52,7 +53,7 @@ function check(checked: boolean) {
   };
 }
 
-function renderItem<Value>({ item, isActive }: { item: Item<Value>; isActive: boolean }) {
```
artfuldev, line 55: If we want to change this to `isActive` it needs to be changed in the `Layout` type. It also needs to be consistent between `select` and `checkbox` - presently `select` uses `active` while `checkbox` uses `isActive` - I guess I missed making this consistent

artfuldev:
Now we have changes specific to this loop setting - although I think we should incorporate the additional test cases as mentioned here https://github.com/SBoudrias/Inquirer.js/pull/1289#issuecomment-1712467084 before merging. It needs additional implementation but we could always add it later as an enhancement.

artfuldev:
Finally figured out the TS compilation issue - it was the unit test for `lines`. I'm a little unsure what you want to do for unit tests like these. In projects I'm familiar with, unit tests are written as `file.spec.[m]ts` for `file.[m]ts` and we ignore `*.spec.[m]ts` files from `src` - unit tests are considered part of source (they may not be testing a part of the public API). The package setup validation now fails as I've changed the tsconfig.json file. I'll let you decide how you want to proceed. We could also write a custom prompt based test that uses pagination in `core.test.mts` instead of unit testing `lines`.

packages/core/tsconfig.cjs.json
```diff
@@ -1,6 +1,5 @@
 {
   "extends": "./tsconfig.json",
-  "include": ["./src"],
```
artfuldev, line 3: This is already included in the `tsconfig.json` file we extend.

packages/core/README.md
```diff
@@ -135,6 +135,7 @@ export default createPrompt((config, done) => {
   const windowedChoices = usePagination(allChoices, {
     active: cursorPosition,
     pageSize: config.pageSize,
+    loop: config.loop,
```
SBoudrias, line 138: Making a note for later, we'll need to document `render` & `items` here.

packages/checkbox/src/index.mts
```diff
@@ -120,13 +123,12 @@ export default createPrompt(
...
 
-    const lines = items
-      .map((item, index) => renderItem({ item, isActive: index === active }))
-      .join('\n');
-
-    const page = usePagination(lines, {
+    const page = usePagination<Item<Value>>({
+      items,
       active,
+      render: renderItem,
```
SBoudrias, line 129: I don't think `render` is a good name here since it renders a single item; so I'd keep the name `renderItem` inside and outside.

packages/core/src/index.mts
```diff
@@ -4,7 +4,7 @@ export { useState } from './lib/use-state.mjs';
 export { useEffect } from './lib/use-effect.mjs';
 export { useRef } from './lib/use-ref.mjs';
 export { useKeypress } from './lib/use-keypress.mjs';
-export { usePagination } from './lib/use-pagination.mjs';
+export * from './lib/pagination/index.mjs';
```
SBoudrias, line 7: To revert

packages/core/tsconfig.cjs.json
SBoudrias: This is what make the CI fail. Those files are auto-generated by a script, so you'd need to modify that script in `/tools/`.

But it doesn't really matter, so let's just keep this out of this PR.

packages/core/src/lib/pagination/use-pagination.mts
```diff
@@ -0,0 +1,52 @@
+import chalk from 'chalk';
+import cliWidth from 'cli-width';
+import { readline } from '../hook-engine.mjs';
+import { useRef } from '../use-ref.mjs';
+import { lines } from './lines.mjs';
+import { finite, infinite } from './position.mjs';
+import { type Layout } from './layout.type.mjs';
+
+export type Options<T> = {
```
SBoudrias, line 9: No need to be exported.

packages/core/src/lib/pagination/lines.mts
```diff
@@ -0,0 +1,80 @@
+import { breakLines, rotate } from '../utils.mjs';
+import { type Layout } from './layout.type.mjs';
+
+type Inputs<T> = {
+  items: readonly T[];
+  /** The width of a rendered line in characters. */
+  width: number;
+  /** Renders an item as part of a page. */
+  render: (layout: Layout<T>) => string;
```
SBoudrias, line 9: Let's `Prettify<Layout<T>>` (util from `@inquirer/types`) here so the exposed type is the raw object and not a named type.

packages/select/src/index.mts
```diff
@@ -17,7 +17,7 @@ import chalk from 'chalk';
 import figures from 'figures';
 import ansiEscapes from 'ansi-escapes';
 
-type Choice<Value> = {
+export type Choice<Value> = {
```
SBoudrias, line 20: Let's revert this line (I'd prefer to not export types interfaces that end up being part of the API contract.)

```suggestion
type Choice<Value> = {
```

SBoudrias (commented):
Small morning pass

packages/core/src/lib/pagination/layout.type.mts
```diff
@@ -0,0 +1,6 @@
+/** Represents an item that's part of a layout, about to be rendered */
+export type Layout<T> = {
+  item: T;
+  index: number;
```
SBoudrias, line 4: We only use `index` in the test. Is it important for the testing?

If we don't have a clear use for that extra argument, let's remove it to keep the exposed interface lean.

artfuldev: Consumers of renderItem might want to render the index of the item as well - it becomes impossible to do that without this argument.

artfuldev:
Thank you so much for the changes! Anything else we need to do?

packages/core/src/lib/pagination/lines.mts
artfuldev: I'm not sure I understand this. Do you just want me to write this as `Prettify<Layout<T>>` instead of `Layout<T>` here?

packages/core/README.md
artfuldev: I've made some changes to the README, please take a look.

packages/core/src/lib/pagination/layout.type.mts
artfuldev: We could also just take a count/size and just use (index, isActive) in renderItem - then it becomes the consuming component's responsibility to lookup the index in its own list of items - then we no longer work on items of type `T`. What do you think?

artfuldev: That also means renderItem can only be defined with a closure to items, and not without it.

packages/demo/demos/select.mjs
```diff
@@ -28,33 +57,18 @@ const demo = async () => {
...
-      { value: 'Z' },
+      ...alphabet,
+    ],
+  });
+  console.log('Answer:', answer);
+
+  answer = await select({
+    message: 'Select your favorite letter (example without loop)',
+    choices: [
+      new Separator('== Alphabet (choices cycle as you scroll through) =='),
```
SBoudrias, line 68: There's a bug here when the first item is not selectable.

It'll allow the loop in that scenario.

packages/core/src/lib/pagination/layout.type.mts
SBoudrias: I'm okay with how it is now.

SBoudrias:
Did a pass today, I was going to merge but ran into that bug. I don't have time to debug it now; but let's add a test, fix the bug and I believe we'll be good to go!

artfuldev:
Hey @SBoudrias !

Thank you for the updates. I've added a few tests to confirm this is fixed. Instead of checking for the first and last item in the list now we check for the indices of the first and last selectable item in the list of choices (they could also both be the same).

I've used useEffect for this which may not be ideal - please let me know what you think. I'm not sure if it's a good idea to recompute this every time - do we assume the items passed will never change?

SBoudrias:
> do we assume the items passed will never change?

Yes, the parameters the prompt received will not change while the prompt runs.

Thanks for sending the update, I'll review the changes more in-depth shortly.

packages/checkbox/src/index.mts
```diff
@@ -82,7 +82,7 @@ export default createPrompt(
 
     const bounds = useMemo(() => {
       const first = items.findIndex(isSelectable);
-      const last = items.findLastIndex(isSelectable);
+      const last = items.length - 1 - [...items].reverse().findIndex(isSelectable);
```
SBoudrias, line 85: Ugly... But works until Node 16 is EOL (well it is, but just since a few days ago) and TS catch up.

artfuldev:
BTW could we participate in hacktoberfest? We could add topics or labels

SBoudrias:
> BTW could we participate in hacktoberfest? We could add topics or labels

I don't really have time to manage this, and I'm not sure which tasks I'd put up there... There's some tricky bug reports in the bug tabs; but that'd need a cleanup first anyway 😓

SBoudrias:
New versions with the feature are out! Thanks for being responsive throughout!

This was a huge feature, and it pointed out some missing feature on the core which is awesome to now have built-in. Cheers!

PR #1290, opened 2023-08-29
chore: add rule to avoid importing the whole of lodash

lukecotter:
Adds an eslint rule to avoid importing the whole of lodash

Relates to [comments](https://github.com/SBoudrias/Inquirer.js/pull/1271#issuecomment-1656245120) in #1271


SBoudrias (approved):
Thanks for the follow up! Truly appreciate it

PR #1292, opened 2023-09-04
Move hooks into separate files

artfuldev:
In this PR, we refactor out hooks into separate files, keeping local types where possible, and sharing types and constants as required from separate files.

This will help future PRs like #1289 

There are no functional changes and this is a pure refactor.


SBoudrias (approved):
Went in and did some changes. I'm not a big fan of that concept of having "1 thing in each file" as I've found it to make it harder to read through the code and understand the context of what things are.

When you use something exported, the only thing you get access to is a name. That name must be hella descriptive for you to understand what it is. But reading the code, you know what it is. As such, it's often better to colocate things that relies on each other if they're not major well known internal APIs to the codebase.

SBoudrias:
@artfuldev no worries! Appreciate your responsiveness and openness.

I really don't mind making some edits myself; it's often faster this way and it helps me better shape the project anyway!

Issue #1297, opened 2023-09-10
Error in `input` for `@inquirer/core` version 4.1.0


patrick-kw-chiu:
<img width="1062" alt="Screenshot 2023-09-10 at 12 49 47 AM" src="https://github.com/SBoudrias/Inquirer.js/assets/42149082/37fd3971-769c-4667-9ee2-7a82828c20ff">

It seems the new version of `use-prefix.mjs` tries to `import * as spinners from 'cli-spinners';`
While the 4.1.0 version is `import spinners from 'cli-spinners';`, which makes the difference

SBoudrias:
Hi, can you provide more details on how you're running the code?

Are you using Typescript? If so, could you paste your `tsconfig`?
I see Node 18.12.1 from the log.

I have some integration tests to validate the imports validity; and they're passing. I'd like to be able to at least replicate this issue and prevent regression.

patrick-kw-chiu:
Thanks for looking into it 🙌  I am using plain JS this time

SBoudrias:
A new patch release is out.

PR #1298, opened 2023-09-10
Update use-prefix.mts

patrick-kw-chiu:
Fix for https://github.com/SBoudrias/Inquirer.js/issues/1297


SBoudrias:
Hi, this change breaks the commonJS build.

PR #1299, opened 2023-09-10
Refactor checkbox

artfuldev:
This is tied to refactors related to #1289 

In this PR, we disambiguate between a choice and an item in the list, and between the cursor position and the active index, and add utility methods. We also extract out the render function. In addition, we simplify `isSelectableChoice` into `selectable`, a refinement function which does not work on `undefined`, and instead handle undefined in the more idiomatic way (which doesn't need comments).

### DIFF
--- a/packages/checkbox/src/index.mts
+++ b/packages/checkbox/src/index.mts
@@ -25,20 +25,34 @@ export type Choice<Value> = {
   type?: never;
 };
 
+type Config<Value> = PromptConfig<{
+  prefix?: string;
+  pageSize?: number;
+  instructions?: string | boolean;
+  choices: ReadonlyArray<Choice<Value> | Separator>;
+}>;
+
 type Item<Value> = Separator | Choice<Value>;
 
-const selectable = <Value,>(item: Item<Value>): item is Choice<Value> =>
-  !Separator.isSeparator(item) && !item.disabled;
+function isSelectable<Value>(item: Item<Value>): item is Choice<Value> {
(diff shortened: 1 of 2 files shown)


packages/checkbox/src/index.mts
```diff
@@ -25,130 +25,107 @@ export type Choice<Value> = {
...
     const [status, setStatus] = useState('pending');
-    const [choices, setChoices] = useState<Array<Separator | Choice<Value>>>(() =>
-      config.choices.map((choice) => ({ ...choice })),
+    const [items, setItems] = useState<ReadonlyArray<Item<Value>>>(
+      choices.map((choice) => ({ ...choice })),
     );
-    const [cursorPosition, setCursorPosition] = useState(0);
+    const [active, setActive] = useState<number>(() => {
+      const selected = items.findIndex(selectable);
+      if (selected < 0) throw new Error(`[checkbox prompt] Nothing selectable`);
```
SBoudrias, line 75: Let's add a test for this.

SBoudrias (approved):
I'm good with the code changes here.

I made a few stylistic edit (e.g I prefer top level utils to use `function` to clearly differentiate what is what.) But nothing logic wise to modify. Let me know once you added a test and we can get that merged!

As always, thanks 😄

artfuldev:
I've incorporated the changes - please take another look, thanks!

SBoudrias:
Perfect, thanks!

PR #1300, opened 2023-09-10
Refactor select

artfuldev:
This is tied to refactors related to #1289

In this PR, we disambiguate between a choice and an item in the list, and between the cursor position and the active index, and add utility methods. We also extract out the render function. In addition, we simplify isSelectableChoice into selectable, a refinement function which does not work on undefined, and instead handle undefined in the more idiomatic way (which doesn't need comments).

### DIFF
--- a/packages/select/src/index.mts
+++ b/packages/select/src/index.mts
@@ -81,7 +81,7 @@ export default createPrompt(
         const offset = isUpKey(key) ? -1 : 1;
         let next = active;
         do {
-          next = (((next + offset) % items.length) + items.length) % items.length;
+          next = (next + offset + items.length) % items.length;
         } while (!selectable(items[next]!));
         setActive(next);
       } else if (isNumberKey(key)) {
--- a/packages/select/src/index.mts
+++ b/packages/select/src/index.mts
@@ -27,7 +27,7 @@ type Choice<Value> = {
 
 type Item<Value> = Separator | Choice<Value>;
 
-function isSelectable<Value>(item: Item<Value>): item is Choice<Value> {
+function isSelectableChoice<Value>(item: Item<Value>): item is Choice<Value> {
   return !Separator.isSeparator(item) && !item.disabled;
 }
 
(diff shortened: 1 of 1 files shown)


packages/select/src/index.mts
```diff
@@ -25,99 +25,86 @@ type Choice<Value> = {
...
-        while (!isSelectableChoice(selectedOption)) {
-          newCursorPosition =
-            (newCursorPosition + offset + choices.length) % choices.length;
-          selectedOption = choices[newCursorPosition];
-        }
-
-        setCursorPos(newCursorPosition);
+        let next = active;
+        do {
+          next = (((next + offset) % items.length) + items.length) % items.length;
```
SBoudrias, line 84: Why did you modify the calculation? I don't think it needs the 2 modulos. Am I missing a edge case?

artfuldev: I suspect I overcorrected out of habit. Typically when we write such code, the offset could be larger than items.length, leading to a negative integer if the calculation were done differently:

Example:
Let's consider an example where items.length is 5 and next is 2:

    For an offset of -3:
        First expression: (2 - 3 + 5) % 5 = 4
        Second expression: ((2 - 3) % 5 + 5) % 5 = 4

    For an offset of -7:
        First expression: (2 - 7 + 5) % 5 = -5 % 5 = -5 (incorrect)
        Second expression: ((2 - 7) % 5 + 5) % 5 = 0
        
The second expression is how I'm used to normalizing arithmetic over an index which could result in negatives. But given the offset is in the interval [-1, 1], the simpler expression as before works. I'll revert this change.

artfuldev: Also you see I was inlining the `index()` helper call in #1289 which was defined this way, so that's the side effect.

artfuldev:
I've incorporated the changes and made it consistent with #1299 - please take another look, thanks!

SBoudrias (approved):
Great thanks!

packages/select/src/index.mts
```diff
@@ -25,9 +25,14 @@ type Choice<Value> = {
...
 
+type SelectConfig<Value> = PromptConfig<{
+  choices: ReadonlyArray<Choice<Value> | Separator>;
+  pageSize?: number;
+}>;
+
 type Item<Value> = Separator | Choice<Value>;
 
-function isSelectableChoice<Value>(item: Item<Value>): item is Choice<Value> {
+function isSelectable<Value>(item: Item<Value>): item is Choice<Value> {
```
artfuldev, line 35: I thought we used `isSelectableChoice` in #1299 so I renamed it specifically 😅

SBoudrias: Haha I figured, didn't want to bother and get this in today 😂

Issue #1306, opened 2023-09-19
Immediately setting state in useEffect hook to object crashes node


mokkabonna:
Affected in v4.1.0+ of @inquirer/core 4.0.0 is verified ok. Have also tried latest on v5

It seems when setting some state to array or object (maybe all non primitive values) immediately in a useEffect hook it breaks with the  following cryptic error:

````
$ /Users/me/code/inquirer-prompts/node_modules/.bin/ts-node -T demo/example-bug.ts
1,2Error: async hook stack has become corrupted (actual: 1090, expected: 1045)
 1: 0x10255602c node::AsyncHooks::FailWithCorruptedAsyncStack(double) [/Users/me/.volta/tools/image/node/18.18.0/bin/node]
 2: 0x102555f98 node::AsyncHooks::pop_async_context(double) [/Users/me/.volta/tools/image/node/18.18.0/bin/node]
 3: 0x10252465c node::AsyncWrap::PopAsyncContext(v8::FunctionCallbackInfo<v8::Value> const&) [/Users/me/.volta/tools/image/node/18.18.0/bin/node]
 4: 0x10279b680 v8::internal::FunctionCallbackArguments::Call(v8::internal::CallHandlerInfo) [/Users/me/.volta/tools/image/node/18.18.0/bin/node]
 5: 0x10279b170 v8::internal::MaybeHandle<v8::internal::Object> v8::internal::(anonymous namespace)::HandleApiCallHelper<false>(v8::internal::Isolate*, v8::internal::Handle<v8::internal::HeapObject>, v8::internal::Handle<v8::internal::HeapObject>, v8::internal::Handle<v8::internal::FunctionTemplateInfo>, v8::internal::Handle<v8::internal::Object>, v8::internal::BuiltinArguments) [/Users/me/.volta/tools/image/node/18.18.0/bin/node]
 6: 0x10279a9c8 v8::internal::Builtin_HandleApiCall(int, unsigned long*, v8::internal::Isolate*) [/Users/me/.volta/tools/image/node/18.18.0/bin/node]
 7: 0x102f9d18c Builtins_CEntry_Return1_DontSaveFPRegs_ArgvOnStack_BuiltinExit [/Users/me/.volta/tools/image/node/18.18.0/bin/node]
 8: 0x102f28198 Builtins_InterpreterEntryTrampoline [/Users/me/.volta/tools/image/node/18.18.0/bin/node]
[...]

PR #1307, opened 2023-09-20
Add unit test for bug with setting object/array in state in useEffect hook

mokkabonna:
Simple test case to show fail for bug described in #1306 

Not sure how to fix it though. 

Test output:

```
 FAIL  packages/core/core.test.mts > createPrompt() > useEffect: works with setting state at once with objects
RangeError: Maximum call stack size exceeded
 ❯ new AsyncResource node:async_hooks:160:14
 ❯ Function.bind node:async_hooks:261:13
 ❯ Module.withUpdates packages/core/src/lib/hook-engine.mts:76:24
     74|   };
     75|
     76|   return AsyncResource.bind(wrapped);
       |                        ^
     77| }
     78|
 ❯ packages/core/src/lib/use-keypress.mts:14:21
 ❯ packages/core/src/lib/hook-engine.mts:111:23
 ❯ packages/core/src/lib/hook-engine.mts:122:9
 ❯ packages/core/src/lib/hook-engine.mts:121:25
 ❯ wrapped packages/core/src/lib/hook-engine.mts:66:25
 ❯ AsyncResource.runInAsyncScope node:async_hooks:203:9
```


packages/core/core.test.mts
```diff
@@ -83,6 +83,31 @@ describe('createPrompt()', () => {
...
+      });
+
+      return String(value);
+    };
+
+    const prompt = createPrompt(Prompt);
+    const { answer, events } = await render(prompt, { message: 'Question' });
+    events.keypress('enter');
+
+    const resolvedAnswer = await answer;
```
mokkabonna, line 107: awaiting it instead of using expect(answer).resolves as this produces a better error message.

SBoudrias: Mind adding an inline comment so we don't loose that context 😄

mokkabonna: done :)

SBoudrias:
Thanks for the test!! It helped a lot 😄 

Got a bit thrown away by that AsyncResource error. But it was a simple infinite loop bug 🤦🏻

PR #1316, opened 2023-10-08
Checkbox - add validation if no choice is selected

### LINKED ISSUE #1257: Add validate function to checkbox
Hi, It would be great to have a validate function on checkbox package. I know items should be valid since user only have to choose between them. But validate is still needed. For instance, what happend is user doesnt pick an option? what if one of two options must me select to continue? 
This behaviour was contemplated previuosly but is not available now in the Checkbox package.

Srenky:
Closes #1257 (first part)

Adds validation logic for checkbox component that notifies the user if no choice has been selected and presses `enter` (enforces that at least one choice must be selected)

image

Added unit test for new logic. Also needed to edit a few existing ones since they were failing because of this new feature.


packages/checkbox/src/index.mts
```diff
@@ -96,9 +96,14 @@ export default createPrompt(
 
     const [active, setActive] = useState(bounds.first);
     const [showHelpTip, setShowHelpTip] = useState(true);
+    const [showValidationMessage, setShowValidationMessage] = useState(false);
 
     useKeypress((key) => {
       if (isEnterKey(key)) {
+        if (items.filter(isChecked).length === 0) {
```
SBoudrias, line 103: We'll want to add a new `config` key (`required?: boolean`) to control this new behaviour. Not everyone will want to force a selection.

Tests that exists today shouldn't need to be modified.

packages/checkbox/src/index.mts
```diff
@@ -96,9 +96,14 @@ export default createPrompt(
...
+          setShowValidationMessage(true);
```
SBoudrias, line 104: Though this work for now, it'll not extend to new behaviour and custom validation logic.

I'd refer you to the [validation implementation inside the `input` prompt](https://github.com/SBoudrias/Inquirer.js/blob/master/packages/input/src/index.mts#L49). Using `setError(string)`, we'll have a more flexible validation feature.

packages/checkbox/src/index.mts
```diff
@@ -96,9 +96,14 @@ export default createPrompt(
...
+          return;
```
SBoudrias, line 105: Nit: Let's instead a `if/else` block. Same idea as before, since we'll eventually add a second block to handle custom validation, it'll fit right in and won't skip future code that might be useful (unlikely, but you could see cleanup code done later in the function.)

SBoudrias (commented):
Thanks for the PR! Did a review pass and left you some comments. Let me know if you need more clairification!

Srenky:
FYI @SBoudrias, I started working on the second part of the issue (adding `min` and `max` options with validation). I'll create a new PR after this one is merged.

packages/checkbox/src/index.mts
```diff
@@ -96,11 +104,16 @@ export default createPrompt(
 
     const [active, setActive] = useState(bounds.first);
     const [showHelpTip, setShowHelpTip] = useState(true);
+    const [errorMsg, setError] = useState<string | undefined>(undefined);
 
     useKeypress((key) => {
       if (isEnterKey(key)) {
-        setStatus('done');
-        done(items.filter(isChecked).map((choice) => choice.value));
+        if (required && items.filter(isChecked).length === 0) {
```
SBoudrias, line 111:
```suggestion
        if (required && !items.some(isChecked)) {
```

SBoudrias (approved):
Great! Thanks for following up on all the comments. I'll get this merge ASAP

PR #1319, opened 2023-10-18
Feat (checkbox): Support `validate` config

Srenky:
Second part of #1257 

Added two new optional properties to config: `minChoices`, `maxChoices`

With these properties, users can specify how many choices can be selected.

The validation logic has been expanded.

image
image

**Possible topic for discussion**
- case when the user wants an exact number of choices selected, one possible option would be to make `minChoices` and `maxChoices` equal and extend validation logic with one more condition and its own message
- is this something to think about?


SBoudrias:
Hey @Srenky, thanks for sending this PR!

I'll have to disagree on the approach taken unfortunately. Adding `required` felt like a very common use case, but limiting the selection amount feels like less so and I would prefer not to introduce a specific validation logic for it.

Instead, I'd prefer adding a generic `validate` config that can be used to implement any custom validation a user might need (prevent selection of conflicting choices, or just like you did, add validation for min/max selections, etc.)

You can see an example of a generic `validate` function in the input & editor prompts.

Srenky:
@SBoudrias Totally agree with you on this one. I didn't even think about using `validate`, but it does make sense to do it this way.

packages/checkbox/checkbox.test.mts
```diff
@@ -134,6 +136,7 @@ describe('checkbox prompt', () => {
     `);
 
     events.keypress('enter');
+    await Promise.resolve();
     expect(getScreen()).toMatchInlineSnapshot('"? Select a number 1"');
 
     await expect(answer).resolves.toEqual([1]);
```
SBoudrias, line 142:
```suggestion
    events.keypress('enter');
    await expect(answer).resolves.toEqual([1]);
    expect(getScreen()).toMatchInlineSnapshot('"? Select a number 1"');
```

Let's flip those so we await the answer, and then check the output.

I'm not against using `await Promise.resolve()` if strictly necessary; but those assertions appear to just be in the wrong order.

packages/checkbox/src/index.mts
```diff
@@ -106,13 +110,16 @@ export default createPrompt(
     const [showHelpTip, setShowHelpTip] = useState(true);
     const [errorMsg, setError] = useState<string | undefined>(undefined);
 
-    useKeypress((key) => {
+    useKeypress(async (key) => {
       if (isEnterKey(key)) {
+        const isValid = await validate(items);
```
SBoudrias, line 115: Did you consider sending in only checked items VS all items?

I'm not sure which one will make for the better API 🤔 I'm wondering what you think?

SBoudrias (commented):
Few comments, but overall this look solid! Thanks!

packages/checkbox/src/index.mts
Srenky: Didn't think about it, but it might make sense to send only checked items. The user wants to check the validity of an answer, which in this case should be only checked items and nothing else. This way, the user doesn't need to think about what is and what isn't checked and will receive just the checked items.

SBoudrias: Yeah, I agree. Let's only send checked items to the validation function then. Thanks!

SBoudrias:
Thanks a ton for the PR! Great addition :)

PR #1327, opened 2023-11-01
fix: allow default on select

sakulstra:
Setting a default for selects is quite handy, especially when the list of options is huge.

One can currently "work around", by sorting the choices to put the "default" first, but imo this is annoying enough to better support it natively.


packages/select/src/index.mts
```diff
@@ -76,7 +77,17 @@ export default createPrompt(
       return { first, last };
     }, [items]);
 
-    const [active, setActive] = useState(bounds.first);
+    const defaultItemIndex = useMemo(() => {
+      if (!_default) return -1;
+      return items.findIndex(
+        (item) =>
+          !Separator.isSeparator(item) && isSelectable(item) && item.value === _default,
```
SBoudrias, line 84:
```suggestion
          isSelectable(item) && item.value === _default,
```

packages/select/src/index.mts
```diff
@@ -76,7 +77,17 @@ export default createPrompt(
       return { first, last };
     }, [items]);
 
-    const [active, setActive] = useState(bounds.first);
+    const defaultItemIndex = useMemo(() => {
+      if (!_default) return -1;
```
SBoudrias, line 81:
```suggestion
      if (!'default' in config) return -1;
```

Thinking we might want to be more liberal with the value we allow to provide 🤔 (like allowing falsy values, null or undefined)

packages/select/README.md
```diff
@@ -51,6 +51,7 @@ const answer = await select({
...
 | choices  | `Array<{ value: string, name?: string, description?: string, disabled?: boolean \| string } \| Separator>` | yes      | List of the available choices. The `value` will be returned as the answer, and used as display if no `name` is defined. Choices who're `disabled` will be displayed, but not selectable. The `description` will be displayed under the prompt when the cursor land over the choice. |
+| default  | `string`                                                                                                   | no       | Selects the default value that is initially. If omitted, the first selectable item is selected.                                                                                                                                                                                     |
```
SBoudrias, line 54:
```suggestion
| default  | `string`                                                                                                   | no       | Defines in front of which item the cursor will initially appear. When omitted, the cursor will appear on the first selectable item.                                                                                                                                                                                     |
```

SBoudrias:
Thanks for sending the PR! That's a great addition 😄 

I merged your code inside #1329

PR #1329, opened 2023-11-05
Feat (select): Allow default config in Select prompt

SBoudrias:
#1327 with my edition to merge it now 😄


packages/select/src/index.mts
```diff
@@ -78,12 +78,11 @@ export default createPrompt(
     }, [items]);
 
     const defaultItemIndex = useMemo(() => {
-      if (!_default) return -1;
+      if (!('default' in config)) return -1;
```
sakulstra, line 81: agree that it's nicer, but perhaps we should just rename all `default` to `defaultValue` as it's a bit annoying to work with the reserved word imo.

Issue #1330, opened 2023-11-14
Feature request idea(input): toggle the transform function when CTRL+SPACE is clicked


matteosacchetto:
This idea comes to mind especially from using the password prompt, where being able to toggle the visibility of the password may be something useful. So, today I was wondering if this was something others could be interested in and may be worth up-streaming to the input prompt.

Basically the idea is that, when the user presses the CTRL+SPACE key combination, the transform function is temporarily bypassed to allow the user to see the non-transformed input. If CTRL+SPACE is pressed again, then the transform function is again activated. Whenever the user presses the ENTER key the transform function will be applied independently if it was previously bypassed or not.

Implementing such functionality is rather easy and could be versatile in different situations from just the use in the password prompt, though could be interesting to hear some feedback on this by other users of this library.

For reference here is an example of the modified input prompt with the suggested feature implemented.

https://stackblitz.com/edit/stackblitz-starters-ed1f1b?file=index.ts

Thank you,
Matteo

SBoudrias:
Hi, I'm not against introducing this change. But I have a few questions UX wise.

1. How do we make this feature discoverable? (small help tip?)
2. Is there common shortcuts used out there to unmask a password input? I'd like us to base ourselves on anything that is out there and potentially known already.

Lastly, we should make sure the feature can be turned off with an option. There might be users who don't want to allow unmasking.

Feel free to open a PR!

matteosacchetto:
Hi, thanks for the reply

So, regarding your UX concerns:
1. I think we could manage to show an help tip, we just need to think of the how for it to be helpful and not distracting
2. I do agree with your view, though I have not yet found a common shortcut used by various applications. I found some references to ALT+F8, other to CTRL+SHIFT+C... I will look into this further

Regarding the last point, I would actually say it would be better if it is disabled by default and you have an option to enable it. This way, the feature can be added without it being a breaking change.

This weekend I will open a PR implementing the basic functionality

manrods:
Can i work on this issue?

SBoudrias:
@manrods there's a PR open that kinda does it, but if I recall properly, we're blocked on finding a good key combination. The PR is very outdated, and it might be easier to figure out now. Feel free to open a new one.

manrods:
Will do!

PR #1334, opened 2023-11-19
Feat (password): add option to toggle password visibility

matteosacchetto:
Implementation of the feature request idea #1330

Steps:
- [x] Add basic functionality to skip the transform function
- [x] Add option to show or hide the password
- [x] Add tests
- [ ] Add documentation
- [ ] Decide the shortcut to use
- [ ] Decide how to manage the UX

### DIFF
--- a/packages/password/src/index.mts
+++ b/packages/password/src/index.mts
@@ -12,13 +12,15 @@ import ansiEscapes from 'ansi-escapes';
 type PasswordConfig = PromptConfig<{
   mask?: boolean | string;
   validate?: (value: string) => boolean | string | Promise<string | boolean>;
+  allowShowPassword?: boolean;
 }>;
 
 export default createPrompt<string, PasswordConfig>((config, done) => {
-  const { validate = () => true } = config;
+  const { validate = () => true, allowShowPassword } = config;
   const [status, setStatus] = useState<string>('pending');
   const [errorMsg, setError] = useState<string | undefined>(undefined);
   const [value, setValue] = useState<string>('');
+  const [isMasked, setIsMasked] = useState<boolean>(true);
 
   const isLoading = status === 'loading';
   const prefix = usePrefix(isLoading);
@@ -44,6 +46,12 @@ export default createPrompt<string, PasswordConfig>((config, done) => {
         setError(isValid || 'You must provide a valid value');
         setStatus('pending');
       }
+    } else if (allowShowPassword && key.ctrl === true && key.name === '`') {
+      // CTRL + Space
+      // I only tried on Linux, but the combination on Linux was reported like that
+      //   key.crtl = true
+      //   key.name = '`'
+      setIsMasked(!isMasked);
     } else {
       setValue(rl.line);
       setError(undefined);
@@ -60,6 +68,10 @@ export default createPrompt<string, PasswordConfig>((config, done) => {
     formattedValue = `${chalk.dim('[input is masked]')}${ansiEscapes.cursorHide}`;
   }
 
+  if (!isMasked && status !== 'done') {
+    formattedValue = value;
+  }
+
   if (status === 'done') {
     formattedValue = chalk.cyan(formattedValue);
   }
@@ -69,5 +81,12 @@ export default createPrompt<string, PasswordConfig>((config, done) => {
     error = chalk.red(`> ${errorMsg}`);
   }
 
-  return [`${prefix} ${message} ${formattedValue}`, error];
+  return [
+    `${prefix} ${message}${
+      allowShowPassword && status !== 'done'
+        ? chalk.dim(' (CTRL+Space to show/hide the password)')
+        : ''
+    } ${formattedValue}`,
+    error,
+  ];
 });
--- a/packages/password/password.test.mts
+++ b/packages/password/password.test.mts
@@ -85,4 +85,80 @@ describe('password prompt', () => {
     events.keypress('enter');
     await expect(answer).resolves.toEqual('12345678');
   });
+
+  it('handle show password: false', async () => {
+    const { answer, events, getScreen } = await render(password, {
+      message: 'Enter your password',
+      mask: true,
+      allowShowPassword: false,
+    });
+
+    expect(getScreen()).toMatchInlineSnapshot('"? Enter your password"');
+
+    events.type('J');
+    expect(getScreen()).toMatchInlineSnapshot('"? Enter your password *"');
+
+    events.type('ohn');
+    expect(getScreen()).toMatchInlineSnapshot('"? Enter your password ****"');
+
+    events.keypress({
+      name: '`',
+      ctrl: true,
+    });
+    expect(getScreen()).toMatchInlineSnapshot('"? Enter your password ****"');
+
+    events.keypress('enter');
+    await expect(answer).resolves.toEqual('John');
+    expect(getScreen()).toMatchInlineSnapshot('"? Enter your password ****"');
+  });
+
+  it('handle show password: true', async () => {
+    const { answer, events, getScreen } = await render(password, {
+      message: 'Enter your password',
+      mask: true,
+      allowShowPassword: true,
+    });
+
+    expect(getScreen()).toMatchInlineSnapshot(
+      '"? Enter your password (CTRL+Space to show/hide the password)"',
+    );
+
+    events.type('J');
+    expect(getScreen()).toMatchInlineSnapshot(
+      '"? Enter your password (CTRL+Space to show/hide the password) *"',
+    );
+
+    events.type('ohn');
+    expect(getScreen()).toMatchInlineSnapshot(
+      '"? Enter your password (CTRL+Space to show/hide the password) ****"',
+    );
+
+    events.keypress({
+      name: '`',
+      ctrl: true,
+    });
+    expect(getScreen()).toMatchInlineSnapshot(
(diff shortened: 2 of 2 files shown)


matteosacchetto:
I started implementing the feature request of #1330
I still have to add tests, and handle the UX

Moreover, we also need to decide how we should call the new options

matteosacchetto:
@SBoudrias is there a way to test keyboard combinations, like CTRL + space?

SBoudrias:
> is there a way to test keyboard combinations, like CTRL + space?

Not at the moment, we should definitively add it to the testing library though!

packages/password/src/index.mts
```diff
@@ -19,6 +23,7 @@ const password: Prompt<string, PasswordConfig> = (config, context) => {
     {
       ...config, // Make sure we do not display the default password
       default: undefined,
+      allowSkipTransformer: config.allowShowPassword,
```
SBoudrias, line 26: I'm not sure this is the right approach.

Approaching the implementation this way, we're limiting ourselves to having the implementation lives inside the `input` prompt. One thing for example that's missing now is an indication of the shortcut to unmask the password.

I feel like this feature addition might call for implementation a password prompt that is not a passthrough to the input one. At the cost of duplicating some code (but I don't think it's so much to be problematic 🤔.)

The alternative would be to work on actual prompt composition within the core (right now it wouldn't work out of the box.) Thinking about this option quickly, it feels like a more complex code change and I'm not sure where else we would want that capacity. So unclear to me if it's worth the effort VS some code duplication.

matteosacchetto: Yeah, I do agree that having the implementation be only in the input prompt is a bit limiting.

So, I do think that the idea of implementing a password prompt might be the right call. While we will have some code duplication, we will have more freedom with managing the two prompts independently, and I do think in the end it looks cleaner (for example, we will not anymore to pass `default: undefined` to avoid having a default value)

Prompt composition is a bit overkill for this task and I do not think it's worth the effort...

I will then revert the changes made and proceed with the implementation of a password prompt

matteosacchetto:
> > is there a way to test keyboard combinations, like CTRL + space?
> 
> Not at the moment, we should definitively add it to the testing library though!

I will look into this further, since I think is not that difficult to add to the library, and it would allow us to write tests for this new feature

matteosacchetto:
> > > is there a way to test keyboard combinations, like CTRL + space?
> > 
> > 
> > Not at the moment, we should definitively add it to the testing library though!
> 
> I will look into this further, since I think is not that difficult to add to the library, and it would allow us to write tests for this new feature

Quick update: to allow for testing keyboard combination is relatively easy to implement. We just need to @inquirer/testing an additional event, which exposes all the parameters a 'keypress' event supports
```ts
export interface Key {
  name?: string | undefined;
  ctrl?: boolean | undefined;
  meta?: boolean | undefined;
  shift?: boolean | undefined;
}
```

We just need to decide whether to implement a new function, modify the current keypress without a breaking change or modify it with a breaking change.
I would say it would be better to implement a new function or to modify the keypress function without introducing a breaking change, since the option of passing keyboard combinations/shortcuts is less used than passing normal key presses

packages/password/src/index.mts
```diff
@@ -1,39 +1,110 @@
...
+    });
 
-        return '';
-      },
-    },
-    context,
-  );
-};
+    const message = chalk.bold(_config.message);
+    let formattedValue = transformer(value, _config.mask, { isFinal: status === 'done' });
```
SBoudrias, line 89: Let's inline the `transformer` function logic - it's a bit longer, but the logic isn't reused and I think it'll be simpler to read through the prompt code.

packages/password/src/index.mts
```diff
@@ -1,39 +1,110 @@
...
+    if (togglePassword && status !== 'done') {
+      formattedValue = value;
+    }
 
-export default password;
+    if (status === 'done') {
+      formattedValue = chalk.cyan(formattedValue);
+    }
```
SBoudrias, line 96: I think there's a small logic mistake here. When the prompt is completed, we shouldn't show the password in plain text.

packages/password/src/index.mts
```diff
@@ -1,39 +1,110 @@
...
-        if (config.mask) {
-          const maskChar = typeof config.mask === 'string' ? config.mask : '*';
-          return maskChar.repeat(str.length);
-        }
+  return createPrompt<string, PasswordConfig>((_config, done) => {
+    const { validate = () => true, allowShowPassword } = _config;
+    const [status, setStatus] = useState<string>('pending');
+    const [errorMsg, setError] = useState<string | undefined>(undefined);
+    const [value, setValue] = useState<string>('');
+    const [togglePassword, setTogglePassword] = useState<boolean>(false);
```
SBoudrias, line 48: I think `isMasked` would be a better name than toggle. "toggle" being an action verb makes it feel like it is a function.

packages/testing/src/index.mts
```diff
@@ -59,8 +59,21 @@ export async function render<TestedPrompt extends Prompt<any, any>>(
...
-      input.emit('keypress', null, { name });
+    keypress(
+      key:
+        | string
+        | {
+            name?: string | undefined;
+            ctrl?: boolean | undefined;
+            meta?: boolean | undefined;
+            shift?: boolean | undefined;
+          },
```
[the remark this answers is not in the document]
matteosacchetto: Done! Once it gets merged I will rebase this branch

packages/password/src/index.mts
```diff
@@ -1,39 +1,110 @@
...
+  if (!isFinal) {
+    return `${chalk.dim('[input is masked]')}${ansiEscapes.cursorHide}`;
+  }
+
+  return '';
 };
 
-const password: Prompt<string, PasswordConfig> = (config, context) => {
+export default (config: PasswordConfig, context?: Context | undefined) => {
   if ('transformer' in config) {
```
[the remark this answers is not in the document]
matteosacchetto: By moving the `if('transformer' in config) ...` inside the `createPrompt` the test which checks that it will throw if the `transformer` property is provided fails

packages/password/src/index.mts
matteosacchetto: I agree. I renamed `togglePassowrd` to `isMasked` and `setTogglePassword` to `setIsMasked`. For consistency, `isMasked` is set to `true` by default, since the password starts masked, and is set to false when password is shown

packages/password/src/index.mts
```diff
@@ -1,39 +1,110 @@
...
+          setStatus('pending');
         }
+      } else if (allowShowPassword && key.ctrl === true && key.name === '`') {
+        // CTRL + Space
+        // I only tried on Linux, but the combination on Linux was reported like that
+        //   key.crtl = true
+        //   key.name = '`'
+        setTogglePassword(!togglePassword);
+        setValue(rl.line);
+        setError(undefined);
```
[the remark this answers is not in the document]
matteosacchetto: We don't need the, since, as you pointed out, are unnecessary

packages/password/src/index.mts
matteosacchetto: When the prompt is completed, status will be equal to `'done'`, so the if at line 90 will be false, and `formattedValue` will remain equal to the transformed value. In the tests, I explicitly check for this scenario, in order to avoid password to be shown in plain text after the prompt is completed.

packages/password/src/index.mts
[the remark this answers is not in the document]
matteosacchetto: Any idea on why it could fail?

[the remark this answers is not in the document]
matteosacchetto: Had to change the test slightly

packages/password/src/index.mts
```diff
@@ -1,39 +1,98 @@
...
+        // get cleared, forcing the user to re-enter the value instead of fixing it.
+        rl.write(value);
+        setError(isValid || 'You must provide a valid value');
+        setStatus('pending');
+      }
+    } else if (allowShowPassword && key.ctrl === true && key.name === '`') {
+      // CTRL + Space
+      // I only tried on Linux, but the combination on Linux was reported like that
+      //   key.crtl = true
+      //   key.name = '`'
```
SBoudrias, line 59: Sooo, did some more digging into this. And it turned out to be a bit of a headache.

Depending on the environment and the terminal used, there can be a lot of conflict.

For one, `CTRL + Space` for me open the AI helper sidebar and never reach the node program.

I then started to think of keys; like `CTRL+H`, but that's a backspace shortcut (and it doesn't even register as the H key.)

So long story short:

1. I added `node tools/keys.mjs` to the repo latest version to play with.
2. Potential solution 1: finding a key combo that'll reasonably work for most environment (terminals, OS & consider non-english keyboards).
3. Have a basic OS switch and use keys per OS that are reasonably going to work.

matteosacchetto:
Separating the rewrite from the new feature was the right choice! 

I'm still experimenting with ideas on how to implement this feature but I have not yet found something which works consistently between different terminals

I think that if we want to ship this feature, in the end we need to find a shortcut which works on most of them and accept that in some environments it won't work as expected

If only there was a way to get which keyboard shortcuts are available in a give environment...

SBoudrias:
Closing, feature was implemented in https://github.com/SBoudrias/Inquirer.js/pull/2237

PR #1340, opened 2023-12-09
fix `when` documentation

mifi:
I found it confusing that `undefined` is the same as `true`, and only `false` will actullay cause it to be skipped, so I think it makes sense to at least document this, maybe a TODO in the future to fix this behaviour (unless it's intended)


SBoudrias:
IMO we could make it throw if a non boolean is returned; but that'd be a breaking change.

The intention is not for `undefined` to be part of our official API contract.

mifi:
If you ask me, undefined should be treated as false. It’s common for APIs to accept truthy/falsy values for booleans (in case developers forgot to explicitly cast to boolan.
E.g it’s common to do something like
```
when: someObject?.someBool
```

…and expect it to be treated as false if someObj is undefined.

But changing that would also be a breaking change.

PR #1344, opened 2024-01-07
Chore: Bump to Yarn 4

SBoudrias:


LitoMore:
Both https://github.com/actions/checkout and https://github.com/actions/setup-node have newer major version. We can bump them to the latest.

SBoudrias:
For future reference in case. There's a problem with old `type-fest` and new typescript. This is a transitive dependency, specifically from ansi-escapes.

Due to latest version of that module not working for common.js - I'm stuck with the `4.x.x` release.

So I force resolution with yarn to a later version: `yarn set resolution type-fest@npm:^0.21.3 npm:^4.0.0`

SBoudrias:
Reporting the hardened mode issue here: https://github.com/yarnpkg/berry/issues/6141

PR #1323, opened 2023-10-22
[major] Introduce theming capacities

SBoudrias:
This PR introduce theming capacity to Inquirer. This idea has been floating around multiple issues in the repo (#822, #1169, and others); but timing wasn't right for major API changes.

# API changes

## Inquirer users
From prompts user perspective, the default prompts (and hopefully the custom ones) will take an optional `theme` object config. That'll define how things are rendered.

Prompts can define their own extension to the theme object for their unique rendering.

## Prompt authors (and internals)
Theme-ability is introduce with the `makeTheme` small utility. A default theme is always applied, but can be extended with each prompts specific theming needs.

# Breaking changes
1. `@inquirer/core`: `usePrefix` interface changed
2. `@inquirer/confirm`: `prefix` config is removed in favour of `theme.prefix`
4. `@inquirer/checkbox`: `prefix` config is removed in favour of `theme.prefix`

# TODOs (to remove draft & ship)
1. [x] Document the feature in `@inquirer/core` for prompt authors
2. [x] Document the feature in each prompt for prompt users
3. [x] Write tests to cover theme-ability support
4. [x] Confirm typing doesn't cause issue & test them

**Want to help?** Comment on the PR with a task you'd like to take on. Open you PR against this PR (base: `theme`)

# Out of scope
Though it is on my radar, I decided to not include an API to handle i18n for now. That'll likely come next with a similar config object but different internal API & types.

PR #1350, opened 2024-01-18
feat: `renderChoices` option

### LINKED ISSUE #1349: feat: `transformChoice` options for checkbox
After the selection within the checkbox prompt is made, the selected values are printed to `stdout`. This can look very messy if the name of the choice is very long. The reason for this is this line:

https://github.com/SBoudrias/Inquirer.js/blob/a3ab2766883fcf2eecd36439b118eb431beffd08/packages/checkbox/src/index.mts#L173

An option like `transformChoice` could solve the problem. The line could then be updated to like this for example:

```ts
.map( transformChoice || (choice => choice.name || choice.value));
```

If I get the OK, I would also work on it right away.

matthyk:
Closes #1349.

Adds the option `renderChoices` to the checkbox prompt. Currently, only the output of the selected choices can be modified by this option. Maybe we should consider to let this function also modify the `prefix` and `message` value?
```ts
      return `${prefix} ${message} ${chalk.cyan(renderChoices(selectedChoices, items))}`;
```


matthyk:
@SBoudrias I have exported the `Config` type to use it in the tests. If we use, we can avoid the usage of `any`

SBoudrias:
@matthyk I was wondering if you saw that PR to add full theming capacity over at https://github.com/SBoudrias/Inquirer.js/pull/1323

Thinking about the changes introduced here, I'm reflecting the better approach might be with a theme. The theming PR is functional and close to ready; I was just hoping to wrap up #1334 before merging it since it'd conflict pretty bad with that PR.

I'd love to hear your thoughts on this!

matthyk:
@SBoudrias Themes are a great feature! This would make it easy to add a `renderChoices` functionality. If the types around `DefaultTheme` would be changed a bit, you could also implement the feature in a way that the `renderChoices` function has access to the current theme and you could even reuse the `renderChoices` function for multiple Themes.

```ts
type CheckboxTheme = {
  checkedIcon: string;
  uncheckedIcon: string;
  cursorIcon: string;
  disabled: (text: string) => string;
};

type Config<Value> = PromptConfig<{
  pageSize?: number;
  instructions?: string | boolean;
  choices: ReadonlyArray<Choice<Value> | Separator>;
  loop?: boolean;
  required?: boolean;
  theme?: Partial<
    Theme<CheckboxTheme> & {
      style: Partial<Theme<CheckboxTheme>['style']>;
    }
  >;
  renderChoices?: (this: Config<Value>['theme'], selected: Choice<Value>[], items: Item<Value>[]) => string
}>;
```

```ts
const config: Config<number> = {
  choices: [],
  message: "undefined",
  renderChoices: function renderChoices(selected: Choice<number>[], items: Item<number>[]): string {
    // `this` is typed as Partial<Theme<CheckboxTheme>
    return this?.disabled?("text") // The current typing of `theme` causes us to have to do this ugly double check
  }
}
```

SBoudrias:
@matthyk I like this idea very much! Would you want to open PRs against the theme branch? (I'm guessing ~2 PRs for the type change, and one for adding renderChoice)

I'll work on wrapping up the password prompt change in ~a week.

SBoudrias:
Slightly thinking back on theme + `renderChoices`. API wise I'd prefer we pass the theme as argument rather than implicitly with `this`.

matthyk:
It possibly offers better usability for the developers.
I'm still happy to implement the changes in two separate PRs, but unfortunately, I don't have time for that for another 2 weeks or so

matthyk:
superseded by #1360

PR #1360, opened 2024-02-17
add `renderSelectedChoices` style option

### LINKED ISSUE #1349: feat: `transformChoice` options for checkbox

matthyk:
Closes #1349 and supersedes #1350

I decided not to pass the current theme to the function. I think the function should really only determine how the selected options should be rendered. Further customisation can already be done with the `answer` style option.

### DIFF
--- a/packages/checkbox/src/index.mts
+++ b/packages/checkbox/src/index.mts
@@ -27,6 +27,10 @@ type CheckboxTheme = {
   };
   style: {
     disabledChoice: (text: string) => string;
+    renderSelectedChoices: <T>(
+      selectedChoices: ReadonlyArray<Choice<T>>,
+      allChoices: ReadonlyArray<Choice<T> | Separator>,
+    ) => string;
   };
 };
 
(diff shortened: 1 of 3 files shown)


packages/checkbox/src/index.mts
```diff
@@ -193,10 +199,9 @@ export default createPrompt(
     });
 
     if (status === 'done') {
-      const selection = items
-        .filter(isChecked)
-        .map((choice) => choice.name || choice.value);
-      return `${prefix} ${message} ${theme.style.answer(selection.join(', '))}`;
+      const selection = items.filter(isChecked);
+
+      return `${prefix} ${message} ${theme.style.answer(theme.style.renderSelectedChoices(selection, items))}`;
```
SBoudrias, line 204:
```suggestion
      const selection = items.filter(isChecked);
      const answer = theme.style.answer(theme.style.renderSelectedChoices(selection, items));

      return `${prefix} ${message} ${answer}`;
```

Nit on style (string being very long.) Might require a prettier pass for CI.

SBoudrias:
Thanks for following-up! LGTM, just left a nit and we'll be good to merge/release.

SBoudrias (approved):

PR #1367, opened 2024-03-04
Feat: Support search in the select prompt

### LINKED ISSUE #1216: Feature request: Native filter API for select
Hi, just trying out the new `@inquirer/prompts` package. Looks much more easy to use. :clap:

I wish there was a native support for filtering values based on typed characters? I found this package was mentioned on previous issues https://github.com/mokkabonna/inquirer-autocomplete-prompt, but wouldn't it make sense to implement this in `@inquirer/select`?

```ts
const answer = await select({
  message: 'Select a package manager',
  choices: [
    {
      name: 'npm',
      value: 'npm',
      description: 'npm is the most popular package manager',
    },
    {
      name: 'yarn',
      value: 'yarn',
      description: 'yarn is an awesome package manager',
    },
    {
      name: 'jspm',
      value: 'jspm',
      disabled: true,
    },
    {
      name: 'pnpm',
      value: 'pnpm',
      disabled: '(pnpm is not available)',
    },
  ],
});
```

If user starts typing out some letters, list of choices would be filtered out or highlighted. 
image

Follow up suggestion. If there was an another selection like `Other` in the list, user could select that option and type out their input.

SBoudrias:
Fix #1216

Search feature for `@inquirer/select`. Basing myself on the `<select>` HTML search ergonomics.

### DIFF
--- a/packages/select/select.test.mts
+++ b/packages/select/select.test.mts
@@ -1,4 +1,4 @@
-import { describe, it, expect } from 'vitest';
+import { describe, it, expect, vi } from 'vitest';
 import { render } from '@inquirer/testing';
 import select, { Separator } from './src/index.mjs';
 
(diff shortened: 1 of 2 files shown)


packages/select/select.test.mts
```diff
@@ -494,4 +502,68 @@ describe('select prompt', () => {
...
+    events.type('UNIT');
+    expect(getScreen()).toMatchInlineSnapshot(`
+      "? Select a number
+        Canada
+        China
+      ❯ United States"
+    `);
+
+    // TODO: Replace with fake timers to make test faster
+    await timeout(700);
```
SBoudrias, line 526: To replace

PR #1369, opened 2024-03-09
Feat: Throw custom errors and expose them from @inquirer/core

### LINKED ISSUE #1368: Don't reject with standard error on exit signal
I'm trying to handle a SIGINT signal (tipically pressing `Ctrl^C`) on a script but this raises a standard error:

https://github.com/SBoudrias/Inquirer.js/blob/c91196e66689cfc05e9237763e6f95f3eb5342a1/packages/core/src/lib/create-prompt.mts#L39

I would like to catch a custom error instead. I know that I can use something like checking the error message, but that sounds like crazy to me, because it could lead to catch other errors or break if another released version changes this message.

Please, could you throw a [custom error](https://stackoverflow.com/questions/783818/how-do-i-create-a-custom-error-in-javascript) that I could be able to properly catch to exit when SIGINT is received?

SBoudrias:
Fix #1368

TODO: I think I'd like to expose as `@inquirer/core/errors` (with export map)

### DIFF
--- a/packages/checkbox/checkbox.test.mts
+++ b/packages/checkbox/checkbox.test.mts
@@ -1,5 +1,6 @@
 import { describe, it, expect } from 'vitest';
 import { render } from '@inquirer/testing';
+import { ValidationError } from '@inquirer/core';
 import checkbox, { Separator } from './src/index.mjs';
 
 const numberedChoices = [
@@ -611,6 +612,7 @@ describe('checkbox prompt', () => {
     await expect(answer).rejects.toThrowErrorMatchingInlineSnapshot(
       `[Error: [checkbox prompt] No selectable choices. All choices are disabled.]`,
     );
+    await expect(answer).rejects.toBeInstanceOf(ValidationError);
   });
 
   it('shows validation message if user did not select any choice', async () => {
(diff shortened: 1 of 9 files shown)


mondeja:
That's great. Thank you! 👍🏼

SBoudrias:
re: I think I'd like to expose as @inquirer/core/errors (with export map)

This doesn't work properly for cjs and typescript.

```
@inquirer/select:tsc: src/index.mts:18:33 - error TS2307: Cannot find module '@inquirer/core/errors' or its corresponding type declarations.
@inquirer/select:tsc:   There are types at '/Users/simon.boudrias/github/Inquirer.js/node_modules/@inquirer/core/dist/esm/types/lib/errors.d.mts', but this result could not be resolved under your current 'moduleResolution' setting. Consider updating to 'node16', 'nodenext', or 'bundler'.
@inquirer/select:tsc:
@inquirer/select:tsc: 18 import { ValidationError } from '@inquirer/core/errors';
```

When adding:

```json
  "exports": {
    "./errors": {
      "import": {
        "types": "./dist/esm/types/lib/errors.d.mts",
        "default": "./dist/esm/lib/errors.mjs"
      },
      "require": {
        "types": "./dist/cjs/types/lib/errors.d.ts",
        "default": "./dist/cjs/lib/errors.js"
      }
    }
```

PR #1373, opened 2024-03-20
Chore: Update timeout example

mdonnalley:
The [Get Default Value After Timeout](https://github.com/SBoudrias/Inquirer.js/?tab=readme-ov-file#get-default-value-after-timeout) recipe didn't work for me since the process would wait until the timeout completed to exit.

I've updated the recipe to abort the timeout once the prompt has been answered.

### DIFF
--- a/packages/demo/demos/timeout.mjs
+++ b/packages/demo/demos/timeout.mjs
@@ -1,0 +1,34 @@
+import * as url from 'node:url';
+import { setTimeout } from 'node:timers/promises';
+import { input } from '@inquirer/prompts';
+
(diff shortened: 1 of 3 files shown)


SBoudrias:
Hi! Thanks a lot for sending this PR. This feedback is point on 😄 

I went ahead and slightly edited the demo to use less open variables and follow the project coding style.

I also added it into a live demo on `@inquirer/demo` as it'll be easier to run and test. Cheers 🎉

SBoudrias (approved):

PR #1377, opened 2024-03-29
added strict equals for null input

coloradude:
Currently any falsy value of a default option will resolve to an empty string. This fixes this behavior.


SBoudrias:
> any falsy value

Well only `null`/`undefined`; not any falsy value unless I'm missing something. I'm not sure why we would want to allow `undefined` but not null.

Maybe it'd be best to check if the key was passed or not; `'default' in this.opt ? ...`. That way we would allow arbitrary default value.

Issue #1386, opened 2024-04-11
Support unicode for Windows


jeffsee55:
Windows supports unicode characters but the figures library that determines this is [outdated](https://github.com/SBoudrias/Inquirer.js/blob/master/packages/select/package.json#L61). I suspect this is intentional because later versions of figures only support ESM.

It would be nice if Windows users got unicode, though I'm not sure what the best path forward would be if dropping CJS support is not an option.

SBoudrias:
I'm not against dropping `figures` and/or inlining core content of the library.

jeffsee55:
Great, I'm happy to take a stab at it. Would you prefer to inline it into it's own package within the monorepo or just put it into `@inquirer/core`?

SBoudrias:
It's own package, like `@inquirer/figures` (basically a simple fork to support cjs), would be best.

If we didn't use figures in multiple packages, then I'd prefer to inline it in those packages. But that'd be a lot of noise today since it's used often.

PR #1388, opened 2024-04-12
Fork figures package so it can be published for esm/cjs

jeffsee55:
Closes https://github.com/SBoudrias/Inquirer.js/issues/1386

This pr ports the `is-unicode-supported` and `figures` packages so they can be dual published in ESM/CJS.


SBoudrias:
Thanks! I pushed an extra update to also cover `inquirer`.

jeffsee55:
Ok great! Wasn't sure about that one. Was about to write some tests to please codecov, they're pretty trivial though so let me know if it's worthwhile doing so.

SBoudrias:
I'm fine ignoring the codecov. But if it's trivial, it does make it easier to manage when people send PRs.

SBoudrias:
Uou can also ignore the linting job; it's because the package is not published, and to make the linting job faster it doesn't depends on building the package.

SBoudrias:
I went ahead and push a release right now since I'm planning to get breaking change in today.

A follow-up PR with test would be great if you're still planning it.

Last point is debugging the eslint failure; since I was wrong 🤦🏻  https://github.com/SBoudrias/Inquirer.js/pull/1389

PR #1390, opened 2024-04-12
Add test for figures package

jeffsee55:
This is a pretty basic test, I didn't want to change the implementation so stubbing the environment was a hands-off way to get a baseline test in there.


SBoudrias:
Thanks!

PR #1397, opened 2024-04-24
Eslint v9 & flat configs

SBoudrias:
Updating to Eslint 9 & the new flat configs.


.github/workflows/main.yml
```diff
@@ -14,6 +14,8 @@ jobs:
           node-version: 20
           cache: 'yarn'
       - run: yarn install --immutable --immutable-cache
+      - name: Typescript
+        run: yarn turbo tsc
```
SBoudrias, line 18: Now required for the import linting to work appropriately.

packages/core/src/lib/use-state.mts
```diff
@@ -1,10 +1,15 @@
...
+// eslint-disable-next-line @typescript-eslint/ban-types
 type NotFunction<T> = T extends Function ? never : T;
 
+export function useState<Value>(
+  defaultValue?: NotFunction<Value> | (() => Value),
+): [Value, (newValue: Value) => void];
 export function useState<Value>(
   defaultValue: NotFunction<Value> | (() => Value),
-): [Value, (newValue: Value) => void] {
+): [Value, (newValue: Value) => void];
```
SBoudrias, line 11: Will extract to a feature branch since an improvement rather than a chore.

packages/inquirer/lib/objects/choices.js
```diff
@@ -47,6 +47,15 @@ export default class Choices {
...
   }
 
+  [Symbol.iterator]() {
+    const data = this.choices;
+    let index = -1;
+
+    return {
+      next: () => ({ value: data[++index], done: !(index in data) }),
+    };
+  }
```
SBoudrias, line 57: Will extract to a feature branch

PR #1402, opened 2024-05-03
chore(deps): Upgrade wrap-ansi to 9.0.0, remove @types/wrap-ansi

donmccurdy:
Updates `wrap-ansi` to the latest version, which includes types, and removes `@types/wrap-ansi`. 

Context: https://github.com/donmccurdy/glTF-Transform/issues/1201

Thank you!

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -61,7 +61,6 @@
     "@inquirer/type": "^1.3.1",
     "@types/mute-stream": "^0.0.4",
     "@types/node": "^20.12.8",
-    "@types/wrap-ansi": "^3.0.0",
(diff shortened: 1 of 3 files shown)


donmccurdy:
Ah, so I should have committed those files in the PR too? Good to know, thanks!

> But as far as I know, wrap-ansi and the other non-updated dependencies are ESM-only...

Correct that `wrap-ansi` v9 is ESM-only, and understood that this would be a blocker. I'll close the PR in that case. 

[Looking forward to Node.js v22](https://www.tomups.com/posts/the-end-of-cjs-esm-hell-node-js/). 😅

PR #1403, opened 2024-05-11
Fix: checkbox icon theming

jeffwcx:
When setting the icon for a checkbox:
```ts
const { answer, events, getScreen } = await render(checkbox, {
  message: 'Select a number',
  choices: numberedChoices,
  theme: {
    icon: {
      checked: '√',
      unchecked: 'x',
    },
  },
});
```
If only partial values are passed in, it will result in the cursor becoming undefined.
```
? Select a number
undefined√ 1
 x 2
 x 3
 x 4
 x 5
 x 6
 x 7
```
This is primarily because `makeTheme` only performs a deep merge on the `style`.

### DIFF
--- a/packages/checkbox/checkbox.test.mts
+++ b/packages/checkbox/checkbox.test.mts
@@ -672,6 +672,59 @@ describe('checkbox prompt', () => {
     await expect(answer).resolves.toEqual([1]);
   });
 
+  describe('theme: icon', () => {
+    it('checked/unchecked', async () => {
+      const { answer, events, getScreen } = await render(checkbox, {
+        message: 'Select a number',
+        choices: numberedChoices,
+        theme: {
+          icon: {
+            checked: '√',
+            unchecked: 'x',
+          },
+        },
+      });
+      events.keypress('space');
+      expect(getScreen()).toMatchInlineSnapshot(`
+        "? Select a number
+        ❯√ 1
+         x 2
+         x 3
+         x 4
+         x 5
+         x 6
+         x 7"
+      `);
+      events.keypress('enter');
+      await answer;
+    });
+
+    it('cursor', async () => {
+      const { answer, events, getScreen } = await render(checkbox, {
+        message: 'Select a number',
+        choices: numberedChoices,
+        theme: {
+          icon: {
+            cursor: '>',
+          },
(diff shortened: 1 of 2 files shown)


jeffwcx:
> Hey 👋🏻 Thanks for sending this in!
> 
> I went with a deeper fix to perform deep merges on the theme objects instead. I think that'll prevent more potential issues like this to occur in the future or in community prompts.
> 
> I reverted the changes to the checkbox file, but the tests you sent are very useful to make sure we won't introduce future regression.
> 
> Feel free to review the code (I believe the types in particular could be improved, right now I'm kinda forcing it), I'll merge later! Cheers

Are you referring to the type issue with `@inquirer/testing`?

image
 
If you're not working on improving this, I can submit a PR.

packages/core/src/lib/make-theme.mts
```diff
@@ -1,10 +1,40 @@
...
+      const prevValue = output[key];
+
+      output[key] =
+        isPlainObject(prevValue) && isPlainObject(value)
+          ? deepMerge(prevValue, value)
+          : value;
+    }
+  }
+
+  return output as T;
```
SBoudrias, line 29: Those `as` uses aren't safe since they force a type. Ideally we could have TS type check without forcing the types.

jeffwcx:
> Ah no, I personally don't get this type error. Could you paste the error message?
> 
> (I'll point out the area in the code)

That’s weird, on my computer (macOS), `@inquirer/testing` `render` seems to have this issue whether it’s used in the inquirer.js project or in my own custom prompts.

image

SBoudrias:
I wonder if either:

1. TS version related? (inquirer is currently using latest)
2. Mismatch version between `@inquirer/testing` and `@inquirer/core` (I assumed this wouldn't be a problem, but never tested it.)

jeffwcx:
> I wonder if either:
> 
> 
> 
> 1. TS version related? (inquirer is currently using latest)
> 
> 2. Mismatch version between `@inquirer/testing` and `@inquirer/core` (I assumed this wouldn't be a problem, but never tested it.)

Personally, I'm inclined towards option 2. I will try to figure it out and then report back to you, please hold on.

PR #1406, opened 2024-05-14
Fix: Open editor with default value

### LINKED ISSUE #1405: Default value for editor prompt not written to file (old API)
### Description

The value of the `default` property is not written to the temporary file when `waitUserInput` is set to false.

### Reproduction

Using `vi` : 
```javascript
import inquirer from "inquirer";

process.env.EDITOR = "vi";

const answer = await inquirer.prompt({
  type: "editor",
  name: "value",
  message: "you shouldn't see this unless your message wasn't validated!",
  default: "replace this with your own message",
  validate(text) {
    if (!text) {
      return "you have to type something";
    }
    return true;
  },
  waitUserInput: false,
});

console.log(answer);

```

#### Expected Behavior
For `vi` to open with the given default contents.

#### Actual Behavior
The opened file is empty (furthermore, the prompt "Press \<enter\> to launch your preferred editor" is overlaid on top of the editing area).

### Cause

The feature introduced in #1150 adds the `waitUserInput` flag. The implementation can be viewed at [this link](https://github.com/SBoudrias/Inquirer.js/commit/8328d659466cce33f0f9171c9daf30927fa7de0d#diff-479a1e309f0fe42c909aa27d39f3e841147f0ea568e4238bf3635c457171b65aR29-R31).  

1. The external editor is opened before the default value can be copied over to `this.currentText`.  
2.  The initial render method is still called even though the external editor is immediately opened.

### Proposed Solution

The conditional should be moved after `this.currentText` is set, and `this.render` should only be called if `waitUserInput` is set to true.

mensah-j:
### Description

Fixes #1405, reverting the bugs introduced in [8328d65](https://github.com/SBoudrias/Inquirer.js/commit/8328d659466cce33f0f9171c9daf30927fa7de0d#diff-479a1e309f0fe42c909aa27d39f3e841147f0ea568e4238bf3635c457171b65aR29-R31) (the editor opens without defaults, and Inquirer renders text on top of in-terminal editors such as `vi`). 

### Changes Made

The editor now opens only after `this.currentText` has been set, and the initial render does not occur if `waitUserInput` is false.

### Testing

The previous tests for the editor prompt do not make use of the `default` property set in the editor fixture.  The message written to the temporary file and its expected contents are turned into test parameters `message` and `expectedAnswer`, respectively. Two additional test cases are added in which the test editor `writeBin` is called without the final argument (`message = ''`), which should leave the contents of the file untouched.

### DIFF
--- a/packages/inquirer/examples/editor.js
+++ b/packages/inquirer/examples/editor.js
@@ -18,6 +18,13 @@ const questions = [
     },
     waitUserInput: true,
   },
+  {
+    type: 'editor',
+    name: 'edition',
+    message: 'Edit the following content.',
+    default: 'Hello, World!',
+    waitUserInput: false,
+  },
 ];
 
 inquirer.prompt(questions).then((answers) => {
--- a/packages/inquirer/lib/prompts/editor.js
+++ b/packages/inquirer/lib/prompts/editor.js
@@ -26,10 +26,6 @@ export default class EditorPrompt extends Base {
     const waitUserInput =
       this.opt.waitUserInput === undefined ? true : this.opt.waitUserInput;
 
-    if (!waitUserInput) {
-      this.startExternalEditor();
-    }
-
     // Trigger Validation when editor closes
     const validation = this.handleSubmitEvents(this.editorResult);
     validation.success.forEach(this.onEnd.bind(this));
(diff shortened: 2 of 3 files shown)


SBoudrias:
Thanks for sending a fix!!

PR #1432, opened 2024-06-24
feat (inquirer): switch to picocolors & more

43081j:
Switches out a few dependencies:

- `through` has been replaced with a regular `Writable` stream
- `chalk` has been replaced with `picocolors`
- `lodash` has been replaced with individual lodash packages (rather than pulling the whole thing)
- `lodash.map` has been replaced with native methods

This improves the leanness of the dependency tree a fair bit.

Let me know if there's some reason you don't want this and i'm happy to close if it makes sense 👍 

**Side note**: i noticed you check in your yarn cache. this makes the repo huge to clone and feels like an odd thing to do. any chance you could let me know what the thinking was there?


packages/inquirer/examples/input.js
```diff
@@ -29,12 +23,11 @@ const questions = [
     name: 'fav_color',
     message: "What's your favorite color",
     transformer(color, answers, flags) {
-      const text = chalk.hex(isHex(color) ? color : 'fff')(color);
```
43081j, line 32: this was a fun example but picocolors doesn't support hex colours, so it doesn't make much sense anymore

i opted to just remove it rather than keep a dep just for an example. i suppose if we're not running these examples locally, we could leave it as-is, knowing it may not run in the repo itself

packages/inquirer/package.json
```diff
@@ -60,15 +60,19 @@
...
   "dependencies": {
     "@inquirer/figures": "^1.0.3",
-    "@ljharb/through": "^2.3.13",
     "ansi-escapes": "^4.3.2",
-    "chalk": "^5.3.0",
     "cli-cursor": "^3.1.0",
     "cli-width": "^4.1.0",
     "external-editor": "^3.1.0",
-    "lodash": "^4.17.21",
+    "lodash.clone": "^4.5.0",
```
kibertoad, line 67: do all those separate lodash dependencies in total still smaller than entire lodash? I'd assume there is at least some overlap between them.

probably some of them, such as set/get could be replaced by something like dot-prop

packages/inquirer/package.json
```diff
@@ -60,15 +60,19 @@
   "license": "MIT",
   "dependencies": {
     "@inquirer/figures": "^1.0.3",
-    "@ljharb/through": "^2.3.13",
```
kibertoad, line 63: removing just this one should already have a significant impact, it does bring quite a few dependencies recursively!

packages/inquirer/package.json
43081j: pkg-size seems to suggest this:

- lodash 1.4MB
- this group of packages 199KB

packages/inquirer/examples/input.js
nicholasgriffintn: I reckon this isn’t a real deal breaker, but an alternative might be https://github.com/webdiscus/ansis, it supports 16 colours rather than 8 and is just a little bigger in size. Not done deep analysis on it though.

43081j: keep in mind this is only for the example, too. we could actually leave the example as using chalk, or whatever package really

inquirer itself doesn't make use of hex colours so should be fine on picocolors

nicholasgriffintn: Yeah just thought I’d shout it out while taking a look :).

Ayc0:
> **Side note**: i noticed you check in your yarn cache. this makes the repo huge to clone and feels like an odd thing to do. any chance you could let me know what the thinking was there?

I think this is for yarn zero-installs feature: https://yarnpkg.com/features/caching#zero-installs

43081j:
i think you're right. i wasn't aware of that before

i do wonder if we could make use github's own caching mechanism and save the repo from having 1GB of non-source in it

it isn't quite the same as 'zero install' i suppose but would still be reading the deps from a cache rather than re-downloading them (via the `setup-node` action, which has a `cache` setting)

packages/inquirer/package.json
SBoudrias: TBH, this one I've got PR wanting to move, then PR to move back to lodash, etc (feel free to search the issues history.) Lodash can be cached and often merged with other dependencies - so we're not comparing the same things when comparing size in isolation.

SBoudrias:
Would you mind splitting in different PRs? There's easy wins in there, but there's changes who're more likely to be controversial.

1. `through` - let's get rid of it 100%
2. picolor - I'm open to it, but let's do it on all packages
3. Lodash - I'll need more convincing (or reducing the usage first; `.map` is a given, let's make that change). But TBH, it's only used in the older `inquirer` package. Folks caring about package size should rely on `@inquirer/prompts` who's much better IMO.

43081j:
#1440 
#1438 
#1437

PR #1434, opened 2024-06-24
Revert "chore: replace `through` with `@ljharb/through` (#1268)"

sj-e2digital:
In my opinion switching to this package added no real benefit, and added a bunch of bloat. The original `through` package has zero dependencies.

Screenshot 2024-06-24 at 12 54 11


kibertoad:
https://github.com/SBoudrias/Inquirer.js/pull/1432 offers an alternative approach, removing the `through` dependency altogether, as it can be done purely with Node.js these days.

nicholasgriffintn:
For context btw, the original issue explains why this was done, the original package wasn't maintained: https://github.com/SBoudrias/Inquirer.js/issues/1266

But if the package doesn't need it at all, and it can be replaced with node internals, go for it IMO.

SBoudrias:
I'll close this one; but note we'll be removing the dependency in favour of a purely native Node.js solution.

PR #1435, opened 2024-06-25
compatible

yuiyideyui:
Compatible with earlier versions of node.js


SBoudrias:
Hi, thanks for the PR. `findLastIndex` is used in multiple places in the codebase - and it is [supported by all LTS versions of Node](https://node.green/#ES2023-features-Array-find-from-last-Array-prototype-findLastIndex). Given this, we'll not add workarounds to support out-of-support versions on Node.js. Highly recommend you to update the Node version of you app/library.

PR #1437, opened 2024-06-25
feat: use writable streams directly

43081j:
Switches to using a writable stream rather than a third party library for the bottom bar logger.


SBoudrias:
Ahh sorry, didn't expect you to react so quick!! I saw you didn't sign your commits, so I jumped the gun and extracted this with `--author` (to make sure you were recognize for that work you did 😄)

I won't do the other one and merge yours. Thanks a ton for the help!

43081j:
no worries, i was browsing around anyway so figured i may as well do the git-fu for you

let me know if you need anything 👍

PR #1438, opened 2024-06-25
feat: switch to picocolors

43081j:
Switches from chalk to picocolors across all packages.

FYI we still pull chalk in via `ora`, which we could replace with `nanospinner` if you like


SBoudrias:
We can replace ora, but I think I would prefer to replace the dependency with `cli-spinners` (or potentially inlining it.) The idea being to make it more consistent with the newer packages dependencies.

PR #1440, opened 2024-06-25
feat: move to individual lodash packages or standard JS

43081j:
Moves to using individual lodash imports to reduce package size.


packages/inquirer/lib/objects/choices.js
```diff
@@ -1,6 +1,5 @@
 import assert from 'node:assert';
-import filter from 'lodash/filter.js';
-import map from 'lodash/map.js';
+import filter from 'lodash.filter';
```
SBoudrias, line 2: Do we need filter from lodash?

packages/inquirer/lib/prompts/base.js
```diff
@@ -1,5 +1,5 @@
-import defaults from 'lodash/defaults.js';
-import clone from 'lodash/clone.js';
+import defaults from 'lodash.defaults';
+import clone from 'lodash.clone';
```
SBoudrias, line 2: I think we could remove those with modern JS

packages/inquirer/lib/objects/choices.js
43081j: the `where` method we have is relying on the object syntax it supports afaict

i.e. `filter(arr, {someKey: 'someValue'})` rather than `arr.filter((v) => v.someKey === 'someValue')`

packages/inquirer/lib/prompts/base.js
43081j: possibly just `{...defaults, ...options}` and `{...cloneMe}`?

SBoudrias: Yeah let's do that, and I don't think you'll need to clone in this context at all.

43081j:
@SBoudrias good work figuring out how to strip a few more of them away 🥳 

really good progress in terms of making it leaner and more maintainable 🙏

SBoudrias:
Thanks for kicking this off!

DJDodds:
https://github.com/advisories/GHSA-p6mc-m468-83gw
Using the lodash.set package has picked up an auditing issue. Is there some workaround for this ?

SBoudrias:
See #1443

PR #1446, opened 2024-06-29
Switch to `yoctocolors`

sindresorhus:
Why?
- Promised long-term maintenance. As you are aware, I maintain things for the long-term. I think this is the most important reason.
- Tree-shakable because Yoctocolors is ESM.
- Better color detection by using the detection built into Node.js.
- As packages using Chalk starts moving to Yoctocolors, it's more likely there will be a Yoctocolor dependency already in the dependency tree.
- Same author as Chalk.


sindresorhus:
I assumed this package was ESM since I saw imports in the code, but noticed now that it publishes to CJS. I can do a `yoctocolors-cjs` package if you're willing to merge this pull request. Let me know.

SBoudrias:
Hey Sindre, I'd be happy to merge. As you know inquirer already relies on many other packages from you. So it does make sense for me to reduce the provenance of all deps.

`inquirer@9` is ESM only, but the new packages under `@inquirer/*` are all dual shipped (too many trouble and complaints with ESM only.) So yeah, a CJS version will be a pre-requisite to adopt here.

Like you cannot wait for all LTS versions to support requiring ESM 🥹

sindresorhus:
@SBoudrias Should be good now.

---

The CJS version lives here: https://github.com/sindresorhus/yoctocolors/tree/cjs

SBoudrias (approved):

SBoudrias:
Thanks!

uncenter:
Sorry for hijacking this pull request but I'm curious what the benefits of yoctocolors are compared to picocolors? I would understand if it is ESM related since picocolors doesn't ship an ESM build yet, but since we are using `yoctocolors-cjs` that isn't it. Other than that I can't see any real differences between the two libraries except for picocolors having larger adoption and yoctocolors being maintainer by you?

sindresorhus:
> Sorry for hijacking this pull request but I'm curious what the benefits of yoctocolors are compared to picocolors?

I have outlined the benefits in the description.

uncenter:
The only reason I'm seeing is that this is maintained by you? Picocolors is still actively maintained and has all the same features and both libraries have the exact same performance according to your own benchmarks.

SBoudrias:
I worked on Yeoman for a long time with Sindre, and relying on many other packages from Sindre still in the dependency tree.

I'd say the trust I have for his packages is high. That's worth a lot for me in context of supply chain attack risks.

So sure that's the main (if not only) benefit I see myself from this switch - and that's enough to warrant it. Nothing against picolors, but in context that's best for Inquirer.

uncenter:
Gotcha, makes sense! Thanks for clarifying.

PR #1447, opened 2024-06-29
[Breaking] Migrate `inquirer` to typescript + dual module bundle

SBoudrias:
Need much more work on fixing core types - might do & ship; but might not.


SBoudrias:
Thinking about this some more, I think what is a more likely outcome is to release `inquirer` without the typescript definitions. `@types/inquirer` are pretty good, and I don't want to bother replicating them right now.

If shipping as-is with definition, the experience would be worse for TS users; simply due to the migration process having been mostly automated.

But, the goal is to dual ship CJS/ESM so we remove a big pain point; not catching up on types (if someone want to help do just that, they're absolutely welcome to.) And this goal, I think we reach if no emitting declarations 🤔 

I'll leave open a few days while I reflect on this. Feel free to comment one way or another!

tools/fix-ext.mjs
```diff
@@ -6,11 +6,15 @@ import { globby } from 'globby';
...
+  // 1. Rename imports & exports
   const fileContent = await fs.readFile(pathname, 'utf8');
   await fs.writeFile(
     pathname,
-    fileContent.replaceAll(/require\(["']([^"']*)\.mjs["']\)/g, "require('$1.js')"),
+    fileContent
+      // from `.mjs` to `.js`
+      .replaceAll(/require\(["']([^"']*)\.mjs["']\)/g, "require('$1.js')")
+      // From export.default to module.exports
+      .replaceAll('exports.default', 'module.exports'),
```
SBoudrias, line 17: Note to self: hacked around this, but not the right approach. The other modules don't do this. I believe it's happening because `inquirer` mixes default and named exports. Might want to clean that instead

prisis:
Small note, you replaced all `/Inquirer.js/` to `/Inquirer.mjs/` urls

SBoudrias:
Moving to #1461 with a v10 release branch.

PR #1448, opened 2024-06-30
Feat(@inquirer/prompts): Add number prompt package. fix #1383

### LINKED ISSUE #1383: Any way to cast the input?
In the old `inquirer` there was a Number input, Im trying to do something similar with the new `@inquirer/prompts` and wonder if there is a way to cast the value before it gets to the `validate` method. Any idea?
I know I can do the cast after calling `await input` but:
1) I think it is nicer to have all the input-related code in one place (the input options)
2) Currently the `validate` method gets a string and I need to "parse" it there as well.

Thx.

ValYouW:


SBoudrias:
Awesome! Thanks for following up and sending a PR!

I went ahead and added a few APIs that felt required to match the [`<input type="number">` HTML API](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number) (min, max, step and required.)

I'd like to validate a decision before merging. I've defaulted `step` to 1. I felt people using the prompt might expect it only allows integers by default. But that's more based on instinct/educated guess than hard data. I'd appreciate a gut check on this! 😄

ValYouW:
Hmm interesting... I guess you have a point providing these built-in validations of min/max/required, on the other hand this lib is a utility that we don't know how is going to be used, I would keep the code as simple as possible and let users provide a `validate` method. After all, if there is min/max why not provide the entire JSON Schema number validations? It's endless...  :)

SBoudrias:
Well, everything is in nuance and balance. Following this argument to the end, then why do we need a number prompt to begin with?

I've found basing myself on the HTML relevant attributes to strike a good balance of ease-of-use while not making the codebase too complex. Like in this case, we're talking about what, less than 20 lines of code change (and I made it a bit more complex by validating `default` which wasn't absolutely necessary.)

Issue #1091, opened 2022-03-14
define question demand (not optional)


sinkcup:
now every questions are optional, there is no way to define it demand.

suggestion: add `demandOption` config, if user select nothing or just press enter, exit error or auto load question to retry.

```
options = {
  things: {
    message: 'What do you want to do?',
    type: 'checkbox',
    choices: ['Order a pizza', 'Make a reservation', 'Bug rate', 'Security'],
    demandOption: true,
  }
};
```

SBoudrias:
Hi, I'm not super clear on your ask.

Do you want to make sure the user select at least one option within the options? If so, I think you can do that safely with a custom `validate` function.

`validate: answers => answers.length > 1 || 'Please select at least one option'`

sinkcup:
@SBoudrias yes, you are right, if this code can evolve to attribute, it will be better, just like `js onblur` vs `html required`:

```
<input type="text" onblur="alter('please input')" />
```

to

```
<input type="text" required />
```

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/required

SBoudrias:
I think this is now handled properly with the `required` option now available on many prompts, and for some prompts the `validate` option that is now exposed.

If anyone need validation/require an answer on a prompt that doesn't provide it, please open a new ticket with the details of your use case! Cheers!

PR #1455, opened 2024-07-01
Add required option to input prompt

### LINKED ISSUE #1157: Allow questions can be marked as mandatory
releated issue: https://github.com/SBoudrias/Inquirer.js/issues/1091

```js
options = {
  things: {
    message: 'App name?',
    type: 'input',
    required: true
  }
};
```

If a user does not provide a non-empty app name, keep asking the same question.

mandryllo:
Resolves #1157 & #1091


SBoudrias:
Thanks! I just slightly modified the code to not always run validate when we know the answer is invalid.

PR #1461, opened 2024-07-07
[Next Major] inquirer@10

SBoudrias:
## Change Log

- Re-implemented with Typescript.
- Re-adding CJS support (now a dual-build CJS/ESM.)
- All core prompts are now coming from `@inquirer/prompt`.
- Custom prompts now should be implemented with `@inquirer/core`. Custom prompts built on `inquirer@9.x.x` will keep working, but should plan a migration.
- `inquirer.ui.BottomBar` is deleted.

## Todo

- [ ] Fix types for non-default prompts
- [x] Update `inquirer` README (mainly/only custom prompt?)
- [x] Get rid of `noImplicitAny`


Tbhesswebber:
Was there a reason for removing the `BottomBar`? Is there some replacement for it?

SBoudrias:
@Tbhesswebber yeah, it didn't fit in the scope or the project and doesn't rely on the same underlying technology. I fitted that in ~15 years ago, and it's been a while I wanted to remove it.

Technically, it had to go with v10 because this version is a major rewrite. And that was removing every underlying code `BottomBar` relied upon.

Feel free to fork v9 and/or extract the BottomBar in a dedicated package. I don't have interest to maintain it going forward.

Tbhesswebber:
@SBoudrias - I appreciate the quick reply - thanks!  I was kind of surprised when I discovered it in the library, but it was nice to have - I'll explore any alternatives (if there are any) and possibly implement my own.

PR #1469, opened 2024-07-11
Fix(typescript): Allow extending inquirer with custom prompts

SBoudrias:


packages/inquirer/package.json
```diff
@@ -61,7 +61,8 @@
     "ansi-escapes": "^4.3.2",
     "mute-stream": "^1.0.0",
     "run-async": "^3.0.0",
-    "rxjs": "^7.8.1"
+    "rxjs": "^7.8.1",
+    "ts-toolbelt": "^9.6.0"
```
SBoudrias, line 65: TODO: Using for dev and built-in utilities; but will inline/simplify just what we need once we get the right types in place.

PR #1476, opened 2024-07-18
Fixed bug where overriding a boolean value with an object path caused a runtime error

### LINKED ISSUE #1473: New set implementation errors when overriding a boolean with an object
# Problem Description
In `9.3.1`, lodash's `_set` method was replaced with a direct implementation. However, this did cause a breaking change.

Consider this example:

```
{
    type: "confirm",
    name: "foo",
    message: "foo?",
    default: true,
  },
  {
    type: "list",
    name: "foo.bar",
    message: "bar?",
    default: 'hello',
    choices: ['hello', 'world'],
    when: (currentAnswers) => currentAnswers.foo,
  },
```

In previous versions, the first question would result in `{foo: true}` and then the second question would override this to be `{foo: { bar: 'hello}}` because it would override the boolean with an object.

In `9.3.1` and later versions, this now throws an error:

```
Cannot create property 'bar' on boolean 'true'
```

# Expected versus Actual

Expected: Boolean answers should be replaced with an object when setting subproperties
Actual: An error is thrown

# Suggested fix

I use this in my local `set` function:

```
if (i !== pathArray.length - 1 && typeof acc[key] === "boolean") {
  acc[key] = {};
}
```

# Alternatives

Because `confirm` does not have a `filter` function where I could override the answer from a boolean to a object myself, this doesn't leave a lot of options other than completely rewriting the entire program to do one prompt at a time instead of passing an array of prompts.

Would prefer to have this small fix to return it back to the original functionality if possible. :)

andrewmcgivery:
Fixes #1473

Tests passing locally:

Screenshot 2024-07-18 at 4 58 13 PM

### DIFF
--- a/packages/inquirer/src/ui/prompt.mts
+++ b/packages/inquirer/src/ui/prompt.mts
@@ -37,7 +37,7 @@ export const _ = {
         pointer[key] = {};
       }
 
-      if (index !== arr.length - 1 && typeof pointer[key] === 'boolean') {
+      if (index !== arr.length - 1 && typeof pointer[key] !== 'object') {
         pointer[key] = {};
       }
 
--- a/packages/inquirer/src/ui/prompt.mts
+++ b/packages/inquirer/src/ui/prompt.mts
@@ -25,7 +25,7 @@ import type {
   StreamOptions,
 } from '../types.mjs';
 
-const _ = {
+export const _ = {
   set: (obj: object, path: string = '', value: unknown): void => {
     let pointer: any = obj;
     path.split('.').forEach((key, index, arr) => {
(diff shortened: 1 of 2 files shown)


packages/inquirer/src/ui/prompt.mts
```diff
@@ -37,6 +37,10 @@ const _ = {
         pointer[key] = {};
       }
 
+      if (index !== arr.length - 1 && typeof pointer[key] === 'boolean') {
```
SBoudrias, line 40: Do we need to make a specific case for boolean? Sounds like it's just the behavior we should get no matter what 🤔 - Even wonder if the issue is the condition on line 36 `!(key in pointer)` - maybe that line should just always run. Or typecheck if not object.

andrewmcgivery: That's actually a good call. I didn't realize at first glance that trying to override a string value would also cause this error.

Updated to do an object check instead.

andrewmcgivery: (side note, I looked at running line 37 every time and it blew up other tests. 😅)

SBoudrias: Uh weird, works for me - I pushed the changes.

SBoudrias (approved):

PR #1485, opened 2024-07-29
fix(@inquirer/number): Allowing decimal numbers

SohailSaghar:
Due to javascript floating point arithmetic the current code does not support decimal numbers as input. However, by multiplying by with 10^6 this now works as intended.


SBoudrias (approved):
Ugh JS 🤦🏻‍♂️  Thanks for sending a fix (and a test 🙌🏻)

SBoudrias:
Released :shipit: https://github.com/SBoudrias/Inquirer.js/releases/tag/%40inquirer%2Fnumber%401.0.7

PR #1486, opened 2024-07-29
Fix(@inquirer/core) Clean hooks immediately on close events.

### LINKED ISSUE #1454: `@inquirer/select`: prompt closes after timeout on `CTRL+C` or `CTRL+D`
When terminating the `@inquirer/select` prompt with `CTRL+C` or `CTRL+D` it properly throws the Error `User force closed the prompt with 0 null`
Though, it does so after the timeout `searchTimeoutRef` is expired.

[Example reproduction](https://stackblitz.com/edit/stackblitz-starters-8duhpc?file=index.mjs)
In this reproduction, try running `node index.mjs` and stop it with `CTRL+C` or `CTRL+D`. To compare the behavior, I also added the `index2.mjs` which uses `@inquirer/input`

By debugging the behavior, I noticed that on `CTRL+C` or `CTRL+D`, the `useKeypress` callback is called, the timeout is cleared and then, since the key combination does not match any of the `if`/`else if` statements, it enters the `else` condition. Here the timeout is created again, and after that timeout expires the Error is thrown.

https://github.com/SBoudrias/Inquirer.js/blob/ac687691f6847816e89a80a589e82738bc4c13b0/packages/select/src/index.mts#L124-L142

To address it, the simplest solution would be to change the `else` statement to an `else if` checking that it does not match `CTRL+C` or `CTRL+D`. Though, other strategies may be used.

Once we agree on a strategy, I am willing to open a PR to address this. 

Tested version:
`@inquirer/select: 2.3.8`

matteosacchetto:
This PR does two main things:
- `@inquirer/core`: it slightly changes the `createPrompt` function, to make sure that when a prompt is force closed, hooks cleanup is handled properly.
- `@inquirer/select`: it adds an effect to clean up the timeout once the prompt has terminated due to either being done, being cancelled or being force closed.

Previously, if a prompt was force closed and the prompt had some pending functions (e.g., `setTimeout`) it ended up in a situation where hooks cleanup was happening only when the `onSignalExit` handler was triggered. But since it was triggered only after pending functions had completed, it created a situation where it was not possible to properly handle the prompt termination (refer to #1454 for a more detailed explanation)

This PR adds a handler on the `close` event of the `readline` class, so that when the prompt is closed, hooks cleanup will be triggered. To make sure it is done after the `keypress` handler, it uses the `setImmediate` function. This allows us to use the `useEffect` cleanup function as a method to clean up any pending resources/functions.

Closes #1454


packages/core/core.test.mts
```diff
@@ -467,6 +467,46 @@ describe('createPrompt()', () => {
...
+
+    const { answer, events } = await render(prompt, { message: 'Question' });
+
+    // This triggers the timeout
+    events.keypress('a');
+    // This closes the readline
+    events.keypress({ ctrl: true, name: 'c' });
+
+    await expect(answer).resolves.toBe('closed');
+    expect(exitSpy).toHaveBeenCalledTimes(1);
```
SBoudrias, line 508: Added this assertion, this is a little logic issue where we should only cleanup one.

I'm fixing now while I review the code.

packages/core/src/lib/use-keypress.mts
```diff
@@ -11,12 +11,15 @@ export function useKeypress(
   signal.current = userHandler;
 
   useEffect((rl) => {
+    let ignore = false;
```
SBoudrias, line 14: Went with this instead of the timeout approach. I think it's fair to ignore any events triggered after the cleanup are called.

SBoudrias:
Let me know what you think! I think I was able to slightly simplify the implementation of what we were trying to do.

packages/core/src/lib/use-keypress.mts
matteosacchetto: This is a fair approach. Do note though that since the `close` event is triggered before the `keypress` event, key combinations like `CTRL+C` won't be propagated to the `useKeypress` hook, so the behavior is slightly different from before.

I do think this behavior is the correct behavior, as those events are automatically handled by readline, so no custom logic on within the `useKeypress` callback should be necessary

matteosacchetto:
I think that the everything looks good! We have now a way to consistently perform cleanup of hooks, and to perform any cleanup of other resources when the readline gets closed

SBoudrias:
Thanks! I'll release later today.

PR #1488, opened 2024-07-30
Add inquirer-file-selector community prompt

br14n-sol:
https://github.com/br14n-sol/inquirer-file-selector


SBoudrias (approved):
That's great! Thanks!

Issue #1521, opened 2024-08-26
Add AbortSignal to @inquirer/core.


mshima:
Allow to programmatically cancel the prompt.

Timeout example.
```js
await input(
  { message: 'Enter your name' },
  { signal: AbortSignal.timeout(5000) },
});
```

In `yeoman-environment` conflicter, conflict check and commit is done in background while the conflict resolution question is prompted. AbortSignal support will make easy to cancel the prompt.

SBoudrias:
Ah, I think that's a great idea.

I'm happy to review a PR, or I'll get to it in a few weeks.

PR #1524, opened 2024-08-28
feat(inquirer): add AbortSignal support

mshima:
Fixes https://github.com/SBoudrias/Inquirer.js/issues/1521

### DIFF
--- a/packages/inquirer/inquirer.test.mts
+++ b/packages/inquirer/inquirer.test.mts
@@ -57,6 +57,19 @@ class StubFailingPrompt {
   close() {}
 }
 
+class StubEventualyFailingPrompt {
+  timeout?: NodeJS.Timeout;
+
+  run() {
+    this.timeout = setTimeout(() => {}, 1000);
+    return Promise.reject(new Error('This test prompt always reject'));
+  }
+
+  close() {
+    clearTimeout(this.timeout);
+  }
+}
+
 beforeEach(() => {
   inquirer.restoreDefaultPrompts();
   inquirer.registerPrompt('stub', StubPrompt);
@@ -760,7 +773,29 @@ describe('inquirer.prompt(...)', () => {
 });
 
 describe('AbortSignal support', () => {
-  it('modern prompts can be aborted through PromptModule constructor', async () => {
+  it('throws on aborted signal', async () => {
+    const localPrompt = inquirer.createPromptModule<TestQuestions>({
+      signal: AbortSignal.abort(),
+    });
+    localPrompt.registerPrompt('stub', StubEventualyFailingPrompt);
+
+    const promise = localPrompt({ type: 'stub', name: 'q1', message: 'message' });
+    await expect(promise).rejects.toThrow(AbortPromptError);
+  });
+
+  it('legacy prompts can be aborted by module signal', async () => {
+    const abortController = new AbortController();
+    const localPrompt = inquirer.createPromptModule<TestQuestions>({
+      signal: abortController.signal,
+    });
+    localPrompt.registerPrompt('stub', StubEventualyFailingPrompt);
+
+    const promise = localPrompt({ type: 'stub', name: 'q1', message: 'message' });
+    abortController.abort();
+    await expect(promise).rejects.toThrow(AbortPromptError);
+  });
+
+  it('modern prompts can be aborted by module signal', async () => {
     const abortController = new AbortController();
     const localPrompt = inquirer.createPromptModule<TestQuestions>({
       signal: abortController.signal,
@@ -772,6 +807,18 @@ describe('AbortSignal support', () => {
 
     const promise = localPrompt({ type: 'stub', name: 'q1', message: 'message' });
     abortController.abort();
+    await expect(promise).rejects.toThrow(AbortPromptError);
+  });
+
+  it('modern prompts can be aborted using ui.close()', async () => {
+    const localPrompt = inquirer.createPromptModule<TestQuestions>();
+    localPrompt.registerPrompt(
+      'stub',
+      createPrompt(() => 'dummy prompt'),
+    );
+
+    const promise = localPrompt({ type: 'stub', name: 'q1', message: 'message' });
+    promise.ui.close();
     await expect(promise).rejects.toThrow(AbortPromptError);
   });
 });
--- a/packages/inquirer/src/ui/prompt.mts
+++ b/packages/inquirer/src/ui/prompt.mts
@@ -14,6 +14,7 @@ import {
 } from 'rxjs';
 import runAsync from 'run-async';
 import MuteStream from 'mute-stream';
+import { AbortPromptError } from '@inquirer/core';
 import type { InquirerReadline } from '@inquirer/type';
 import ansiEscapes from 'ansi-escapes';
 import type { Answers, AnyQuestion, PromptSession, StreamOptions } from '../types.mjs';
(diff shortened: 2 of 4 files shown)


packages/prompts/README.md
```diff
@@ -169,6 +169,7 @@ The context options are:
 | input             | `NodeJS.ReadableStream` | no       | The stdin stream (defaults to `process.stdin`)               |
 | output            | `NodeJS.WritableStream` | no       | The stdout stream (defaults to `process.stdout`)             |
 | clearPromptOnDone | `boolean`               | no       | If true, we'll clear the screen after the prompt is answered |
+| signal            | `AbortSignal`           | no       | An AbortSignal to cancel prompts asynchronously              |
```
SBoudrias, line 172: We'll also need to document this for the new `@inquirer/*` packages (might just be needed in the root README file)

packages/inquirer/inquirer.test.mts
```diff
@@ -760,6 +771,39 @@ describe('inquirer.prompt(...)', () => {
...
+    localPrompt.registerPrompt('stub', StubEventualyFailingPrompt);
+
+    const promise = localPrompt({ type: 'stub', name: 'q1', message: 'message' });
+    abortController.abort();
+    await expect(promise).rejects.toThrow(AbortPromptError);
+  });
+
+  it('modern prompts can be aborted', async () => {
+    const abortController = new AbortController();
+    const localPrompt = inquirer.createPromptModule({ signal: abortController.signal });
```
SBoudrias, line 795: Interface wise, wouldn't it make more sense to pass signal to `localPrompt([...], { signal })`?

It feels a bit weird to me to abort the module and not a specific prompt session.

SBoudrias (changes requested):

packages/inquirer/inquirer.test.mts
mshima: That's the Modern implementation which makes sense in modern api.
In yeoman-environment, when a failure happens we need to cancel any prompt that is been showing.

packages/prompts/README.md
mshima: This file is linked to the root.

packages/inquirer/inquirer.test.mts
SBoudrias: Since `yeoman-environment` wrap inquirer with the `prompt` method, couldn't it inject the signal there instead?

mshima: Inquirer is wrapped by `@yoeman/adapter` and the only reason is to close the prompt:
https://github.com/yeoman/yeoman-api/blob/7c9c24374b1b166bd01b3dbcfdd6ce6976de87b3/workspaces/adapter/src/adapter.ts#L86-L92

Wrapping can be dropped if the signal is added in the module instantiation:
https://github.com/yeoman/yeoman-api/blob/7c9c24374b1b166bd01b3dbcfdd6ce6976de87b3/workspaces/adapter/src/adapter.ts#L44-L48

That's the issue motivation.

An additional signal can be added to the api using `AbortSignal.any()` to merge both signals.

SBoudrias: I don't understand what the issue would be to change where the signal is passed to:

```js
const promptPromise = this.promptModule(questions, initialAnswers, { signal });
```

Can you take some time explain why this won't work in Yeoman? (or the amount of extra complexity this adds if that's the issue?)

SBoudrias:
I took the liberty to:

1. Add more documentation and update the timeout demo. AbortSignal is a much better interface, so I want to make sure we surface it :)
2. Refactored the cleanup logic in `@inquirer/core` - it wasn't great, and I felt it caused some struggle. It'd have been simpler without `promise.cancel()` but oh well, hindsight is 20/20! I'm looking forward to `Promise.withResolver` to make it to all LTS to simplify that further.

I'm very happy with the state of that feature inside `@inquirer/core`. If you want to merge this sooner, we could extract while we sidebar the `inquirer` interface question.

packages/inquirer/inquirer.test.mts
mshima: `@yeoman/adapter` lifecycle is 1 instance per execution.
A single adapter can be shared between many environments. So a prompt will block inputs and queue logs for every environment.

An adapter creates a single inquirer module instance. So:
- the inquirer should be canceled on adapter close at any point. And new prompt will be denied to be shown unless a new inquirer module is created.
- a per execution cancel requires the prompt call to be wrapped, like it is now. There is no reason for this breaking change if the wrapping cannot be dropped.

packages/inquirer/src/ui/prompt.mts
```diff
@@ -303,48 +304,60 @@ export default class PromptsRunner<A extends Answers> {
...
+            if (signal) {
+              const abort = () => {
+                reject(new AbortPromptError({ cause: signal.reason }));
+                cleanup();
+              };
+              if (signal.aborted) {
+                abort();
+                return;
+              }
+              signal.addEventListener('abort', () => abort());
```
SBoudrias, line 347: I think we'll need a `removeEventListener` here too

mshima: Done

packages/inquirer/inquirer.test.mts
SBoudrias: I reflected a bit on this. For `inquirer` as a library, I definitively think it only make sense from the API perspective to abort a "prompt session" and rather than the whole module.

```js
inquirer.prompt([...], { signal });

// or
const prompt = inquirer.createPromptModule();
prompt([...], { signal });
```

Even if this leads to a breaking change in Yeoman, it's better to surface a coherent API in `inquirer`. But I'm pretty sure you could plaster over that and maintain the API backward compatible (like store the signal on the class, and pass it to every prompt call)

mshima:
`AbortSignal.any()` requires node 18.17.0 and 20.3.0.
I don't know a easy way to replace it.

packages/inquirer/inquirer.test.mts
mshima: @SBoudrias I agree about the default provided PromptModule.
Not so much about `createPromptModule` because that's a completely different use case.
As you can see in https://github.com/SBoudrias/Inquirer.js/pull/1535#issuecomment-2325379295.
AbortSignal support has been added to `createPromptModule` with Modern prompts just by adding support to the core.

mshima:
This PR now adds Signal support to the `createPromptModule`, to `inquirer.prompt()` and uses signal to handle `promise.ui.close()`.

SBoudrias:
Did you look into polyfilling AborySignal.any() yet?

mshima:
> Did you look into polyfilling AborySignal.any() yet?

There is AbortSignal polyfills but not `any()`.

packages/inquirer/inquirer.test.mts
SBoudrias: Ah see, this is what I hadn't realized. `inquirer` is prettttty old code, so I'm not so familiar anymore with every details. I didn't remember the options where passed to `createPromptModule` already! - And so truly didn't understand the conflict with this.

I'm torn between:
1. Keeping the module level abort then, since it works out of the box with the built-in prompts. And then let's bring back the module level abort check, since without it the behaviour won't be consistent with plugins. (so having 2 interfaces: `createPromptModule` and `inquirer.prompt`)
2. Drop the `signal` parameters at the prompt module level (ignore it + disallow in type). And only allow within the `prompt` function.

Any preference? Since it pre-exist, I'm not feeling super strongly anymore about either options.

mshima: I really want to keep the signal at module level.

Current PR adds module level and method level.

I suggest to drop at method level for simplicity.
And change `inquirer.prompt` to create a new module each time:
```js
inquirer.prompt = (question, answers, options) => {
  return createModulePrompt(options)(questions, answers);
}
```

SBoudrias: Ah yeah okay, that direction works for me.

SBoudrias (approved):

SBoudrias:
Great! Thanks for all the back and forth on this one 🤞🏻

PR #1525, opened 2024-08-29
feat: export types

mshima:
Export Answers and Questions types. 
`@types/inquirer` have them exported.

### DIFF
--- a/packages/inquirer/src/index.mts
+++ b/packages/inquirer/src/index.mts
@@ -32,7 +32,14 @@ import type {
   StreamOptions,
 } from './types.mjs';
 
-export type { QuestionMap } from './types.mjs';
+export type {
+  Answers,
+  Question,
+  QuestionAnswerMap,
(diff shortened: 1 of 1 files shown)


mshima:
> A problem with this is exporting generic interfaces prevent `inquirer` from later doing any type inferring of what is returned where.
> 
> Also, I'm slightly worried of tying the versioning/breaking changes to the shapes of the types. Especially as they live now since there's a lot of improvements to be implemented yet.

I see, types are not ready to be exported now.

> I'm not sure that's good for the core library. Yeoman could maybe export generic types it owns for backward compatibility reason?

That's the fallback plan, but I probably will create a new breaking change to adopt Inquirer 10 and avoid bad failures, and another breaking change for v10 types.
So I plan to adopt v10 and types at once.

mshima:
Closing since its types needs improvements in other PRs.

Issue #1527, opened 2024-08-30
Inquirer 10/9 compatibility problems.


mshima:
Inquirer 10 is designed to be backward compatible with Inquirer 9.
That's great and allows yeoman-generator/environment stack to adopt Inquirer 10.

Problems:
- [x] types are not behaving correctly: https://github.com/SBoudrias/Inquirer.js/issues/1495, https://github.com/SBoudrias/Inquirer.js/issues/1523
- [x] some types should be exported
- [x] choices doesn't provides a string value option (types only).
- [x] `promise.ui.close()` does not work with modern prompts and there is no alternative
- [x] `promise.ui.rl` is not reliable. Inquirer 9 shared a rl instance for every question in a session. It’s not created for modern questions and a new instance is created for each legacy question. It should be private now.

Those issues needs to be addressed for yeoman-environment to officially adopt it.

SBoudrias:
Do you have a full-repro for the `message` type issue? I don't think it always fail, so I wonder if it's contextual to how the message is passed down somehow...

mshima:
Anoter issue: `ui.close()` does not work with modern prompts and there is no alternative.

SBoudrias:
I didn't look too much into it yet, but I think `filter` might not work at all anymore - not sure when it went (maybe it was initially implemented in prompt classes 😮‍💨). But that'll be a major backward compat issue to fix.

Ref https://github.com/SBoudrias/Inquirer.js/issues/1540

SBoudrias:
Hey Marcelo, anything that's still a blocker? (I'm not sure how accurate the list in the ticket is after the work we put in this 😄)

DelliriumX:
I will ride in on this and add, since TS was not supported in v9, we used `@types/inquirer`. The interfaces that package exported compared to the now TS ready version is very different. Now, i dont mind changing a few things, afterall, it was a major up, but you no longer export types, other than `QuestionMap` no type is exported from inquirer. How are we suppose to type stuff?

I want to make an array of questions which I will later put into an `inquirer.prompt(...my array)` what do I do? How do I type my array when the "Question" type is not exported by the package? We used to have `DistinctQuestion` in `@types/`, i now find something similar with `AnyQuestion` but it is not exported.

mshima:
@DelliriumX if you want to add your question to types, you can implement like this:
[...]

PR #1530, opened 2024-08-30
Fix(inquirer): Allow string choices (backward compat with v9 & prior)

SBoudrias:
Ref #1527

Inquirer v9 & prior allowed choices to be array of strings. Given how widespread that pattern is, we should maintain this backward compatibility. It'll also simplify the work of folks migrating from `inquirer` to `@inquirer/prompts` with less code change.

### TODO

- [x] Select prompt
- [x] Checkbox prompt
- [x] Search prompt
- [x] Rawlist prompt
- [x] Documentation

### DIFF
--- a/packages/checkbox/README.md
+++ b/packages/checkbox/README.md
@@ -101,13 +101,15 @@ type Choice<Value> = {
 
 Here's each property:
 
-- `value`: The value is what will be returned by `await select()`.
+- `value`: The value is what will be returned by `await checkbox()`.
 - `name`: This is the string displayed in the choice list.
 - `short`: Once the prompt is done (press enter), we'll use `short` if defined to render next to the question. By default we'll use `name`.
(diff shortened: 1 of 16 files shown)


mshima:
I saw later that there is backward compatibility in place:
https://github.com/SBoudrias/Inquirer.js/blob/ce5d3b6ab903b43581c34f8069a6b4f108002eb1/packages/inquirer/src/ui/prompt.mts#L297-L306.

It's possible to add to backward compatibility types only:
https://github.com/SBoudrias/Inquirer.js/blob/9b77c6777922cd8fd0a5c1711cf1cef19031c74b/packages/inquirer/src/types.mts#L50-L57

SBoudrias:
Ohh I forgot about that. I'll still bring it over to `@inquirer/*` since it makes migration easier.

PR #1531, opened 2024-08-30
Fix(inquirer): Rework type interface

mshima:


packages/inquirer/src/types.mts
```diff
@@ -1,21 +1,20 @@
...
+  NumberConfig,
+  RawlistConfig,
+  ExpandConfig,
+  PasswordConfig,
+  EditorConfig,
 } from '@inquirer/prompts';
-import type { Prettify, KeyUnion, DistributiveMerge, Pick } from '@inquirer/type';
-import { Observable } from 'rxjs';
+import type { Prettify } from '@inquirer/type';
+import type { Merge, ValueOf } from 'type-fest';
```
SBoudrias, line 15: Let's inline those in `@inquirer/type` and skip an extra external dependency for types.

packages/inquirer/src/types.mts
```diff
@@ -1,88 +1,93 @@
...
+    filter?(input: any, answers: A): any;
+    message: KeyValueOrAsyncGetterFunction<Q, 'message', A>;
+    default?: KeyValueOrAsyncGetterFunction<Q, 'default', A>;
+    choices?: KeyValueOrAsyncGetterFunction<Q, 'choices', A>;
+  }
+>;
 
-export type QuestionArray<A extends Answers> = readonly Question<A>[];
+export type Question<A extends Answers = object> =
+  | LegacyAsyncQuestion<'confirm', ConfirmConfig, A>
```
SBoudrias, line 75: Why export/import the types instead of inferring them?

e.g. more similar to how it was before

```diff
- LegacyAsyncQuestion<'confirm', ConfirmConfig, A>
- LegacyAsyncQuestion<'confirm', Parameters<typeof confirm>[0], A>
```

SBoudrias: I gave it a quick try and it seemed to work just fine.

I strongly prefer the second format as I don't want to tie the library to the internal type implementation. This grows the potential breaking change surface, and makes our public API larger.

SBoudrias:
Alright, I'm pretty happy with the state this is in! I'll merge.

Thanks a ton for the work you put in there! I think this implementation better balance the tradeoffs around types.

PR #1536, opened 2024-09-04
rework prompt lifecycle

mshima:
It's difficult to understand when cleanup hooks are called.
I've added signal, abort and onFinally support to CancelablePromise.
So the lifecycle is managed by CancelablePromise instead of cleanup callbacks.

### DIFF
--- a/packages/type/src/errors.mts
+++ b/packages/type/src/errors.mts
@@ -1,0 +1,4 @@
+export class CancelPromptError extends Error {
+  override name = 'CancelPromptError';
+  override message = 'Prompt was canceled';
+}
--- a/packages/core/src/index.mts
+++ b/packages/core/src/index.mts
@@ -1,5 +1,6 @@
 export * from './lib/key.mjs';
 export * from './lib/errors.mjs';
+export { CancelPromptError } from '@inquirer/type';
 export { usePrefix } from './lib/use-prefix.mjs';
 export { useState } from './lib/use-state.mjs';
 export { useEffect } from './lib/use-effect.mjs';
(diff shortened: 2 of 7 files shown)


SBoudrias:
In absolute, I'd prefer not adding new APIs on top of `Promise`. That's also kind of hard to understand for folks new to the codebase since it's not a Node built-in. I added `withResolver` as-is since it's polyfill for Node 18.

Instead of exposing `onFinally`, could we keep the logic more similar to how it is, but instead of calling `onExit`, we could clear the cleanups on `promise.finally`? (There's a risk approaching it this way to modify the order of operations; I'm not sure without trying if it'll cause side-effects)

packages/type/src/errors.mts
```diff
@@ -0,0 +1,4 @@
+export class CancelPromptError extends Error {
```
SBoudrias, line 1: Not against doing this, but let's move all errors to keep them consistent if we go this route.

mshima:
> In absolute, I'd prefer not adding new APIs on top of `Promise`. That's also kind of hard to understand for folks new to the codebase since it's not a Node built-in. I added `withResolver` as-is since it's polyfill for Node 18.

Since a custom CancelablePromise was implemented a custom `withResolver` looks ok to me.

> Instead of exposing `onFinally`, could we keep the logic more similar to how it is, but instead of calling `onExit`, we could clear the cleanups on `promise.finally`? (There's a risk approaching it this way to modify the order of operations; I'm not sure without trying if it'll cause side-effects)

I’ve tried this approach but a `finally()` will return a bare Promise instead of a CancelablePromise.

SBoudrias:
The difference with the Promise is that it's a polyfill https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers

SBoudrias:
Made 2 demos of alternative to that problem #1542 and #1543 - they're not ideal, so I'd love your review and your take if you have any other ideas to improve this!

PR #1542, opened 2024-09-08
Refactor(@inquirer/core): Streamline exit handlers

SBoudrias:
Alternative answer to #1536.

2 challenges this code brings:

1. Calling done from a hook cleanup can easily lead to an infinite loop (this code handles it, and there's a unit test covering this case already.)
2. The cycle function becomes a bit more complex - might need to revisit this piece. Looked into extracting it out in #1543

### DIFF
--- a/packages/core/src/lib/create-prompt.mts
+++ b/packages/core/src/lib/create-prompt.mts
@@ -38,23 +38,11 @@ export function createPrompt<Value, Config>(view: ViewFunction<Value, Config>) {
     } = PromisePolyfill.withResolver<Value>();
     const promise = Object.assign(rootPromise, {
       /** @deprecated pass an AbortSignal in the context options instead. See {@link https://github.com/SBoudrias/Inquirer.js#canceling-prompt} */
-      cancel: () => fail(new CancelPromptError()),
+      cancel: () => reject(new CancelPromptError()),
     });
 
-    function onExit() {
-      cleanups.forEach((cleanup) => cleanup());
-
-      screen.done({ clearContent: Boolean(context?.clearPromptOnDone) });
-      output.end();
-    }
-
-    function fail(error: unknown) {
-      onExit();
-      reject(error);
-    }
-
     if (signal) {
-      const abort = () => fail(new AbortPromptError({ cause: signal.reason }));
+      const abort = () => reject(new AbortPromptError({ cause: signal.reason }));
       if (signal.aborted) {
         abort();
         return promise;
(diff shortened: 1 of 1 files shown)


packages/type/src/inquirer.mts
```diff
@@ -32,4 +32,4 @@ export type Context = {
 export type Prompt<Value, Config> = (
   config: Config,
   context?: Context,
-) => CancelablePromise<Value>;
+) => Promise<Value> & { cancel: () => void };
```
SBoudrias, line 35: I'll completely remove CancelablePromise in this case; getting it closer to Node 20 deprecation.

packages/core/src/lib/create-prompt.mts
```diff
@@ -86,17 +81,23 @@ export function createPrompt<Value, Config>(view: ViewFunction<Value, Config>) {
...
+              hooksCleanup();
+              resolve(value);
+            });
+          } else {
+            afterCycle = () => {
+              hooksCleanup();
+              resolve(value);
+            };
+          }
+        }
```
SBoudrias, line 99: Not super satisfied with this block of logic, might revisit.

SBoudrias: The tl;dr: if we don't cleanup the hooks as part of the cycle, we ignore/silence the errors thrown while cleaning up and end up with a resolved promise.

If we didn't have to surface hook errors, then the whole logic goes away and we simply end up with:

```diff
- const nextView = view(config, done);
+ const nextView = view(config, resolve);
```

But I think surfacing those errors is important 🤔

SBoudrias: Latest mixes some promises together to get to the right result and order of operation. I'll reflect more on it, but it might be an acceptable solution.

packages/core/src/lib/create-prompt.mts
```diff
@@ -77,48 +61,54 @@ export function createPrompt<Value, Config>(view: ViewFunction<Value, Config>) {
...
+          )
+          // Wait for the promise to settle, then cleanup.
+          .finally(() => {
+            cleanups.forEach((cleanup) => cleanup());
+
+            screen.done({ clearContent: Boolean(context?.clearPromptOnDone) });
+            output.end();
+          })
+          // Once cleanup is done, let the expose promise resolve/reject to the internal one.
+          .then(() => promise),
```
mshima, line 108: Why is this needed?
```suggestion
```

mshima:
Looks 👍 .

PR #1543, opened 2024-09-08
Refactor(@inquirer/core): Streamline exit handlers

SBoudrias:
Alternative answer to #1536.

2 challenges this code brings:

1. Calling done from a hook cleanup can easily lead to an infinite loop (this code handles it, and there's a unit test covering this case already.)
2. The cycle function becomes a bit more complex - might need to revisit this piece.


packages/type/src/inquirer.mts
```diff
@@ -32,4 +32,4 @@ export type Context = {
 export type Prompt<Value, Config> = (
   config: Config,
   context?: Context,
-) => CancelablePromise<Value>;
+) => Promise<Value> & { cancel: () => void };
```
SBoudrias, line 35: I'll completely remove CancelablePromise in this case; getting it closer to Node 20 deprecation.

SBoudrias:
e2e solution in #1542 ended up being more elegant.

PR #1544, opened 2024-09-09
Feat(@inquirer/core): Pass status to `usePrefix`/`theme.message` + Change default prefix

### LINKED ISSUE #1537: I want to change prefix after answered.
Hi, I want to change prefix from ? to ✔ after answered.
like `create-next-app`

스크린샷 2024-09-05 오후 7 31 09

Is there any solution?

nvlang:
This PR includes 5 commits:

- 82778c12a30755e87c1f1c6c42014199d313dc92 — Add `.DS_Store` to `.gitignore` for better DX on macOS
- 6eb53814aa7ad47f85ac811ee8ff6a47594c36b7 — Pass `status` to `usePrefix` (instead of `isLoading`). Also updates references to `isLoading` in README. This allows for behavior like the one described in #1537 to be implemented by the user.
- eced5d67b7229dbac760a48544c211b5626b11c0 — Pass `status` to `style.message` function as a second argument. This allows for behavior like the one described in https://github.com/SBoudrias/Inquirer.js/issues/1537#issuecomment-2336762834 to be implemented by the user.
- cc87d136685b9877afa50df313e7bd8cbcb881ad — Add TSDocs for `Theme` type (via local `DefaultTheme` type). This should improve IntelliSense.
- bf0e8ea61355dfd7935a98857f09c7bde6718c11 — Modify a unit test to test that the `status` property passed to `usePrefix` works as expected.

What's missing is a unit test for the changes to `style.message`. However, I couldn't immediately find unit tests for `style.message` to use as a template, and I didn't want to write one from scratch, since I'm not familiar with the setup. I tested the new `style.message` functionality locally via `corepack yarn demo`, and there it seemed to work fine.

Closes #1537.


packages/core/src/lib/theme.mts
```diff
@@ -3,18 +3,155 @@ import spinners from 'cli-spinners';
...
+     * @param status - The current status of the prompt.
+     * @returns The styled message.
+     *
+     * @defaultValue
+     * ```ts
+     * // import colors from 'yoctocolors-cjs';
+     * (text, status) => colors.bold(text)
+     * ```
+     */
+    message: (text: string, status: string) => string;
```
SBoudrias, line 85: Let's give `status` specific allowed strings. (and everywhere relevant, there's quite a few so I won't call them all 😅 )

packages/core/src/lib/use-prefix.mts
```diff
@@ -5,18 +5,18 @@ import { makeTheme } from './make-theme.mjs';
 import { type Theme } from './theme.mjs';
 
 export function usePrefix({
-  isLoading = false,
+  status = 'pending',
   theme,
 }: {
-  isLoading?: boolean;
+  status?: string;
```
SBoudrias, line 11: Let's give `status` specific allowed strings.

SBoudrias (approved):
Any thoughts on changing the default prefix?

nvlang:
I think `?` for when the question is pending and `✔` for when it's done could make sense? And maybe making the question mark blue and the checkmark green? I had left it alone since it seemed like it'd be a matter of personal preference, but I'd be glad to try to add whatever default you'd prefer.

SBoudrias:
1. Prefix, yeah blue/green question mark/check mark sounds good to me.

2. I think we can go with 3 status. I prefer `idle` to `pending`; so I'd normalize this way.

nvlang:
@SBoudrias I guess changing the default prefix means that we should also update the screenshots in the READMEs — however, I'm not sure what the setup / workflow there is (theme, font, etc.).

SBoudrias:
Thanks for the work you've put on that PR!

I went ahead a made a few adjustments before merging (just an FYI):

- Used `@inquirer/figures` for the tick; since I don't think there was a fallback on windows.
- Replaced the interface of `prefix` from a function to a record.
- Updated documentation.

SBoudrias:
About the screenshots, they're been taken over time by various people and I don't think we documented any process... Feel free to regenerate screenshot and document a reproducible process in the CONTRIBUTING.md doc!

PR #1556, opened 2024-09-17
fix(inquirer): adjust parameters

mshima:
Related to https://github.com/SBoudrias/Inquirer.js/issues/1527

### DIFF
--- a/packages/inquirer/src/ui/prompt.mts
+++ b/packages/inquirer/src/ui/prompt.mts
@@ -117,7 +117,7 @@ export interface LegacyPromptConstructor {
 
 export type PromptFn<Value = any, Config = any> = (
   config: Config,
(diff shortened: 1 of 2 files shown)


packages/inquirer/src/ui/prompt.mts
```diff
@@ -197,7 +197,6 @@ export default class PromptsRunner<A extends Answers> {
   process: Observable<any> = EMPTY;
   private abortController: AbortController = new AbortController();
   private opt: StreamOptions;
-  rl?: InquirerReadline;
```
mshima, line 200: In v9 this is a singleton. In v10 it's undefined for modern and ephemeral for legacy prompts.

SBoudrias (approved):

PR #1557, opened 2024-09-17
feat(inquirer): implement createPromptSession

mshima:
Related to https://github.com/SBoudrias/Inquirer.js/issues/1527.

### DIFF
--- a/packages/inquirer/src/ui/prompt.mts
+++ b/packages/inquirer/src/ui/prompt.mts
@@ -1,8 +1,8 @@
 /* eslint-disable @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-assignment */
 import readline from 'node:readline';
+import { isProxy } from 'node:util/types';
 import {
   defer,
-  EMPTY,
   from,
   of,
   concatMap,
@@ -10,7 +10,9 @@ import {
   reduce,
   isObservable,
   Observable,
+  Subject,
   lastValueFrom,
+  tap,
 } from 'rxjs';
 import runAsync from 'run-async';
 import MuteStream from 'mute-stream';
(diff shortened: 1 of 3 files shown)


packages/inquirer/src/ui/prompt.mts
```diff
@@ -389,10 +409,7 @@ export default class PromptsRunner<A extends Answers> {
   };
 
   private shouldRun = async (question: AnyQuestion<A>): Promise<boolean> => {
-    if (
-      question.askAnswered !== true &&
-      _.get(this.answers, question.name) !== undefined
-    ) {
+    if (question.askAnswered !== true && this.answers[question.name] !== undefined) {
```
SBoudrias, line 412: This will break where `name` contains a sub-path like `name: 'foo.bar'` (which is why `_.get` was used)

SBoudrias:
I like the concept/interface, I'd really prefer to eventually move out `rxjs` from `inquirer` and into a wrapper. Most people do not use the reactive interface, and `rxjs` is a very large package that end up loaded in all those CLIs for no values.

packages/inquirer/src/ui/prompt.mts
mshima: It's using a proxy.
Tests are in place for sub-paths.

mshima:
The idea is to make PromptsRunner lifecycle more consistent.
A new PromptsRunner is created each time prompt is called, but properties are populated at `run()` method.
This refactoring not sets PromptsRunner properties at the constructor so we don't have a wrong idea that it will be reused.

packages/inquirer/src/ui/prompt.mts
SBoudrias: Oh, I missed that. I think it's fair to do internally, but I'm unclear why we want to expose the capacity to provide a proxy as argument?

mshima: I want to provide a custom store.
It's possible to use a proxy or implement an api.

This is actually required to simplify supporting untouched inquirer api.
So Observables are supported by loading the question answer on demand.
Currently the prefilled answers are built from questions array.

mshima: This is a feature request, I can drop it.

SBoudrias: I'm not against it if there's a use-case. Just want to make sure we don't add unnecessary complexity we then need to support in "perpetuity". Or open doors for people to unnecessarily overcomplexify their implementations.

PR #1559, opened 2024-09-18
feat(inquirer): export types Question, DistinctQuestion and Answers

mshima:
Related to https://github.com/SBoudrias/Inquirer.js/issues/1527.

### DIFF
--- a/packages/inquirer/inquirer.test.mts
+++ b/packages/inquirer/inquirer.test.mts
@@ -106,6 +106,16 @@ describe('exported types', () => {
     expectTypeOf({ type: 'stub', name: 'q1' }).not.toMatchTypeOf<Question>();
   });
 
+  it('Exported types can be used with "as const" and "satifies" to keep prompt type inference', async () => {
+    const question = {
+      type: 'stub',
+      name: 'q1',
+      message: 'message',
+    } as const satisfies PromptSession;
+
+    expectTypeOf(await inquirer.prompt(question)).toEqualTypeOf<{ q1: any }>();
+  });
+
   it('exported Answers type is not any', () => {
     expectTypeOf(false).not.toMatchTypeOf<Answers>();
   });
--- a/packages/inquirer/inquirer.test.mts
+++ b/packages/inquirer/inquirer.test.mts
@@ -12,8 +12,8 @@ import { vi, expect, beforeEach, afterEach, describe, it, expectTypeOf } from 'v
 import { of } from 'rxjs';
 import { AbortPromptError, createPrompt } from '@inquirer/core';
 import type { InquirerReadline } from '@inquirer/type';
-import inquirer, { type QuestionMap } from './src/index.mjs';
-import type { Answers } from './src/types.mjs';
+import inquirer from './src/index.mjs';
+import type { QuestionMap, Answers, Question, DistinctQuestion } from './src/index.mjs';
 import { _ } from './src/ui/prompt.mjs';
 
 declare module './src/index.mjs' {
@@ -58,7 +58,7 @@ class StubFailingPrompt {
   close() {}
 }
 
-class StubEventualyFailingPrompt {
+class StubEventuallyFailingPrompt {
   timeout?: NodeJS.Timeout;
 
   run() {
@@ -81,11 +81,6 @@ beforeEach(() => {
 });
 
 describe('exported types', () => {
-  type Answers = import('./src/index.mjs').Answers;
-  type Question = import('./src/index.mjs').Question;
-  type BuiltInQuestion = import('./src/index.mjs').BuiltInQuestion;
-  type PromptSession = import('./src/index.mjs').PromptSession;
-
   it('Question type is not any', () => {
     expectTypeOf({}).not.toMatchTypeOf<Question>();
   });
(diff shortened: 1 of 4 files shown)


packages/inquirer/src/types.mts
```diff
@@ -47,7 +47,7 @@ export interface QuestionMap {
 type KeyValueOrAsyncGetterFunction<T, k extends string, A extends Answers> =
   T extends Record<string, any> ? T[k] | AsyncGetterFunction<T[k], A> : never;
 
-export type AnyQuestion<A extends Answers, Type extends string = string> = {
+export type AnyQuestion<A extends Answers = Answers, Type extends string = string> = {
```
SBoudrias, line 50: Do we want to rename it to match what `@type/inquirer` used?

packages/inquirer/inquirer.test.mts
```diff
@@ -80,6 +80,35 @@ beforeEach(() => {
...
+
+  it('exported Answers type is not any', () => {
+    expectTypeOf(false).not.toMatchTypeOf<Answers>();
+  });
+
+  it('exported Answers type matches any object', () => {
+    expectTypeOf({}).toMatchTypeOf<Answers>();
+    expectTypeOf({ foo: 'bar' }).toMatchTypeOf<Answers>();
+  });
+});
```
SBoudrias, line 110: In term of test, I was more thinking along the line of ensuring proper behavior when used; e.g

```ts
it('Question[] infers the right return type', async () => {
  const questions: Question[] = [
    { ... },
  ];

  const answers = await inquirer.prompt(questions);

  expectTypeOf(answers).toMatchTypeOf<{ foo: any; bar: any; }>();
});
```

mshima: I am not sure it's possible.
`prompt()` function types are too complex to be reused.
Types are internal to the prompt function.

I think we should expose a generic type.

SBoudrias: I don't think I follow - I'm just talking of testing the behavior, not changing the type interface (unless it'd become necessary.)

mshima: Returned type of Question will alway be `Record<string, any>`.

packages/inquirer/src/types.mts
mshima: AnyQuestion is been exported as Question.
Answers matches too.

Question has a more advanced type called DistinctQuestion which is the current BuiltInQuestion with name property.
I prefer BuiltInQuestion over DistinctQuestion.

packages/inquirer/inquirer.test.mts
SBoudrias: Ah yes, this is what I thought would happen. That's what I was telling @DelliriumX over https://github.com/SBoudrias/Inquirer.js/issues/1527#issuecomment-2358017515 - but they seemed pretty sure of themselves 🤷🏻 Figured I was making missing something.

So yeah then exporting the types (as-is) does break type inference.

mshima: - I am pretty sure answers type inference does not exist in v9 types.
- in yeoman-generator there is no type inference and I don’t plan to add it.

Exported types is just for easy custom apis:
- if any type inference is needed it should use the function type.
- It’s not even possible for inquirer to create the function type, maybe it’s not possible to extract it and keep type inference.

Plain Question/Answers and maybe BuiltInQuestion should be enough to replace `@types/inquirer`.

packages/inquirer/src/types.mts
DelliriumX: DistinctQuestion was used in `@types/inquirer`. I honestly do not care so long as we get the type, but if its all the same, i'd use the same one that was used before for the sake of "easier" transition.

packages/inquirer/inquirer.test.mts
DelliriumX: > Ah yes, this is what I thought would happen. That's what I was telling @DelliriumX over [#1527 (comment)](https://github.com/SBoudrias/Inquirer.js/issues/1527#issuecomment-2358017515) - but they seemed pretty sure of themselves 🤷🏻 Figured I was missing something.
> 
> So yeah then exporting the types (as-is) does break type inference.

Do not use the `Question` use what was equivalent to the `BuiltInQuestion` used inside the prompt.
If I can make a type that behaves properly with 1 line of code, inferring from `prompt` then I see no reason why this cannot be exported from the package.

DelliriumX: > Plain Question/Answers and maybe BuiltInQuestion should be enough to replace `@types/inquirer`.

I think that these should be enough to replace the package as it stands right now (well, at least for my use-case it will) because most `@types/inquirer` uses were through `DistinctQuestion`

SBoudrias (approved):

SBoudrias:
Implemented a few basic changes:

1. Rename types to `Question`/`DistinctQuestion` to match those in `@type/inquirer`.
2. Remove test examples of `PromptSession` - I updated the test with `[] const satisfies Question[]` which achieve the same thing without needing to know about another type. I kept the export as I assumed you'll need it for Yeoman.

`inquirer@11.1.0` includes this.

PR #1569, opened 2024-09-28
Fix(all): Remove dependency on @types/node (to reduce package size)

### LINKED ISSUE #1326: Consider making `@types/*` dev dependencies
It would be great if the `@types/*` were not direct dependencies but dev dependencies as to avoid the pit fall of having the types for node 20, but running under node 18.

SBoudrias:
Fixes #1326 

Not 100% how to test this effectively. Types passes in the repo, but that's kind of a low warranty for dependents.

Also, I think `@types/mute-stream` include `@types/node` - so we'd probably also want to do away with it.


SBoudrias:
Added an e2e build suite to test `@inquirer/*` outside the repo setup. See https://github.com/SBoudrias/Inquirer.js/pull/1570

PR #1572, opened 2024-10-02
Fix/issue-1537

nicholas-tayag:
d0fd699 - Pass status to usePrefix instead of isLoading. Also updates references to isLoading in README. This allows for behavior like the one described in issue #1537. 

a159e18 - Pass status to style.message function as a second argument. This allows for behavior like the one described, "it could be really nice to be able to get an additional argument for the style.message function in theme objects; currently, it receives only the message string, but if it also received, e.g., a second argument status: 'pending' | 'done', users could, e.g., dim questions that have already been answered, or make the currently pending question brighter", to be implemented by the user.


SBoudrias:
Hey, not sure how where this code come from, but those changes are already merged and released.

PR #1573, opened 2024-10-04
Chore(test): Isolate a tsc run

SBoudrias:
Partial extraction from #1569 - and further improvements (ideally don't wanna use npm just for this...)

### DIFF
--- a/packages/demo/package.json
+++ b/packages/demo/package.json
@@ -76,7 +76,7 @@
     "yoctocolors-cjs": "^2.1.2"
   },
   "devDependencies": {
-    "@sindresorhus/tsconfig": "^6.0.0",
+    "@repo/tsconfig": "workspace:*",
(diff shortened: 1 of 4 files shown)


packages/demo/package.json
```diff
@@ -96,5 +96,6 @@
       "./package.json": "./package.json",
       ".": "./src/index.ts"
     }
-  }
+  },
+  "packageManager": "yarn@4.5.0+sha512.837566d24eec14ec0f5f1411adb544e892b3454255e61fdef8fd05f3429480102806bac7446bc9daff3896b01ae4b62d00096c7e989f1596f2af10b927532f39"
```
SBoudrias, line 100: Might want to do this with corepack github action instead.

SBoudrias:
TODO: also extract `@repo/tsconfig`, and remove the direct dep.

PR #1575, opened 2024-10-11
Feat: gracefully error on missing message

mshima:


packages/core/src/lib/create-prompt.ts
```diff
@@ -75,6 +75,10 @@ export function createPrompt<Value, Config>(view: ViewFunction<Value, Config>) {
             setImmediate(() => resolve(value));
           });
 
+          if (nextView === undefined) {
+            throw new Error('Question is missing a message');
+          }
```
SBoudrias, line 80: This isn't necessarily true here right; many reasons why a view function might return nothing (could just fail or have an internal bug that is not due to message.)

Do you have an example error where this happens?

mshima: The use case is to create a dummy non interactive prompt to be used in yeoman-test.
Error is:
```
TypeError: (intermediate value)(intermediate value)(intermediate value) is not iterable
 ❯ node_modules/@inquirer/core/dist/esm/lib/create-prompt.js:57:54
 ❯ Object.store.handleChange node_modules/@inquirer/core/dist/esm/lib/hook-engine.js:23:17
 ❯ cycle node_modules/@inquirer/core/dist/esm/lib/hook-engine.js:25:19
 ❯ node_modules/@inquirer/core/dist/esm/lib/create-prompt.js:52:13
 ❯ node_modules/@inquirer/core/dist/esm/lib/hook-engine.js:27:16
 ❯ withHooks node_modules/@inquirer/core/dist/esm/lib/hook-engine.js:19:24
 ❯ prompt node_modules/@inquirer/core/dist/esm/lib/create-prompt.js:45:16
 ❯ PromptsRunner.fetchAnswer node_modules/inquirer/dist/esm/ui/prompt.js:227:16
```

mshima: Maybe the error should be intercepted elsewhere but it's quite complicated to understand.

mshima: executing `await input({})` shows:
```
? undefined
```

Maybe a default message instead then?

SBoudrias: Sorry I'm not completely following.

1. If you create a dummy prompt, you could still return a string from the render function and that wouldn't error out no?
2. I don't understand how `input` ends up called with no `message` - can you clarify how it gets there?

I'm okay validating and throwing an error inside `@inquirer/core`, but we should review the error message.

mshima: > 1. If you create a dummy prompt, you could still return a string from the render function and that wouldn't error out no?

The dummy prompt is returning the config.message which was undefined.
https://github.com/yeoman/yeoman-api/blob/e43ca2ca7e842c53a06cce3d990d2050aa2d1ccb/workspaces/adapter/src/testing/test-adapter.ts#L73

I had to debug to discover the error while migrating to new inquirer in tests.
`message` is mandatory in runtime, it was missing in some tests.

SBoudrias: Sorry for the slow response on this issue. I went ahead and rephrase the error message to match what happens from the core library point of view (e.g `message` being common, but optional for custom prompts; and also accounting for other behaviors that could lead to breaking.)

The only doubt I got after reviewing that code is if we shouldn't rather just default to an empty string 🤷🏻

PR #1593, opened 2024-11-04
Expose `external-editor` file options on the editor prompt

bilalesi:
Using the editor plugin in macos raise an error due the directory path was not found, this PR will allow more flexibility to change the temporary file for the editor, by leveraging the `external-editor` options

- Updated editor prompt configuration to accept `file` options (`IFileOptions`) instead of only `postfix`, allowing for more flexible file handling within the editor's prompt.
- Refactored test cases in `editor.test.ts` to accommodate the new file option structure.

Error ex: 
`The path '/var/folders/kg/v5lhl9ps5d7_pc7l1_vggwv89zcpf2/T/tmp-19210mJVCVgQam7Er.json' does not exist on this computer.`


packages/editor/src/index.ts
```diff
@@ -16,14 +16,18 @@ import type { PartialDeep, InquirerReadline } from '@inquirer/type';
 type EditorConfig = {
   message: string;
   default?: string;
-  postfix?: string;
```
SBoudrias, line 19: This is going to create a breaking change that risk leaving a fair amount of the community behind. Let's maintain backward compatibility for this option.

packages/editor/src/index.ts
```diff
@@ -1,5 +1,5 @@
 import { AsyncResource } from 'node:async_hooks';
-import { editAsync } from 'external-editor';
+import { editAsync, IFileOptions } from 'external-editor';
```
SBoudrias, line 2: Can you add documentation of this new option in the prompt README?

packages/editor/src/index.ts
```diff
@@ -16,14 +16,18 @@ import type { PartialDeep, InquirerReadline } from '@inquirer/type';
...
   validate?: (value: string) => boolean | string | Promise<string | boolean>;
   theme?: PartialDeep<Theme>;
+  file?: IFileOptions;
 };
 
 export default createPrompt<string, EditorConfig>((config, done) => {
-  const { waitForUseInput = true, postfix = '.txt', validate = () => true } = config;
+  const {
+    waitForUseInput = true,
+    file: { postfix = '.txt', ...fileProps } = {},
```
SBoudrias, line 28: To keep it backward compatible without too much hassle:

```suggestion
    file: { postfix = config.postfix ?? '.txt', ...fileProps } = {},
```

packages/editor/src/index.ts
bilalesi: good point, i will make the change

packages/editor/src/index.ts
bilalesi: I added the docs for both `editor readme file`  and `global docs`

packages/editor/src/index.ts
bilalesi: done 👍

SBoudrias (approved):

SBoudrias:
Thanks!

PR #1601, opened 2024-11-14
Chore: use vitest workspace

SBoudrias:
Ref https://vitest.dev/guide/workspace


SBoudrias:
New workspace API: https://vitest.dev/blog/vitest-3#inline-workspace

PR #1638, opened 2025-01-11
Chore: Update README to Include Warp

ericdachen:
If there's anything that needs to be changed let me know and we can collaborate!


SBoudrias:
Reopen under https://github.com/SBoudrias/Inquirer.js/pull/1639

PR #1642, opened 2025-01-11
Update README

ericdachen:
Adding centering and change the "Warp, the intelligent terminal"


SBoudrias:
Ah thanks, I didn't know that trick to mix HTML and markdown. I updated #1640 with it in all readmes, will try with npm later.

PR #1683, opened 2025-03-07
Support Node's --frozen-intrinsics flag

jdalton:
Skip assignment of
```js
Error.prepareStackTrace = _prepareStackTrace
```
when the value hasn't changed to avoid triggering a mutation error when Node flag `--frozen-intrinsics` is used.
https://nodejs.org/api/cli.html#--frozen-intrinsics

### DIFF
--- a/packages/core/src/lib/create-prompt.ts
+++ b/packages/core/src/lib/create-prompt.ts
@@ -16,20 +16,22 @@ type ViewFunction<Value, Config> = (
 
 function getCallSites() {
   const _prepareStackTrace = Error.prepareStackTrace;
+  let result: NodeJS.CallSite[] = [];
   try {
-    let result: NodeJS.CallSite[] = [];
     Error.prepareStackTrace = (_, callSites) => {
       const callSitesWithoutCurrent = callSites.slice(1);
       result = callSitesWithoutCurrent;
       return callSitesWithoutCurrent;
     };
-
     // eslint-disable-next-line @typescript-eslint/no-unused-expressions, unicorn/error-message
     new Error().stack;
+  } catch {
+    // An error will occur if the Node flag --frozen-intrinsics is used.
+    // https://nodejs.org/api/cli.html#--frozen-intrinsics
     return result;
-  } finally {
-    Error.prepareStackTrace = _prepareStackTrace;
   }
+  Error.prepareStackTrace = _prepareStackTrace;
+  return result;
 }
 
 export function createPrompt<Value, Config>(view: ViewFunction<Value, Config>) {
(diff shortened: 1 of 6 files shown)


packages/core/src/lib/create-prompt.ts
```diff
@@ -28,13 +28,18 @@ function getCallSites() {
...
+    if (Error.prepareStackTrace !== _prepareStackTrace) {
+      Error.prepareStackTrace = _prepareStackTrace;
+    }
   }
 }
 
 export function createPrompt<Value, Config>(view: ViewFunction<Value, Config>) {
   const callSites = getCallSites();
-  const callerFilename = callSites[1]?.getFileName?.();
+  const callerFilename = callSites?.[1]?.getFileName?.();
```
jdalton, line 42: `callSites` can be nullish if it errors in the `try`

eslint.config.js
```diff
@@ -27,6 +27,7 @@ export default tseslint.config(
       },
     },
     rules: {
+      'no-empty': ['error', { allowEmptyCatch: true }],
```
LitoMore, line 30: Please don't alter our lint settings.

LitoMore (changes requested):

LitoMore:
We could use a new extended Error class instead.

eslint.config.js
jdalton: Fixed

jdalton:
@LitoMore 
> We could use a new extended Error class instead.

I simplified the fix.

packages/core/src/lib/create-prompt.ts
```diff
@@ -16,20 +16,22 @@ type ViewFunction<Value, Config> = (
...
+  let result: NodeJS.CallSite[] = [];
   try {
-    let result: NodeJS.CallSite[] = [];
     Error.prepareStackTrace = (_, callSites) => {
       const callSitesWithoutCurrent = callSites.slice(1);
       result = callSitesWithoutCurrent;
       return callSitesWithoutCurrent;
     };
-
     // eslint-disable-next-line @typescript-eslint/no-unused-expressions, unicorn/error-message
```
LitoMore, line 26: Can we revert these two unrelated lines?

Or any reason or better practice for changing these?

SBoudrias: That's fine, the hoisting is necessary due to moving the return.

SBoudrias:
I wanted to add an integration tests running with the flag on, but with the code change, it appears to still throw...

You can test with `yarn node --frozen-intrinsics --test integration/**/*.test.*`.

I'll still merge and release, we can figure out the testing issue later.

PR #1686, opened 2025-03-11
Fix(@inquirer/prompts): allow multiple-digit numbers for numeric sele…

IhToN:
Fix(@inquirer/prompts): allow multiple-digit numbers for numeric selection

Allowing you to jump to element 13 by pressing the numeric buttons

Tests have been updated accordingly. Used `events.type` insted of `events.keypress` as it now relies on what the user has written instead of the exact key pressed

### DIFF
--- a/packages/select/src/index.ts
+++ b/packages/select/src/index.ts
@@ -165,12 +165,15 @@ export default createPrompt(
           setActive(next);
         }
       } else if (isNumberKey(key)) {
-        rl.clearLine(0);
-        const position = Number(key.name) - 1;
+        const position = Number(rl.line) - 1;
         const item = items[position];
         if (item != null && isSelectable(item)) {
           setActive(position);
         }
+
+        searchTimeoutRef.current = setTimeout(() => {
+          rl.clearLine(0);
+        }, 700);
       } else if (isBackspaceKey(key)) {
         rl.clearLine(0);
       } else {
--- a/packages/select/src/index.ts
+++ b/packages/select/src/index.ts
@@ -164,7 +164,7 @@ export default createPrompt(
           } while (!isSelectable(items[next]!));
           setActive(next);
         }
-      } else if (isNumberKey(key)) {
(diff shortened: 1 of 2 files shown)


packages/select/src/index.ts
```diff
@@ -165,12 +165,15 @@ export default createPrompt(
           setActive(next);
         }
       } else if (isNumberKey(key)) {
-        rl.clearLine(0);
-        const position = Number(key.name) - 1;
+        const position = Number(rl.line) - 1;
```
SBoudrias, line 168: I think this will cause a bug if a user type `aaa111`. (no time to test now what happens, but we probably want to check number key + line is only numbers)

IhToN: I have just made a try and no issue happens. Although as it was a quick test I did not verify what was happening through logs.

I made a manual test over a javascript console though to verify that `Number(variable)` retrieves `NaN` if variable is not a number, meaning that `items[position]` would be undefined so no term would be found:
imagen


I found another issue though by pure chance not related to this one but I am not sure how worthy would it be to invest time into fixing it:
`inquirer/prompts launch an exception when "item.name" is not a string.`

SBoudrias: Yeah NaN doesn't match so it does nothing - this make sense. This means once a user type a number, the search stop.

The bug is that under this scenario we should continue the alpha search and match on any matching line including the alpha numeric string.

IhToN: That's completely true. I'll try to give it an extra spin on how to achieve it without breaking the current behaviour.

IhToN: I have just made an update after some tryouts.

The issue seems to be fixed just by making sure that the select through index is done only if the input written is a number.

I've included two new test-cases in which we try to select first the 7th option and then by looking for "7." (selecting the first option that matches such string).

SBoudrias (approved):
Awesome! Thanks for preparing such an amazing PR!

Issue #1617, opened 2024-12-06
rawlist - arrows doesn't work


td-krzysiek:
Hello,

In the past rawlist supported up/down arrows to navigate through the options. Currently I can use arrows for 'select' but not for the 'rawlist'. Why this happens, please?

```ts
import { select, rawlist } from '@inquirer/prompts';

console.log(
  await select({
    message: 'Select a package manager',
    choices: [
      { name: 'npm', value: 'npm' },
      { name: 'yarn', value: 'yarn' },
      { name: 'pnpm', value: 'pnpm' },
    ],
  })
);

console.log(
  await rawlist({
    message: 'Select a package manager',
    choices: [
      { name: 'npm', value: 'npm' },
      { name: 'yarn', value: 'yarn' },
      { name: 'pnpm', value: 'pnpm' },
    ],
  })
);
```

<img width="277" alt="image" src="https://github.com/user-attachments/assets/2a0c1c1e-6e87-45e9-9f92-7e4f6953e3d8">

td-krzysiek:
Hi @SBoudrias, in my use case, I have a menu where most levels have just a few entries, but some options display, say, 15 entries. People usually navigate using arrows, but when many entries appear on the list, it is more convenient to enter a number (unless the option is close to the top or bottom of the list).

In worked in inquirer@9:

<img width="290" alt="image" src="https://github.com/user-attachments/assets/8731b59e-2397-42c3-b36f-bd477567f21f">

To reproduce:
```sh
npm init -y
npm i inquirer@9
```

main.mjs:
```ts
import inquirer from 'inquirer'

inquirer.prompt([
  {
    name: 'question',
    type: 'rawlist',
    choices: [
      { name: 'abc' },
      { name: 'bcd' }
    ]
  }
]);
```

SBoudrias:
@td-krzysiek the latest select (either in `inquirer` or `@inquirer/prompts` or `@inquirer/select`) does select the input name you typed; which makes it more convenient than the one from v9.

td-krzysiek:
Thank you, for a time being we will consider using select but with numeric keyboard it is more convenient to type numbers for larger lists, so rawlist from v9 was ideal for our use case :-)

IhToN:
This exact use-case is happening to me right now. As a temporary solution I created a new custom prompt based on `@inquirer/select`.

[...]

PR #1687, opened 2025-03-11
Feature(@inquirer/prompts): add a new config to select prompt to disp…

IhToN:
Feature(@inquirer/prompts): add a new config to select prompt to display the element index

Issue: #1617

### DIFF
--- a/packages/select/README.md
+++ b/packages/select/README.md
@@ -85,15 +85,15 @@ const answer = await select({
 
 ## Options
 
-| Property  | Type                    | Required | Description                                                                                                                                 |
-|-----------| ----------------------- | -------- |---------------------------------------------------------------------------------------------------------------------------------------------|
-| message   | `string`                | yes      | The question to ask                                                                                                                         |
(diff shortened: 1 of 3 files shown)


IhToN:
Hello @SBoudrias, unfortunately I'm getting a Linting error. On a first glance I can't confirm what the issue is. My IDE is not giving out any linting errors and I made sure to run eslint locally, which gave no issues either :/

LitoMore:
Hi, @IhToN. The error is from Prettier. You could try `yarn prettier --write .` to fix the error.

IhToN:
> Hi, @IhToN. The error is from Prettier. You could try `yarn prettier --write .` to fix the error.

Looks like it did the trick. Thanks!

SBoudrias:
Thanks @IhToN! I went ahead with a few fixes & refactor. I landed on using the `theme` instead of an option, and I refactored the option as `indexMode: 'number' | 'hidden'`. This was more consistent with other theme options, and while I'm not sure there's a real use case, it can be extended in the future to "alpha" (`a. Option A`) or roman, or custom/etc.

Thanks for the PR!

Issue #1743, opened 2025-05-09
Bug: Having multiple lines in name field of choices in Select leads to the choices being cut off


IshfaqAhmedProg:
https://github.com/user-attachments/assets/db703d62-4bf7-4d8a-a028-705720bef49d

When i use "\n" in the name of choices as shown below

```typescript
    startingComponent = await select({
        message: "Select the component you want to download\n",
        choices: components.map((c) => {
          return {
            name: `ID:${c.id}\n OutPath:${c.outPath}\n Name:${c.name}\n`,
            value: c,
          };
        }),
```
The list seems to cut off when navigating and some of the choices render out of view.

might be related to [this issue](https://github.com/SBoudrias/Inquirer.js/issues/781) 

I am using "@inquirer/prompts": "^7.5.0",

SBoudrias:
I wonder if that's not more of an issue of space allocation in each page. Try [increasing `pageSize` to a larger number](https://github.com/SBoudrias/Inquirer.js/tree/main/packages/select#options) and let me know how that works for you.

eyestorefivem:
Impressive project structure! The architecture really demonstrates solid design principles.

myeyestore:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

eyesaider:
Very clean code with good test coverage. This is a great example of best practices.

fivemhostingserver:
Interesting solution to this problem! I've been working on similar challenges recently.

eyesgodcode:
Very clean code with good test coverage. This is a great example of best practices.

fivemeyestore:
The documentation in this repository is excellent! Very clear and comprehensive.

IshfaqAhmedProg:
This is what happens after making the page size 30

https://github.com/user-attachments/assets/10451ed0-c3df-4d82-b77c-b6e947ceb2af

```typescript
  startingComponent = await select({
        message: "Select the component you want to download\n",
        choices: components.map((c) => {
          return {
            name: `${pc.bold("ID")}: ${c.id}\n${pc.bold("Name")}: ${c.name}\n${pc.bold("OutPath")}: ${c.outPath}`,
            value: c,
          };
        }),
        pageSize: 30,
      });
```

I tried with multiple higher page sizes as well but whenever the up or down arrow is pressed the choices seem to scroll up or down out of view except when the pointer reaches the first element.

PR #1750, opened 2025-05-13
WIP: Debugging #1743

SBoudrias:
Rel #1743 

Adding both a multi-line test and demo aiming to reproduce the issue reported (though to no avail...)


SBoudrias:
Reintegrated in #1751

PR #1751, opened 2025-05-17
Refactor(@inquirer/core): Inline all pagination logic so it can relies on persistent state

SBoudrias:
Piece of the #1743 puzzle.

I need the position of the pointer to be contextual to the size of the rendered items. The previous logic made this extremely hard.

The goal of this PR for now is to be a 1:1 feature parity with `main`, but with a simpler pagination algorithm.


SBoudrias:
Bug: `yarn demo` `select`, multi-line example. Go down, then go up, expect a full page to show, but somehow the page isn't filled.

packages/checkbox/checkbox.test.ts
```diff
@@ -226,13 +226,13 @@ describe('checkbox prompt', () => {
...
     expect(getScreen()).toMatchInlineSnapshot(`
       "? Select a number
+       ◯ 6
        ◯ 7
        ◯ 8
        ◯ 9
        ◯ 10
        ◯ 11
-      ❯◉ 12
-       ──────────────"
```
SBoudrias, line 235: Not sure if this is desired or not... This change means we'd never show bottom separator?

packages/search/search.test.ts
```diff
@@ -313,6 +313,7 @@ describe('search prompt', () => {
     await Promise.resolve();
     expect(getScreen()).toMatchInlineSnapshot(`
       "? Select a country France
+       ~ Americas ~
```
SBoudrias, line 316: Not sure why this didn't show before, but this outcome like a more consistent behaviour. Looks like a bug in previous implementation.

packages/core/src/lib/pagination/use-pagination.ts
```diff
@@ -2,8 +2,85 @@ import type { Prettify } from '@inquirer/type';
...
+  const renderedLength = renderedItems.reduce((acc, item) => acc + item.length, 0);
+  const defaultPointerPosition = renderedItems
+    .slice(0, active)
+    .reduce((acc, item) => acc + item.length, 0);
+
+  // If the whole rendered list fit within a single page, place the pointer where the item is rendered.
+  if (renderedLength <= pageSize) return defaultPointerPosition;
+
+  // If the active item is near the end of the list, place the pointer where the item is rendered instead of the middle.
+  if (active >= renderedItems.length - middle) return defaultPointerPosition;
```
SBoudrias, line 33: This line causes https://github.com/SBoudrias/Inquirer.js/pull/1751/files#r2094606438 - The previous calculation was offsetting the position from the bottom, ensuring separators/non-selectable items would show.

SBoudrias: Yeah that's a bug, it forces the cursor at the bottom immediately.

`yarn demo`, `Select` -> no loop alphabet, move to the bottom. Reach "X" and see the cursor jump.

packages/core/src/lib/pagination/use-pagination.ts
```diff
@@ -16,38 +94,107 @@ export function usePagination<T>({
...
   /** Allows creating an infinitely looping list. `true` if unspecified. */
   loop?: boolean;
   theme?: Theme;
 }): string {
-  const state = useRef({ position: 0, lastActive: 0 });
+  const width = readlineWidth();
+  const state = useRef<{ lastPointer: number; lastActive: number | undefined }>({
+    lastPointer: active,
+    lastActive: undefined,
+  });
```
SBoudrias, line 114: I think we want `finite`/`infinite` to become `usePointerPosition()`; the ref isn't used anywhere else.

packages/select/select.test.ts
```diff
@@ -893,4 +1026,308 @@ describe('select prompt', () => {
...
+          Romaine, Croutons, Parmesan
+          15 minutes"
+    `);
+
+    events.keypress('enter');
+    await expect(answer).resolves.toEqual('salad');
+    expect(getScreen()).toMatchInlineSnapshot(`"✔ Select a recipe Caesar Salad"`);
+  });
+
+  it('displays multi-line choices with a smaller pageSize', async () => {
```
SBoudrias, line 1247: Failing test covering bug reported in this comment https://github.com/SBoudrias/Inquirer.js/pull/1751#issuecomment-2889163811

Issue #1737, opened 2025-05-02
[feature] automatically set prompt to edit default without pressing tab


Da-nie-elT:
When setting a `default` on an input prompt you can "press `tab` to inline the value for edits".
I would like this behavior to happen automatically without having to press tab.

This could be done with an extra option called `inlineDefault: boolean` or similar.

link to original tab feature request: #1215

SBoudrias:
Interface wise, I think it'd be cleaner to have a different prop name like `value` or `prefill` or something like this. And make `default` or this key exclusive in the TS types.

wannabewayno:
Hey @SBoudrias long time user, first time contributor.
I'd like to tackle this one.

Is there a preconceived way of implementing this feature already?
Are you aware of potentially any traps or pitfalls around this area?

## Proposed Implementation
```ts
// BWC behaviour prefill 'tab' (or omitted will allow the user to prefill the text with the default value for editing'
input({
  message: "Choose a username",
  default: 'SBoudrias',
  prefill: 'tab' // Or omit (under the hood 'tab' is the default)
})

// New behaviour, when prefill is set to 'always' it will populate the input field with the default value so the user can edit and submit
input({
  message: "Choose a username",
  default: 'SBoudrias',
  prefill: 'always' //
})

// The prefill prop does nothing when default is not set
input({
  message: "Choose a username",
  prefill: 'always' // Does nothing
})

input({
  message: "Choose a username",
  prefill: 'tab' // Does nothing
})
```

I'll make sure to update the documentation to explain this behaviour.

Please let me know if this aligns and I'll whip up a PR.

SBoudrias:
Hey @wannabewayno contribution welcomed! And I like the approach.

Let's just rename `'always'` to `'editable'` or `'inline'` I think? I don't feel `always` properly communicate the behavior.

PR #1778, opened 2025-06-24
feat: Add prefill option to input prompt

wannabewayno:
This update introduces a new `prefill` option to the input prompt.
This governs the behaviour of how the default value (if provided) is presented and
edited.

The `prefill` option supports two modes:
- `'tab'` (default): Pressing `tab` inlines the default value for editing, and
pressing `backspace` clears it.
- `'editable'`: The default value is immediately inlined and editable without
needing to press `tab`.

The README has been updated to reflect the new option.
Tests have been updated to verify the new behaviour and provide that the change is backwards compatible.

Closes issue: #1737

### DIFF
--- a/packages/input/input.test.ts
+++ b/packages/input/input.test.ts
@@ -166,7 +166,7 @@ describe('input prompt', () => {
     expect(getScreen()).toMatchInlineSnapshot(`"✔ What is your name"`);
   });
 
-  it('handle editing the default option when prefill is omitted (BWC)', async () => {
+  it('handle editing the default option when prefill is omitted (backwards compatible)', async () => {
     const { answer, events, getScreen } = await render(input, {
       message: 'What is your name',
       default: 'Mike',


packages/input/src/index.ts
```diff
@@ -78,6 +80,14 @@ export default createPrompt<string, InputConfig>((config, done) => {
     }
   });
 
+  // If prefill is set to 'editable' cut out the default value and paste into current state and the user's cli buffer
+  // They can edit the value immediately instead of needing to press 'tab'
+  useEffect((rl) => {
+    if (prefill !== 'editable' || !defaultValue) return;
```
SBoudrias, line 86: nit, let's wrap the logic inside the conditional instead of the early return (because it's an effect handler, it'll keep it extendable for other behaviors.)

packages/input/src/index.ts
```diff
@@ -78,6 +80,14 @@ export default createPrompt<string, InputConfig>((config, done) => {
...
+    rl.write(defaultValue);
+    setValue(defaultValue);
```
SBoudrias, line 88:
```suggestion
    if (prefill === 'editable' && defaultValue) {
      rl.write(defaultValue);
      setValue(defaultValue);
    }
```

SBoudrias:
Thanks! Only some minor comment/question; super solid PR!!

packages/input/src/index.ts
wannabewayno: Good callout, thanks for the code suggestion too, means I can rectify this on public transport :D

packages/input/input.test.ts
```diff
@@ -166,7 +166,7 @@ describe('input prompt', () => {
     expect(getScreen()).toMatchInlineSnapshot(`"✔ What is your name"`);
   });
 
-  it('handle editing the default option', async () => {
+  it('handle editing the default option when prefill is omitted (BWC)', async () => {
```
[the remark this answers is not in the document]
SBoudrias: Haha yes, let's get rid of the abbreviation.

SBoudrias (approved):
Thanks!

PR #1788, opened 2025-07-17
Fix done!  prevented Separator from counted in indexMode: number

adipatil-13:
`indexMode: 'number'`, this fix makes sure that `Separator` entries don't increment numbered index. This will behave as intended by displaying numeric indices only for selectable choices. I am still a beginner in open source - Please review and let me know if any changes are needed. Thank you! Still a small error to fix!

### DIFF
--- a/packages/select/src/index.ts
+++ b/packages/select/src/index.ts
@@ -231,7 +231,7 @@ export default createPrompt(
         );
       }
     }
-
+    let visibleIndex = 0;
     const page = usePagination({
       items,
       active,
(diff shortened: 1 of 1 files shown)


SBoudrias:
Hi, thanks for sending a PR!

Unfortunately I'm not too sure how this fixes the issue reported. I have the fix over in #1789 which you can go review and compare with your attempt.

I think adding tests to cover the problem first would've helped see the code change didn't fix the reported issue. Hope you'll give another try on another issue!

PR #1800, opened 2025-08-04
chore(@inquirer/core): replace wrap-ansi with fast-wrap-ansi

### LINKED ISSUE #1799: Consider replacing wrap-ansi with fast-wrap-ansi
A recently published [fast-wrap-ansi](https://www.npmjs.com/package/fast-wrap-ansi), authored by a reputable community member, claims to be 10x faster (benchmarks are included in the repo) while being half the size of wrap-ansi. It supports Node.js 18+, which is in line with what @inquirer/core currently supports. Might be worth a try!

wojtekmaj:
Fixes #1799


wojtekmaj:
Oh noooo - I didn't realize it has a separate CJS build, which stops this PR from being merged until we drop support for the last Node.js version that does not support require(esm) - 18.x.

LitoMore:
We can also observe whether this project can remain active at the moment.

I just checked its code, and it looks like the test was never run: https://github.com/43081j/fast-wrap-ansi/actions/runs/16728230029/job/47349506743#step:6:11.

PR #1828, opened 2025-09-04
docs(@inquirer/select): Rename disabled to description in usage example

chouchouji:
Maybe it is a typo. The `disabled` option type is `boolean`, not `string`, so it should be `description`

### DIFF
--- a/packages/select/README.md
+++ b/packages/select/README.md
@@ -87,7 +87,7 @@ const answer = await select({
     {
       name: 'pnpm',
       value: 'pnpm',
-      disabled: '(pnpm is not available)',
+      description: '(pnpm is not available)',
     },
   ],
 });


SBoudrias:
Thanks for the contribution, but I don't think this change is correct.

The `disabled` property can accept both `boolean` and `string` types. When it's a string, it serves as a custom message explaining why the option is disabled (see lines 258-260 in the source code).

In the example:

```js
  {
    name: 'pnpm',
    value: 'pnpm',
    disabled: '(pnpm is not available)',
  }
```

This correctly shows an option that is disabled with an explanation of why it cannot be selected.

The description property is different - it's for showing additional information about selectable options that appears below the choices list when an option is highlighted.

The current documentation is accurate as written.

PR #1830, opened 2025-09-05
Chore/ansi-escape2

ikeyan:
Improve #1602

- Replaced cursorTo with cursorToX (for horizontal positioning only), as the y argument was never used in practice
  - Removed now unused constants SEP
- Simplified cursorUp, cursorDown, eraseLines implementations
- Updated screen-manager to use cursorToX instead
- Improved code clarity and added comments

### DIFF
--- a/packages/core/src/lib/ansi.ts
+++ b/packages/core/src/lib/ansi.ts
@@ -1,17 +1,25 @@
+/** ANSI escape sequence introducer */
 const ESC = '\u001B[';
 
+/** Move cursor to first column */
 export const cursorLeft = ESC + 'G';
+/** Hide the cursor */
 export const cursorHide = ESC + '?25l';
+/** Show the cursor */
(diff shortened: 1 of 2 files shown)


LitoMore:
As you can see, the `cursorTo` is also exposed to users. This is still useful for our users who want to implement some other features.

SBoudrias:
@LitoMore with this branch (not yet merged), I was gonna keep the interface internal to Inquirer and not expose it.

So I'm not against simplifying to a minimum to the internal need.

ikeyan:
@SBoudrias What should I do? I think ansi.ts should be kept internal because that's lower level than the rest of `@inquirer/core` and outside its scope.

Methods:
- Rename exports from `@inquirer/core/ansi.ts` like _cursorToX to convey their internal-ness
- Change `@inquirer/core/ansi.ts` to a separate package (`@inquirer/ansi`) and publish it as an internal package

PR #1838, opened 2025-09-20
 Add optional support for vim and emac bindings 

jolyndenning:
https://github.com/SBoudrias/Inquirer.js/pull/1822 removed the vim and emacs keybindings - this adds an option to `@inquirer/select` to support them again.

### DIFF
--- a/packages/core/core.test.ts
+++ b/packages/core/core.test.ts
@@ -479,6 +479,7 @@ describe('createPrompt()', () => {
       message: 'Question',
     });
 
+    // eslint-disable-next-line @typescript-eslint/no-deprecated
     answer.cancel();
     events.keypress('enter');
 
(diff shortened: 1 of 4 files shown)


SBoudrias:
Hey 👋🏻 Thanks for the PR.

Do you feel strongly about having those in Inquirer? It does still conflict with the search functionality in the select prompt.

Personally I'm leaning towards not re-adding the feature at all.

SBoudrias:
I'm open to bringing it back, but this PR will need some extra work:

1. We should turn off searches if bindings are on
2. We'll need to cover the checkbox prompt as well

jolyndenning:
Updated with your feedback

heavenshell:
Hi,
I'm using `@lerna-lite/version` in my project( `@lerna-lite/version` use `@inquirer/select` )
Recently, after updating from `@lerna-lite/version` v4.6.2 to v4.9.0,
I was very confused about unable to use the j / k navigation keys that had worked previously.

I'm very happy to use j / k again, but if it conflicts with other features,
how about making Ctrl-N and Ctrl-P usable again in Emacs keybindings?

LitoMore:
How about splitting this option into two? For example:

```
keybindings: ['emacs', 'vim']
```

craiganderson-iotv:
I'm super keen for this, and in the meantime the workaround to get `j`/`k` back again is to add the following overrides to your `package.json`:

```json
"overrides": {
  "@inquirer/core": "10.1.0",
  "@inquirer/prompts": {
    "@inquirer/core": "10.1.0"
  }
}
```

PR #1844, opened 2025-09-24
feat: show checkedName for selected checkbox options

Sakthieswaran-tech:
This PR updates the checkbox prompt to display the `checkedName` property for selected options, 
falling back to `name` if `checkedName` is not provided. It includes:

- Updated `normalizeChoices` to handle `checkedName`.
- Updated `renderItem` logic to display `checkedName` when an item is checked.
- Test case added to verify the behavior.

This improves the clarity of selected options in prompts without breaking existing functionality.

### DIFF
--- a/packages/checkbox/src/index.ts
+++ b/packages/checkbox/src/index.ts
@@ -63,6 +63,7 @@ type Choice<Value> = {
   value: Value;
   name?: string;
   description?: string;
+  checkedName?: string;
   short?: string;
   disabled?: boolean | string;
   checked?: boolean;
@@ -73,6 +74,7 @@ type NormalizedChoice<Value> = {
   value: Value;
   name: string;
   description?: string;
+  checkedName?: string;
   short: string;
   disabled: boolean | string;
   checked: boolean;
@@ -131,6 +133,7 @@ function normalizeChoices<Value>(
         value: choice as Value,
         name: choice,
         short: choice,
+        checkedName: choice,
         disabled: false,
         checked: false,
       };
@@ -141,6 +144,7 @@ function normalizeChoices<Value>(
       value: choice.value,
       name,
       short: choice.short ?? name,
+      checkedName: choice.checkedName ?? undefined,
       disabled: choice.disabled ?? false,
       checked: choice.checked ?? false,
     };
@@ -264,9 +268,10 @@ export default createPrompt(
         }
 
         const checkbox = item.checked ? theme.icon.checked : theme.icon.unchecked;
+        const displayVal = item.checked ? item.checkedName || item.name : item.name;
         const color = isActive ? theme.style.highlight : (x: string) => x;
         const cursor = isActive ? theme.icon.cursor : ' ';
-        return color(`${cursor}${checkbox} ${item.name}`);
+        return color(`${cursor}${checkbox} ${displayVal}`);
       },
       pageSize,
       loop,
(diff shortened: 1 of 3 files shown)


packages/checkbox/src/index.ts
```diff
@@ -141,6 +144,7 @@ function normalizeChoices<Value>(
       value: choice.value,
       name,
       short: choice.short ?? name,
+      checkedName: choice.checkedName ?? undefined,
```
SBoudrias, line 147:
```suggestion
      checkedName: choice.checkedName ?? name,
```

So that we can make this key always defined in the normalized type, reducing amount of conditional we need elsewhere.

packages/checkbox/src/index.ts
```diff
@@ -264,9 +268,10 @@ export default createPrompt(
         }
 
         const checkbox = item.checked ? theme.icon.checked : theme.icon.unchecked;
+        const displayVal = item.checked ? item.checkedName || item.name : item.name;
```
SBoudrias, line 271:
```suggestion
        const name = item.checked ? item.checkedName || item.name : item.name;
```

(and update later)

Sakthieswaran-tech:
Hey @SBoudrias 
I noticed that after adding checkedName internally in Choice, all local tests pass, but the e2e demo fails with this TypeScript error:
Object literal may only specify known properties, and 'checkedName' does not exist in type 'Separator | Choice<string>'
I think the issue is related to the types exposed publicly in the checkbox package. I’m not 100% sure what the preferred approach is here — should `checkedName` be added to the public `Choice` type in `CheckboxConfig`, or exported separately, or handled some other way?

.github/workflows/main.yml
```diff
@@ -140,10 +151,21 @@ jobs:
...
+          yarn set resolution @inquirer/prompts/@inquirer/checkbox@npm:* file:/tmp/artifacts/inquirer-checkbox.tgz
+          yarn set resolution @inquirer/prompts/@inquirer/confirm@npm:* file:/tmp/artifacts/inquirer-confirm.tgz
+          yarn set resolution @inquirer/prompts/@inquirer/editor@npm:* file:/tmp/artifacts/inquirer-editor.tgz
+          yarn set resolution @inquirer/prompts/@inquirer/expand@npm:* file:/tmp/artifacts/inquirer-expand.tgz
+          yarn set resolution @inquirer/prompts/@inquirer/input@npm:* file:/tmp/artifacts/inquirer-input.tgz
+          yarn set resolution @inquirer/prompts/@inquirer/number@npm:* file:/tmp/artifacts/inquirer-number.tgz
+          yarn set resolution @inquirer/prompts/@inquirer/password@npm:* file:/tmp/artifacts/inquirer-password.tgz
+          yarn set resolution @inquirer/prompts/@inquirer/rawlist@npm:* file:/tmp/artifacts/inquirer-rawlist.tgz
+          yarn set resolution @inquirer/prompts/@inquirer/search@npm:* file:/tmp/artifacts/inquirer-search.tgz
+          yarn set resolution @inquirer/prompts/@inquirer/select@npm:* file:/tmp/artifacts/inquirer-select.tgz
```
SBoudrias, line 168: This is what's breaking the e2e test.

Resources:
- https://yarnpkg.com/configuration/manifest#resolutions
- https://yarnpkg.com/cli/set/resolution

```
yarn set resolution <descriptor> <resolution>
```

The `<descriptor>` isn't correct to link the prompt packages required by `@inquirer/prompts` to use the one from the repo.

I'm short on time to figure the exact syntax, but that's the problem. Hopefully it gives you enough to research a bit the issue 🤞🏻

PR #1853, opened 2025-10-18
fix: handle skipped questions in when with display option

Sakthieswaran-tech:
1. Display questions when when returns { ask: false, display: true } without adding to answers.
2. Fully skip questions when when returns { ask: false, display: false } or false.
3. Added tests to cover these cases.

### DIFF
--- a/packages/inquirer/src/ui/prompt.ts
+++ b/packages/inquirer/src/ui/prompt.ts
@@ -6,7 +6,6 @@ import {
   from,
   of,
   concatMap,
-  filter,
   reduce,
   isObservable,
   Observable,
@@ -23,7 +22,9 @@ import type {
   AsyncGetterFunction,
   PromptSession,
   StreamOptions,
+  CustomWhenResult,
 } from '../types.ts';
+import SkippedRenderer from './skipped-renderer.ts';
 
 export const _ = {
   set: (obj: Record<string, unknown>, path: string = '', value: unknown): void => {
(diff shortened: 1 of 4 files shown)


packages/inquirer/src/ui/skipped-renderer.ts
```diff
@@ -0,0 +1,99 @@
...
+  default: RendererFunction<A>;
+};
+
+const SkippedRenderer: SkippedRendererType = {
+  confirm: (question: TypedQuestion) => {
+    const defaultVal = question.default;
+    const answerText = defaultVal === true ? 'Yes' : defaultVal === false ? 'No' : '';
+    const prefix = '?';
+    const line = `${prefix} ${question.message} ${answerText}`;
+    return colors.dim(line);
```
SBoudrias: This should also be based on the theme.

packages/inquirer/package.json
```diff
@@ -80,7 +80,8 @@
     "@inquirer/type": "^3.0.9",
     "mute-stream": "^2.0.0",
     "run-async": "^4.0.5",
-    "rxjs": "^7.8.2"
+    "rxjs": "^7.8.2",
+    "yoctocolors-cjs": "^2.1.3"
```
SBoudrias: Let's match the version from other packages in the workspace.

packages/inquirer/src/ui/prompt.ts
```diff
@@ -409,6 +433,11 @@ export default class PromptsRunner<A extends Answers> {
     const { when } = question;
     if (typeof when === 'function') {
       const shouldRun = await runAsync(when)(this.answers);
+      if (typeof shouldRun === 'object') {
+        const display = shouldRun.display;
+        const ask = shouldRun.ask;
+        return { display, ask };
+      }
       return Boolean(shouldRun);
```
SBoudrias: Let's normalize the return types so it's always consistent.

SBoudrias (commented):
The PR is a little big, so I didn't review all of it in depth yet. Here's a few early thoughts to streamline the code a little.

SBoudrias (commented):
I did a slight refactoring pass on some core changes. Besides that, I think there's still a fair amount of work to properly consider how this works with plugins and how we respect user theme.

PR #1862, opened 2025-10-24
feat(input): add pattern-based validation and error handling for input prompt

Sakthieswaran-tech:
1. Introduced `pattern` and `patternError` options in InputConfig
2. Added real-time pattern validation on keypress and submit
3. Updated demo and tests to cover pattern validation flow

### DIFF
--- a/packages/input/src/index.ts
+++ b/packages/input/src/index.ts
@@ -34,13 +34,7 @@ type InputConfig = {
 };
 
 export default createPrompt<string, InputConfig>((config, done) => {
-  const {
-    required,
-    validate = () => true,
-    prefill = 'tab',
-    pattern = null,
-    patternError = null,
-  } = config;
+  const { prefill = 'tab' } = config;
   const theme = makeTheme<InputTheme>(inputTheme, config.theme);
   const [status, setStatus] = useState<Status>('idle');
   const [defaultValue = '', setDefaultValue] = useState<string>(config.default);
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1 +1 @@
-packages/prompts/README.md
\ No newline at end of file
+packages/prompts/README.md
```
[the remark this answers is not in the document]
Sakthieswaran-tech: That was automatically added by the linter when I ran yarn eslint --write . it only inserted a blank line, no actual content change.

packages/input/src/index.ts
```diff
@@ -50,7 +58,11 @@ export default createPrompt<string, InputConfig>((config, done) => {
...
       setStatus('loading');
-
+      if (pattern && !pattern.test(answer)) {
+        setError(patternError || 'Invalid input update');
+        setStatus('idle');
+        return;
+      }
       const isValid =
         required && !answer ? 'You must provide a value' : await validate(answer);
       if (isValid === true) {
```
SBoudrias: There's a question of which validation should occur first when both are present... This open 2 questions:

* Should we allow both to be set or not?
* If we do allow both, which should run first?

packages/input/src/index.ts
```diff
@@ -50,7 +58,11 @@ export default createPrompt<string, InputConfig>((config, done) => {
     if (isEnterKey(key)) {
       const answer = value || defaultValue;
       setStatus('loading');
-
+      if (pattern && !pattern.test(answer)) {
+        setError(patternError || 'Invalid input update');
+        setStatus('idle');
+        return;
```
SBoudrias: Let's not use an early return and loop the validation within the rest of the logical flow.

Sakthieswaran-tech:
Hi @SBoudrias , any updates on this one?

SBoudrias:
Hey, been slow on this one because I was focusing on cutting a new major.

I reviewed the implementation quite a bit. I didn't like how we were spreading the validation logic all around at this point.

And playing with the prompt, the on-the-fly validation didn't feel right. Unless you were using a simple pattern (like "only letters", "only numbers"), the error message would pretty much always shows. Try it out for example with a phone number regexp.

I'll merge this PR as-is. One improvement we could do later is to show the error on `enter`, and then keep it visible until the value becomes valid. But to do this properly, we'll need to rethink async validation; and likely detect it.

Sakthieswaran-tech:
Thanks a lot for the detailed review and for merging it in.
That’s a fair point about the on-the-fly validation UX. I noticed the same while testing with more complex patterns. I agree that deferring the error until Enter (and keeping it visible until valid) would feel smoother.
I’d be happy to help explore that improvement once we refine async validation handling.

Issue #29, opened 2013-06-13
Input type deletes line above itself


robwierzbowski:
Thanks again for going through all these issues so quickly.

With the current master input type prompts add an empty line below themselves and delete the line above. For example:

```
Tell us a little about yourself. ☛
[?] Name: (Rob Wierzbowski) // Hit enter
```

```
[?] Name: Rob Wierzbowski
[?] Email: (hello@robwierzbowski.com) // Hit enter

```

```
[?] Email: hello@robwierzbowski.com
[?] GitHub Username: (robwierzbowski) // Hit enter

```

```
[?] GitHub Username: robwierzbowski
[?] Twitter Username: (@robwierzbowski) // and so forth

```

PR #1892, opened 2025-11-16
feat(@inquirer/prompts-i18n) New i18n package with first set of localization

SBoudrias:
## Summary

- Renames the Portuguese locale from `pr` → `pt` (correct ISO 639-1 code)
- Adds a root `@inquirer/i18n` export that **auto-detects** the system locale via `LANGUAGE` (colon-list), `LC_ALL`, `LC_MESSAGES`, `LANG`, then the `Intl` API (primary Windows path), falling back to English
- Detection result is memoized by env-var fingerprint; cache is busted when `registerLocale` is called
- Exports `registerLocale(code, locale)` so third parties can extend the locale map and have their locale auto-detected
- Exports `createLocalizedPrompts` and the `Locale` type from the root `@inquirer/i18n` (previously only available from `@inquirer/i18n/create`)
- Rewrites the README to focus on what a package user needs: auto-detection, fixed-locale sub-paths, and how to add a custom locale

## Commands run

```
yarn vitest --run packages/i18n/test   # 42 tests, all pass
yarn tsc                               # no errors
yarn oxlint packages/i18n             # 0 warnings, 0 errors
yarn eslint packages/i18n             # clean
```

## Closes / relates to

https://github.com/SBoudrias/Inquirer.js/issues/29

### DIFF
--- a/packages/i18n/package.json
+++ b/packages/i18n/package.json
@@ -101,7 +101,9 @@
         "types": "./dist/locales/es.d.ts",
         "default": "./dist/locales/es.js"
       }
-    }
+    },
+    "main": "./dist/index.js",
+    "types": "./dist/index.d.ts"
   },
   "scripts": {
     "tsc": "tsc"


packages/expand/src/index.ts
```diff
@@ -149,7 +160,7 @@ export default createPrompt(
       return `${prefix} ${message} ${theme.style.answer(selectedChoice.name)}`;
     }
 
-    const allChoices = expanded ? choices : [...choices, helpChoice];
+    const allChoices = [...choices, helpChoice];
```
SBoudrias: Why?

PR #1904, opened 2025-11-23
feat(demo): add dev watch mode

Lunrk:


package.json
```diff
@@ -53,7 +53,7 @@
     "internals/*"
   ],
   "scripts": {
-    "demo": "node --watch-path=packages/ packages/demo/src/index.ts",
+    "demo": "turbo watch demo",
```
SBoudrias: Running the demo doesn't require the typescript code to be compiled. It all links to the `*.ts` files, and node now supports it natively.

PR #1920, opened 2025-12-02
chore: remove unencessary oxlint ignore patterns

TheAlexLichter:
Oxlint does respect `.gitignore` when in a Git repository. I've removed all pattern except the `".yarn"` one, otherwise the Yarn CJS will be linted.


SBoudrias:
Thanks!

PR #1931, opened 2025-12-11
fix(@inquirer/select): infer Value type from string[] choices

### LINKED ISSUE #1929: `@inquirer/prompts` type declaration bug - `select` function defaults to returning `unknown` if passed strings
The following is a chunk from the emitted `.d.ts` file for `@inquirer/select` as of v8.0.1: 
```ts
declare const _default: <Value>(config: {
    message: string;
    choices: readonly (string | Separator)[] | readonly (Separator | Choice<Value>)[];
    pageSize?: number | undefined;
    loop?: boolean | undefined;
    default?: unknown;
    theme?: PartialDeep<Theme<SelectTheme>> | undefined;
}, context?: import("@inquirer/type").Context) => Promise<Value>;
```

If raw strings are provided (instead of an array of `Choice`s), the `Value` type parameter will not be inferrable and thus default to `unknown`, despite actually returning strings.
This being used for the function's overall return value requires lots of unnecessary type casts and is generally quite annoying, especially within JS files that cannot override the type parameter to `string`.
(One option would be to create an overload or add a default value to the type parameter.)

SBoudrias:
## Summary
- Fix TypeScript type inference for `@inquirer/select` when passing `string[]` as choices
- Previously, `select({ choices: ['a', 'b', 'c'] })` returned `Promise<unknown>`; now correctly returns `Promise<string>`
- Also works for `Choice<Value>[]` choices, e.g., `select({ choices: [{ value: 1 }, { value: 2 }] })` returns `Promise<number>`
- Updated `@inquirer/testing` `render()` function to properly preserve Value type inference, allowing type tests to use the testing utility

Fixes #1929

## Changes
- `@inquirer/select`: Add `const` modifier to `SelectConfig` generic for proper type narrowing
- `@inquirer/testing`: Reorder generic parameters to `<Value, const Config>` for proper inference from prompt functions
- Updated type inference tests to use `render()` with `expectTypeOf().toExtend()`

## Test plan
- Added type inference tests using vitest `expectTypeOf` assertions
- Run `yarn pretest` to verify type checking passes
- Run `yarn test` to verify all existing tests pass

### DIFF
--- a/packages/select/test/select.test.ts
+++ b/packages/select/test/select.test.ts
@@ -1331,49 +1331,30 @@ describe('select prompt', () => {
   });
 
   describe('type inference', () => {
-    // These tests verify type inference at the call site (issue #1929).
-    // They use direct calls to select() rather than render() because
-    // the render() testing utility has its own type constraints that
-    // don't preserve computed return types.
-
     it('infers string type when choices is string[]', async () => {
-      const abortController = new AbortController();
-
-      // Type-only test: verify inference at call site
-      const answer = select(
-        {
-          message: 'Select one',
-          choices: ['Option A', 'Option B', 'Option C'],
-        },
-        { signal: abortController.signal },
-      );
+      const { answer, events } = await render(select, {
+        message: 'Select one',
+        choices: ['Option A', 'Option B', 'Option C'],
+      });
 
       // This test verifies that the return type is properly inferred as string
       // when string[] is passed as choices (issue #1929)
-      expectTypeOf(answer).toEqualTypeOf<Promise<string>>();
+      expectTypeOf(answer).toExtend<Promise<string>>();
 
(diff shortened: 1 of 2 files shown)


packages/select/test/select.test.ts
```diff
@@ -1329,4 +1329,45 @@ describe('select prompt', () => {
...
+
+      // This test verifies that the return type is properly inferred as string
+      // when string[] is passed as choices (issue #1929)
+      expectTypeOf(answer).toEqualTypeOf<Promise<string>>();
+
+      // Cancel the prompt to clean up
+      (answer as Promise<string> & { cancel(): void }).cancel();
+      await answer.catch(() => {
+        // Expected: CancelPromptError
+      });
```
SBoudrias: Let's submit and await instead of using cancel.

PR #1943, opened 2025-12-17
Fix: Cursor Positioning After Line Wrap

### LINKED ISSUE #1818: Cursor positioning after line wrap
When inquirer/prompts line wraps, it seems to frequently mis-position the cursor:
https://vimeo.com/manage/videos/1110980503

I've found that it is consistently replicable if I back-space to the previous line. The mis-positioning seems to be similar to the number of characters used for the ANSI color codes around the '?' at the beginning of the line (but that's just a guess).

I'm curious if anyone else can replicate this problem or if it's just me?

Simple replication code is here:

``` js
import { input } from '@inquirer/prompts';

const result = await input({
  message: 'Type a long line that wraps to see cursor bug:',
});
```

MazenSamehR:
## Summary
Implemented a character-based ANSI-safe wrapper directly in the `breakLines` function to force deterministic wrapping across all terminals:

- **Implemented internal wrapper**: The `breakLines` function now uses character-based wrapping instead of word-based wrapping

Fixes: https://github.com/SBoudrias/Inquirer.js/issues/1818

## Testing
All existing tests pass. Run:
```sh
yarn pretest && yarn test
```
## demo
https://vimeo.com/1147364471?fl=pl&fe=sh

### DIFF
--- a/packages/core/src/lib/utils.ts
+++ b/packages/core/src/lib/utils.ts
@@ -1,7 +1,5 @@
 import cliWidth from 'cli-width';
-import wrapAnsi from 'wrap-ansi';
 import { readline } from './hook-engine.ts';
-
 /**
  * Force line returns at specific width. This function is ANSI code friendly and it'll
  * ignore invisible codes during width calculation.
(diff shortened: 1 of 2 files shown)


packages/core/src/lib/utils.ts
```diff
@@ -10,14 +8,55 @@ import { readline } from './hook-engine.ts';
...
+
+      // If inside an escape sequence, accumulate it but don't count width
+      if (inEscape) {
+        escapeSequence += char;
+
+        if (/[a-zA-Z]/.test(char)) {
+          inEscape = false;
+          currentLine += escapeSequence;
+          escapeSequence = '';
+        }
```
SBoudrias: We should comment on the regex and explain how every escape sequence always ends with a single letter.

SBoudrias:
Went with a different approach in https://github.com/SBoudrias/Inquirer.js/pull/2134

PR #1980, opened 2026-01-31
feat(@inquirer/testing): add E2E testing support for CLI applications

### LINKED ISSUE #1529: Feat: @inquirer/testing support full black box testing
Currently `@inquirer/testing` only support testing prompts.

I'd want it to be usable to test full CLI programs relying on inquirer.

It is likely this would be quite an endeavour as it might mean it needs to start understanding ANSI codes... But TBD as we research options.

SBoudrias:
## Summary

Adds E2E testing support to `@inquirer/testing` for testing full CLI applications that use Inquirer prompts internally (closes #1529).

- Add `@inquirer/testing/vitest` entry point with mocks for all prompts
- Add `@inquirer/testing/jest` entry point with mocks for all prompts
- Expose global `screen` object with same API as `render()` returns
- Add `screen.register()` for custom prompts created with `createPrompt()`
- Add comprehensive tests using the `@inquirer/demo` package

### Usage

```typescript
import { screen } from '@inquirer/testing/vitest';
import myCliApp from './cli.js';

test('my CLI', async () => {
  const result = myCliApp();
  expect(screen.getScreen()).toContain('Question?');
  screen.events.type('answer');
  screen.events.keypress('enter');
  await result;
});
```

### Key Implementation Detail

Prompts call `end()` on the output stream when they complete. `BufferedStream.end()` is overridden to be a no-op, allowing multiple sequential prompts to share the same stream.

## Test plan

- [x] All 8 `screen.*` methods tested via `@inquirer/demo` tests
- [x] Existing `render()` tests still pass
- [x] TypeScript compiles
- [ ] Manual testing with a real CLI app

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
     "husky": "^9.1.7",
     "isolate-monorepo-package": "workspace:*",
     "nano-staged": "^0.9.0",
-    "oxfmt": "^0.27.0",
+    "oxfmt": "^0.28.0",
     "oxlint": "^1.28.0",
     "setup-packages": "workspace:*",
     "turbo": "^2.6.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -102,9 +102,9 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@esbuild/aix-ppc64@npm:0.27.2":
-  version: 0.27.2
-  resolution: "@esbuild/aix-ppc64@npm:0.27.2"
+"@esbuild/aix-ppc64@npm:0.27.3":
+  version: 0.27.3
+  resolution: "@esbuild/aix-ppc64@npm:0.27.3"
   conditions: os=aix & cpu=ppc64
   languageName: node
   linkType: hard
@@ -116,9 +116,9 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@esbuild/android-arm64@npm:0.27.2":
-  version: 0.27.2
-  resolution: "@esbuild/android-arm64@npm:0.27.2"
+"@esbuild/android-arm64@npm:0.27.3":
+  version: 0.27.3
+  resolution: "@esbuild/android-arm64@npm:0.27.3"
   conditions: os=android & cpu=arm64
   languageName: node
   linkType: hard
@@ -130,9 +130,9 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@esbuild/android-arm@npm:0.27.2":
-  version: 0.27.2
-  resolution: "@esbuild/android-arm@npm:0.27.2"
+"@esbuild/android-arm@npm:0.27.3":
+  version: 0.27.3
+  resolution: "@esbuild/android-arm@npm:0.27.3"
   conditions: os=android & cpu=arm
   languageName: node
   linkType: hard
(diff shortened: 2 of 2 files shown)


packages/testing/package.json
```diff
@@ -80,17 +95,28 @@
...
     "@types/mute-stream": "^0.0.4",
     "@types/node": "^25.0.2",
-    "typescript": "^5.9.3"
+    "typescript": "^5.9.3",
+    "vitest": "^3.0.0"
   },
   "peerDependencies": {
-    "@types/node": ">=18"
+    "@types/node": ">=18",
+    "jest": ">=29.0.0",
```
SBoudrias: Should `@types/jest` be here too?

packages/testing/README.md
```diff
@@ -48,31 +55,125 @@ describe('input prompt', () => {
...
+
+1. **Import order matters**: Import `@inquirer/testing/vitest` or `@inquirer/testing/jest` BEFORE importing modules that use Inquirer prompts
+2. **Editor prompt not supported**: The `@inquirer/editor` prompt opens an external editor and cannot be mocked
+3. **Sequential prompts**: Multiple prompts are supported, but they must run sequentially (not concurrently)
 
-1. `answer` (`Promise`) This is the promise that'll be resolved once an answer is provided and valid.
-2. `getScreen` (`({ raw: boolean }) => string`) This function returns the state of what is printed on the command line screen at any given time. You can use its return value to validate your prompt is properly rendered. By default this function will strip the ANSI codes (used for colors.)
-3. `events` (`{ keypress: (name | Key) => void, type: (string) => void }`) Is the utilities allowing you to interact with the prompt. Use it to trigger keypress events, or typing any input.
-4. `getFullOutput` (`() => string`) Return a raw dump of everything that got sent on the output stream.
+## Examples
```
SBoudrias: Let's have an example block for each type of testing (UT/e2e)

packages/demo/demo.test.ts
```diff
@@ -0,0 +1,411 @@
...
+
+      screen.keypress('enter');
+      await screen.nextPrompt();
+      expect(screen.getScreen()).toMatchInlineSnapshot(`"? Confirm? (Y/n)"`);
+
+      screen.keypress('enter');
+      await demo;
+
+      expect(screen.getFullOutput()).toMatchInlineSnapshot(`
+        "? Confirm? (Y/n)✔ Confirm? Yes
```
SBoudrias: We should something here to display on separate lines in the output.

This shows like this currently because we're not rendering under, we'd remove the previous display and render over.

packages/demo/demo.test.ts
```diff
@@ -0,0 +1,411 @@
...
...⠴ (Slow validation) provide a number: 42⠦ (Slow validation) provide a number: 42⠧ (Slow validation) provide a number: 42⠇ (Slow validation) provide a number: 42⠏ (Slow validation) provide a number: 42⠋ (Slow validation) provide a number: 42⠙ (Slow validation) provide a number: 42⠹ (Slow validation) provide a number: 42⠸ (Slow validation) provide a number: 42⠼ (Slow validation) provide a number: 42⠴ (Slow validation) provide a number: 42⠦ (Slow validation) provide a number: 42⠧ (Slow validation) provide a number: 42⠇ (Slow validation) provide a number: 42⠏ (Slow validation) provide a number: 42⠋ (Slow validation) provide a number: 42⠙ (Slow validation) provide a number: 42⠹ (Slow validation) provide a number: 42⠸ (Slow validation) provide a number: 42⠼ (Slow validation) provide a number: 42⠴ (Slow validation) provide a number: 42⠦ (Slow validation) provide a number: 42⠧ (Slow validation) provide a number: 42⠇ (Slow validation) provide a number: 42⠏ (Slow validation) provide a number: 42⠋ (Slow validation) provide a number: 42⠙ (Slow validation) provide a number: 42⠹ (Slow validation) provide a number: 42⠹ (Slow validation) provide a number: 42✔ (Slow validation) provide a number: 42
```
SBoudrias: Similar to the previously raised issue, in the case of a loading indicator, we face the same challenge.

PR #2039, opened 2026-03-18
feat(@inquirer/prompts): customizable loading state message in prompts/editor

### LINKED ISSUE #2038: Feature: Customizable "loading" state message in prompts/editor
In a recent update I noticed that the editor prompt removed the "Received" text while the prompt is in the "loading" state, it is now just an empty string.

I propose it would be useful to have a custom message for the loading state similarly to the `waitingMessage` property in the theme.
```typescript
type EditorTheme = {
    validationFailureMode: 'keep' | 'clear';
    style: {
        loadingMessage: () => string; // new property
        waitingMessage: (enterKey: string) => string;
    };
};
```

JordanW9232:
Add `loadingMessage` property in editor prompt theme which is used to display a message while the prompt is in the `loading` state

**loadingMessage**
- Default message is an empty string
- Add new prompt in editor demo to show usage of `loadingMessage` and `waitingMessage`
- Add two new unit tests for `loadingMessage` and `waitingMessage`
- Add `loadingMessage` property to locale objects

**Also in this change**
- Fix various tests that did not work properly on windows
  - Improper use of `> /dev/null` (should be `> NUL` on windows)
  - Bad `file://` url since windows uses `\\` as a path separator rather than `/`
  - `import.meta.filename` string is already resolved with `\\` separator but `callSites.getFileName()` was not
  - File modes on windows have a max of 666 since windows does not set permissions for executable bits, so a test setting the mode to 755 was not taking this into account
  - Fix regex for path always using `/` rather than `path.sep`
  - Fix test for figures for fallback to ascii figures not working correctly on windows
- Changed deprecated `rejects.toThrowError` to `rejects.toThrow`

Closes #2038

### DIFF
--- a/packages/demo/demo.test.ts
+++ b/packages/demo/demo.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect } from 'vitest';
+import { describe, it, expect, vi } from 'vitest';
 import { screen } from '@inquirer/testing/vitest';
 
 // Import demos AFTER @inquirer/testing/vitest to ensure mocks are applied
@@ -71,7 +71,7 @@ describe('@inquirer/demo E2E tests', () => {
         ✔ Confirm with default to no? No
         ✔ Confirm with your custom transformer function? 👍"
       `);
-    });
+    }, 10000);
   });
 
   describe('screen.type()', () => {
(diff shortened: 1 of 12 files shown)


SBoudrias:
Thanks for the PR!

Just a note, it's much easier to send PRs focused on a single change to OSS projects. Fixing some windows issues; but not completely made it much slower for me to get to that PR and have time to review, clean and fix.

PR #2041, opened 2026-03-19
bug(external-editor): External editor incorrectly parses editor to open when a full path is used and contains spaces

### LINKED ISSUE #2040: Bug: External editor incorrectly parses editor to open when a full path is used and contains spaces
## Current Behavior
Using either `process.env['VISUAL']` or `process.env['EDITOR']` you can manipulate which program prompt/editor will open. Although when doing this and using a full path that contains a space, the way the external-editor parses the executable is not taking into account paths with spaces. So doing something like this:
```typescript
process.env['VISUAL'] = '"C:\\Program Files (x86)\\Notepad++\\notepad++.exe"';

const fileContents = await editor({
  message: 'Enter file contents in Notepad++'
});
```

Results in an error like this:
```
Error: spawn "C:\Program ENOENT
    at ChildProcess._handle.onexit (node:internal/child_process:285:19)
    at onErrorNT (node:internal/child_process:483:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:90:21)
Emitted 'error' event on ChildProcess instance at:
    at ChildProcess._handle.onexit (node:internal/child_process:291:12)
    at onErrorNT (node:internal/child_process:483:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:90:21) {
  errno: -4058,
  code: 'ENOENT',
  syscall: 'spawn "C:\\Program',
  path: '"C:\\Program',
  spawnargs: [
    'Files',
    '(x86)\\Notepad++\\notepad++.exe"',
    'C:\\Users\\Jordan.TB\\AppData\\Local\\Temp\\ddb2eddb-8082-4c59-8c79-1d93f4c6b8be.txt'
  ]
}
```

You can see it splits at the first space even within quotes, and it's trying to use the remaining part of the path as arguments.

## Expected Behavior
The way the external-editor parses the command should only split spaces if it's NOT within quotes. So if I were to change process.env['VISUAL'] like this WITHOUT quotes I should expect the same error:
```typescript
process.env['VISUAL'] = 'C:\\Program Files (x86)\\Notepad++\\notepad++.exe';
```

Although if I introduce quotes this should open notepad++ no problem:
```typescript
process.env['VISUAL'] = '"C:\\Program Files (x86)\\Notepad++\\notepad++.exe"';
```

Quotes are definitely still needed either way, but it's still parsing the same way with or without them.

## Proposed Solution
Update the `splitStringBySpace` function to ignore splitting spaces if the space is within quotes

JordanW9232:
Tweak logic for splitting spaces when determining editor to open so that it will not split a space if it's inside quotes
Add unit tests for `determineEditor`

Closes #2040

### DIFF
--- a/packages/external-editor/src/index.ts
+++ b/packages/external-editor/src/index.ts
@@ -9,6 +9,7 @@ import { CreateFileError } from './errors/CreateFileError.ts';
 import { LaunchEditorError } from './errors/LaunchEditorError.ts';
 import { ReadFileError } from './errors/ReadFileError.ts';
 import { RemoveFileError } from './errors/RemoveFileError.ts';
+import { parseEditorCommand } from './parse-editor-command.ts';
 
 export interface IEditorParams {
   args: string[];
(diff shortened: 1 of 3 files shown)


SBoudrias:
Thanks!!

PR #2109, opened 2026-05-07
fix: reset cursor to column 0 after prompt completes (Windows)

### LINKED ISSUE #2108: Windows: cursor not reset to column 0 after prompt completes — garbled output on wrapped lines
## Description

After an `inquirer.prompt()` call completes on Windows terminals, the cursor is left at the column where the rendered answer text ended. If the prompt + answer text wraps past the terminal width, all subsequent output (via `console.log`, `logInfo`, etc.) starts at the wrong horizontal offset instead of column 0.

This only manifests on Windows terminals because `\n` (LF) on Windows moves the cursor down but does **not** reset to column 0. On Linux/macOS terminals, `\n` resets the column as a side effect, masking the issue.

## Reproduction

Run a checkbox prompt where the rendered answer wraps across the terminal:

```js
import inquirer from "inquirer";

const { selected } = await inquirer.prompt([
  {
    type: "checkbox",
    name: "selected",
    message: "Select editors:",
    choices: [
      { name: "Claude Code (detected)", value: "claude", checked: true },
      { name: "Cursor (detected)", value: "cursor", checked: true },
      { name: "VS Code (detected)", value: "vscode", checked: true },
      { name: "opencode (detected)", value: "opencode", checked: true },
    ],
  },
]);

console.log("info: This line should start at column 0");
console.log("info: And so should this one");
```

### Expected output (Linux / macOS)

```
✔ Select editors: Claude Code (detected), Cursor (detected), VS Code (detected), opencode (detected)
info: This line should start at column 0
info: And so should this one
```

### Actual output (Windows)

```
✔ Select editors: Claude Code (detected), Cursor (detected), VS Code (detected), opencode (detected)
                                                                                                      info: This line should start at column 0
                                                                                                                                               info: And so should this one
```

Each line starts at the column where the previous line ended, producing garbled output that shifts further to the right with each line.

## Root cause

In `dist/ui/prompt.js`, the `onClose` handler writes `cursorShow` and closes the readline interface, but never resets the cursor column:

```js
const onClose = () => {
  process.removeListener('exit', onForceClose);
  rl.removeListener('SIGINT', onForceClose);
  rl.setPrompt('');
  rl.output.unmute();
  rl.output.write(cursorShow);
  rl.output.end();
  rl.close();
  // ← missing: cursorTo(rl.output, 0) or equivalent
};
```

## Suggested fix

Add a `cursorTo(rl.output, 0)` call before `rl.output.end()`:

```js
import { cursorTo } from "node:readline";

const onClose = () => {
  process.removeListener('exit', onForceClose);
  rl.removeListener('SIGINT', onForceClose);
  rl.setPrompt('');
  rl.output.unmute();
  rl.output.write(cursorShow);
  cursorTo(rl.output, 0);  // reset cursor to column 0
  rl.output.end();
  rl.close();
};
```

This is a no-op on terminals where `\n` already resets the column, and fixes the garbled output on Windows.

## Workaround

We're currently calling `cursorTo(process.stdout, 0)` from `node:readline` after every `inquirer.prompt()` call in our codebase, guarded by `process.stdout.isTTY`:

```js
import { cursorTo } from "node:readline";

const answers = await inquirer.prompt([...]);
if (process.stdout.isTTY) cursorTo(process.stdout, 0);
```

## Environment

- **inquirer:** 13.4.2
- **OS:** Windows 11 (Windows Terminal, PowerShell 7)
- **Runtime:** Bun 1.3.13 (also reproducible with Node.js)
- **Affects all prompt types:** checkbox, list, input, confirm — they all share the same `onClose` path

rhuanbarreto:
## Summary

On Windows terminals, after a prompt completes, all subsequent console output starts at the wrong horizontal offset — producing garbled output. This happens because `\n` on Windows moves the cursor down **without** resetting to column 0 when the rendered prompt+answer text wraps past the terminal width.

Fixes #2108

### `packages/inquirer/src/ui/prompt.ts` (legacy API)

Added `readline.cursorTo(rl.output, 0)` in the `onClose` handler before `rl.output.end()`:

```diff
 const onClose = () => {
   ...
   rl.output.write(cursorShow);
+  // Reset cursor to column 0 — on Windows, \n does not reset column after line wrap
+  readline.cursorTo(rl.output, 0);
   rl.output.end();
   rl.close();
 };
```

### `packages/core/src/lib/screen-manager.ts` (modern `@inquirer/*` API)

Appended `cursorLeft` (already exported from `@inquirer/ansi` as `\x1b[G`) to the `done()` output before `cursorShow`:

```diff
 let output = cursorDown(this.extraLinesUnderPrompt);
 output += clearContent ? eraseLines(this.height) : '\n';
+output += cursorLeft;
 output += cursorShow;
```

### `packages/core/core.test.ts`

Updated the `'allow cleaning the prompt after completion'` assertion to include the new `cursorLeft` in the expected raw output.

## Why this is safe

- `cursorLeft` (`\x1b[G`) is a no-op when the cursor is already at column 0, which is the case on Linux/macOS where `\n` resets the column automatically
- The escape sequence is standard ANSI CSI and supported by all modern terminals
- Affects all prompt types (checkbox, list, input, confirm, etc.) since they all share the same close/done path

## Reproduction

See #2108 for the full reproduction case and screenshots. In short: any prompt whose rendered answer wraps past the terminal width on Windows produces garbled output for all subsequent `console.log` calls.

### DIFF
--- a/packages/core/core.test.ts
+++ b/packages/core/core.test.ts
@@ -23,7 +23,7 @@ import {
   makeTheme,
   type Status,
 } from './src/index.ts';
-import { cursorShow, eraseLines } from '@inquirer/ansi';
+import { cursorLeft, cursorShow, eraseLines } from '@inquirer/ansi';
 
 describe('createPrompt()', () => {
   it('onKeypress: allow to implement custom behavior on keypress', async () => {
@@ -489,7 +489,7 @@ describe('createPrompt()', () => {
     events.keypress('enter');
 
     await expect(answer).resolves.toEqual('done');
-    expect(getScreen({ raw: true })).toEqual(eraseLines(1) + cursorShow);
+    expect(getScreen({ raw: true })).toEqual(eraseLines(1) + cursorLeft + cursorShow);
   });
 
   it('clear timeout when force closing', { timeout: 1000 }, async () => {
--- a/packages/core/src/lib/screen-manager.ts
+++ b/packages/core/src/lib/screen-manager.ts
@@ -1,6 +1,13 @@
 import { stripVTControlCharacters } from 'node:util';
 import { breakLines, readlineWidth } from './utils.ts';
-import { cursorDown, cursorUp, cursorTo, cursorShow, eraseLines } from '@inquirer/ansi';
+import {
+  cursorDown,
+  cursorLeft,
+  cursorUp,
+  cursorTo,
+  cursorShow,
+  eraseLines,
+} from '@inquirer/ansi';
 import type { InquirerReadline } from '@inquirer/type';
 
 const height = (content: string): number => content.split('\n').length;
@@ -97,6 +104,11 @@ export default class ScreenManager {
 
     let output = cursorDown(this.extraLinesUnderPrompt);
     output += clearContent ? eraseLines(this.height) : '\n';
+    // Reset cursor to column 0. On Windows terminals, \n moves the cursor
+    // down without resetting the column when the rendered prompt+answer
+    // wraps past the terminal width. Without this, all subsequent output
+    // starts at the wrong horizontal offset.
+    output += cursorLeft;
     output += cursorShow;
     this.write(output);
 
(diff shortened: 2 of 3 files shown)


SBoudrias (approved):
Thanks for sending a fix!

PR #2120, opened 2026-05-14
feat(@inquirer/checkbox): add exclusive choice option

swseverance:
I'd like to introduce this feature PR by illustrating how the checkbox module is currently
being used in a very popular library, @angular/cli. When creating a new angular workspace
the user is prompted to select AI tools to add to the workspace. However, it currently allows
for illogical selections such as having None, Claude, and Cursor selected at the same time.

Screenshot 2026-05-14 at 7 11 05 PM

This PR introduces a new `exclusive: boolean` property on checkbox `Choice` objects. What this
would allow a library like @angular/cli to do is setup their options as:
```typescript
[
  { name: 'None', value: 'none', exclusive: true },
  { name: 'Claude', value: 'claude' },
  { name: 'Cursor', value: 'cursor' },
]
```

Selecting "Claude" would automatically deselect "None". Selecting "None" would automatically
deselect "Claude" and "Cursor" if they were selected.

**Edge cases**

If you're open to the functionality added in this PR one thing we'll need to address is how to handle exclusive options in relation to the "select all" and "invert selection" features that inquirer already provides. I would think that users would not typically want the "select all" and "invert selection" features alongside usage of exclusive options but it's probably something we'd need to address at some point.

### DIFF
--- a/packages/checkbox/checkbox.test.ts
+++ b/packages/checkbox/checkbox.test.ts
@@ -1480,6 +1480,120 @@ describe('checkbox prompt', () => {
       await expect(answer).resolves.toEqual([]);
     });
   });
+
+  describe('exclusive choices', () => {
+    const exclusiveChoices = [
+      { value: 'none', name: 'None', exclusive: true },
+      { value: 'apple', name: 'Apple' },
+      { value: 'orange', name: 'Orange' },
+    ];
+
+    it('selecting an exclusive choice deselects all other choices', async () => {
+      const { answer, events } = await render(checkbox, {
+        message: 'Select fruits',
+        choices: exclusiveChoices,
+      });
+
+      // Select apple and orange first
+      events.keypress('down');
+      events.keypress('space');
+      events.keypress('down');
+      events.keypress('space');
+
+      // Navigate to None (exclusive) and select it
+      events.keypress('up');
+      events.keypress('up');
+      events.keypress('space');
+
+      events.keypress('enter');
+      await expect(answer).resolves.toEqual(['none']);
+    });
+
+    it('selecting a non-exclusive choice deselects all exclusive choices', async () => {
+      const { answer, events } = await render(checkbox, {
+        message: 'Select fruits',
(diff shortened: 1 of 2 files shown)


swseverance:
I've given it some more thought and perhaps this is not worth the added complexity to the library

SBoudrias:
Thanks for opening a PR nonetheless.

I agree this might not be a great UX patterns and/or would be a better fit for a custom prompt.

In the specific example of Angular-CLI, it makes me wonder why "None" needs to be an option. It's possible to leave the list empty if the user don't want any option. Maybe the way the prompt is worded was leading to confusion rather than the lack of a "None" option?

swseverance:
@SBoudrias perhaps their thought process was "let's be explicit when letting the user know it is ok to not select any AI agents at all".

If a user sees the following UI alone they may think to themselves "Am i being required to select one of these options?". Perhaps they are unaware that they can simply hit "Enter" and not select any agents (provided `required: true` is not set)
```
? Select AI agents
  ◯ Claude
  ◯ Cursor
  ◯ Gemini
```

However, if a user sees the following UI it is very clear that it's perfectly acceptable to select no AI agents.
```
? Select AI agents
  ◉ None
  ◯ Claude
  ◯ Cursor
  ◯ Gemini
```

The main reason I abandoned the `exclusive: boolean` approach was that it became very difficult to reason about what should happen if the user hits "a" or "i" to select or invert selection. How should we treat the `exclusive` options when they do so? The edge cases started building up and I abandoned the approach.

If allowing a user to explicitly select "None" (as opposed to just leaving all options blank and hitting enter) would be a valuable contribution to the library I can definitely explore that. For example, perhaps we could introduce a new property to the `CheckboxConfig` like `allowNone: { name: string; value: Value }`.  In the below video I demonstrate how selecting an option deselects "None". How selecting "None" deselects other options. Select all, and invert selection.

https://github.com/user-attachments/assets/23fe9550-7cf6-45f9-9ca1-ec05c0414a4d

PR #2136, opened 2026-05-24
refactor(inquirer): remove rxjs runtime dependency

SBoudrias:
## Summary

Remove `rxjs` from the `inquirer` package runtime and public type surface while preserving the legacy reactive API shape.

- replace the internal RxJS prompt pipeline with explicit async sequencing
- add a lightweight local Observable implementation for `ui.process`
- accept structural Observable-like question sources instead of importing RxJS types
- keep `rxjs` as a dev dependency for compatibility tests and examples
- update reactive-interface docs to describe supported usage without implementation details

## Validation

- `yarn install --immutable`
- `yarn vitest --run packages/inquirer/inquirer.test.ts`
- `yarn workspace inquirer tsc --pretty false`
- `yarn pretest` passes, with an existing warning in `packages/input/src/index.ts`

## Notes

The full `yarn test` run, and therefore the local pre-push hook, currently fails in unrelated prompt package tests (`input`, `number`, `expand`, `rawlist`, `search`) with keypress snapshot mismatches and timeouts. The branch was pushed with `--no-verify` after the focused checks above passed.

### DIFF
--- a/packages/inquirer/inquirer.test.ts
+++ b/packages/inquirer/inquirer.test.ts
@@ -9,7 +9,7 @@ import stream from 'node:stream';
 import tty from 'node:tty';
 import readline from 'node:readline';
 import { vi, expect, beforeEach, afterEach, describe, it, expectTypeOf } from 'vitest';
-import { of } from 'rxjs';
+import { from as observableFrom, map, of, Subject } from 'rxjs';
 import { AbortPromptError, createPrompt } from '@inquirer/core';
 import type { InquirerReadline } from '@inquirer/type';
 import inquirer from './src/index.ts';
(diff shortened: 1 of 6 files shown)


packages/inquirer/src/ui/prompt.ts
```diff
@@ -219,58 +216,48 @@ export default class PromptsRunner<A extends Answers> {
 
   async run(questions: PromptSession<A>, answers?: Partial<A>): Promise<A> {
     this.abortController = new AbortController();
+    this.processController = createObservableController<AnswerEvent>();
```
SBoudrias: `const processController` given it's not used in any other methods.

packages/inquirer/src/types.ts
```diff
@@ -12,7 +12,7 @@ import {
   select,
 } from '@inquirer/prompts';
 import type { Context, DistributiveMerge, Prettify } from '@inquirer/type';
-import { Observable } from 'rxjs';
+import type { ObservableLike } from './utils/observable.ts';
```
SBoudrias: Why not just keep `Observable` as the type name?

packages/inquirer/inquirer.test.ts
```diff
@@ -638,6 +683,37 @@ describe('inquirer.prompt(...)', () => {
...
+        type: 'stub',
+        name: 'name',
+        message: 'message',
+        answer: 'doe',
+      },
+    ]);
+    const process = promise.ui.process as ObservableInput<{
+      name: string;
+      answer: unknown;
+    }>;
```
SBoudrias: Can we make the types compatible so we don't need to use an unsafe `as` type casting?

PR #2141, opened 2026-05-26
fix: validate package engine compatibility

### LINKED ISSUE #2140: mute-stream version bump issue
PR #2119 bumped mute-stream to 4.0.0, which introduced a very aggressive `engines` field [`"node": "^22.22.2 || ^24.15.0 || >=26.0.0"`](https://github.com/npm/mute-stream/blame/e79d27e7553f563a1417b40596e30561dea440b4/package.json#L38).

This means that when installing `@inquirer/core` 11.2.0 you now get a EBADENGINE warning using `npm` when on a slightly outdated v24 LTS for example:

```
npm warn EBADENGINE Unsupported engine {
npm warn EBADENGINE   package: 'mute-stream@4.0.0',
npm warn EBADENGINE   required: { node: '^22.22.2 || ^24.15.0 || >=26.0.0' },
npm warn EBADENGINE   current: { node: 'v24.14.0', npm: '11.12.0' }
npm warn EBADENGINE }
```

I'm not sure what the remediation should be here. Is npm usually this aggressive with the engines fields in their first-party packages?

Please consult if I need to open an issue over there as well.

SBoudrias:
## Summary

Fixes #2140 by aligning the published Node engine range with the runtime dependencies we ship.

- pin direct `mute-stream` usage back to `^3.0.0`
- update workspace `engines.node` to the range supported by both `mute-stream@3` and the repo's `styleText` usage: `>=23.5.0 || ^22.13.0 || ^20.17.0`
- add `validate-package-engines` and run it in CI to catch package/dependency engine mismatches
- make setup package resolution fall back when a dependency does not export `package.json`
- cover the setup fallback, peer dependency hoisting, and engine validation behavior with focused tests

## Why

`mute-stream@4` advertises a narrower Node range than Inquirer packages currently publish, so consumers can install an incompatible runtime dependency while satisfying Inquirer's engines. `mute-stream@3` has the minimum range we need, but the package manifests still need to state the real intersection explicitly so future dependency updates cannot widen the published support by accident.

## Validation

- `yarn install`
- `yarn setup`
- `yarn check:engines`
- `yarn vitest --run internals/setup-packages`
- `yarn tsc`
- `yarn oxlint .`
- `yarn oxfmt --check .`
- `yarn eslint .` (passes with an existing warning in `packages/input/src/index.ts`)
- `git diff --check`

`yarn test` still fails in 21 prompt interaction tests around backspace/input state in `expand`, `rawlist`, `number`, `search`, and `input`. This is the same failure pattern observed before pushing and is unrelated to this setup/engine change.


KirillTregubov:
This is looking great, thanks for swiftly addressing my issue!

PR #2166, opened 2026-06-17
fix(@inquirer/number): avoid floating-point step errors

JSap0914:
## Bug

`isStepOf` scaled values by a fixed `1e6` before checking the remainder. IEEE-754 multiplication can still produce a fractional result:

```js
1.001 * 1e6; // 1000999.9999999999
```

As a result, the number prompt rejected valid decimal steps such as `1.001` with `step: 0.001`.

## Fix

Convert the supplied value, step, and finite minimum from their canonical decimal representations into integer significands. Scale them to their shared decimal exponent and perform the remainder check with `bigint` arithmetic.

This avoids floating-point remainder errors without imposing a fixed six-decimal precision limit. It also preserves correct behavior for:

- steps smaller than `1e-6`
- decimal `min` offsets
- near-multiples that must remain invalid
- values represented with scientific notation

## Verification

```sh
yarn pretest
yarn workspace @inquirer/number tsc
yarn vitest --run packages/number/src/is-step-of.test.ts
yarn vitest --run packages/number/number.test.ts -t 'decimal steps'
```

The full local suite also ran during the push hook: 360 tests passed and 21 unrelated synthetic-backspace tests failed across existing prompt suites in this terminal environment. GitHub CI will rerun the complete supported matrix.

> This fix was developed with AI assistance.

### DIFF
--- a/packages/number/number.test.ts
+++ b/packages/number/number.test.ts
@@ -366,4 +366,18 @@ describe('number prompt', () => {
     await expect(answer).resolves.toEqual(10.01);
     expect(getScreen()).toMatchInlineSnapshot(`"✔ Enter a decimal number 10.01"`);
   });
+
+  it('accepts decimal steps without floating-point precision errors', async () => {
+    const { answer, events, getScreen } = await render(number, {
+      message: 'Enter a value',
+      step: 0.001,
+    });
+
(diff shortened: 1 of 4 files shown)


SBoudrias (approved):
Thanks!

PR #2181, opened 2026-06-27
fix(@inquirer/core): fall back for readline native _ttyWrite on Node.js v25+

### LINKED ISSUE #2180: bug(@inquirer/core): keypress events emitted without sequence field do not update rl.line/rl.cursor in Node.js v25 (breaks backspace/arrow keys in @inquirer/testing)
## Bug

In Node.js v25, readline's native `_ttyWrite` implementation does not process keypress events that are emitted programmatically without a `sequence` field. This affects `@inquirer/testing`'s `events.keypress()` helper, which emits keypress events directly on the input stream without including a sequence string.

The result: `rl.line` and `rl.cursor` are not updated when `events.keypress('backspace')`, `events.keypress('left')`, etc. are called in tests.

## Affected tests (failing on Node.js v25 before this fix)

- `@inquirer/number`: 8 tests fail — backspace/arrow after validation errors
- `@inquirer/input`: 6 tests fail — same root cause
- `@inquirer/rawlist`: 1 test fails — `displays description when choice is selected`

## Root cause

When `events.keypress('backspace')` is called:
```js
input.emit('keypress', null, { name: 'backspace' });
```

readline's internal `onkeypress` handler calls `_ttyWrite(null, { name: 'backspace' })`. In Node.js v25 the native `_ttyWrite` requires a non-empty `sequence` field to process the key — with `null` or no sequence it is a no-op.

## Reproduction

Run the existing test suite on Node.js v25:
```
yarn vitest run packages/number/number.test.ts
# → 8 failures involving backspace / cursor movement
```

## Environment

- Node.js v25.6.1
- `mute-stream` 4.0.0

JSap0914:
## Bug

Node.js v25 moved readline's `_ttyWrite` to a native C++ implementation. Unlike the previous JS implementation, the native version does **not** process keypress events emitted programmatically without a `sequence` field. This breaks `@inquirer/testing`'s `events.keypress()` helper, which emits keypress events directly on the input stream (e.g. `input.emit('keypress', null, { name: 'backspace' })`) without a sequence string.

**Closes #2180**

## Impact before this fix

9 test failures across three packages:

| Package | Failing tests |
|---|---|
| `@inquirer/number` | 8 — including backspace/arrow after validation errors, cursor-in-middle editing |
| `@inquirer/input` | 6 — same root cause |
| `@inquirer/rawlist` | 1 — `displays description when choice is selected` |

## Fix

Added two listeners to `rl.input` in `create-prompt.ts` (inserted between readline's own listener and the prompt's listener via `prependListener` + `on`):

1. **`prependListener`**: records `rl.cursor` **before** readline's own keypress handler runs.
2. **Post-readline listener**: fires **after** readline's handler; if `rl.cursor` hasn't changed (readline's native handler was a no-op), applies the appropriate internal readline method (`_deleteLeft`, `_moveCursor`, `_deleteWordLeft`, etc.) as a fallback.

Additionally, for **non-TTY inputs** (test environments): intercept `input.write` so that printable character insertions use `_insertString` (which respects `rl.cursor`) rather than the native insertion path that ignores the JS cursor position.

Both listeners are registered in `cleanups` and removed when the prompt closes.

## Verification

Full suite:

> AI-assisted: the investigation and implementation were done with AI assistance.

### DIFF
--- a/packages/core/src/lib/create-prompt.ts
+++ b/packages/core/src/lib/create-prompt.ts
@@ -61,6 +61,122 @@ export function createPrompt<Value, Config>(
       output,
     }) as unknown as InquirerReadline;
 
+    // Fix for Node.js v25+ where the native readline `_ttyWrite` implementation
+    // does not process special keypress events (backspace, delete, arrow keys)
+    // when they are emitted programmatically without a `sequence` field — as is
+    // done by @inquirer/testing's `events.keypress()` helper.
+    //
+    // Strategy: use prependListener to record the cursor position *before*
+    // readline's own keypress handler runs, then use a second listener that fires
+    // *after* readline's handler to apply the missing edit if readline's handler
+    // left rl.line / rl.cursor unchanged.
+    let cursorBeforeKey = rl.cursor;
+    const saveCursorBeforeKey = () => {
+      cursorBeforeKey = rl.cursor;
+    };
+    rl.input.prependListener('keypress', saveCursorBeforeKey);
+
+    const fixupKeypress = (
+      _s: string | null | undefined,
+      key: { name?: string; ctrl?: boolean; meta?: boolean } | undefined,
+    ) => {
+      // If readline's native handler processed the key, rl.cursor would have
+      // changed (for cursor-movement / destructive keys). We only need to
+      // intervene when the cursor is unchanged.
+      if (rl.cursor !== cursorBeforeKey) return;
+
+      const rlAny = rl as unknown as Record<string, ((...args: unknown[]) => void) | undefined>;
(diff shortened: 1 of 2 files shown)


JSap0914:
Withdrawing this. While it does make the suite pass on Node v25, the approach relies on private/undocumented readline internals (_deleteLeft, _moveCursor, _insertString) and intercepts input.write in core create-prompt.ts, which is too invasive and fragile to maintain across Node versions. I'd rather not push that into the core path. I've left the analysis in #2180 in case it's useful for a maintainer-preferred approach. Apologies for the noise.

Issue #2110, opened 2026-05-08
@inquirer/prompts input can't accept default: undefined (typing inconsistency)


LorisRue:
Using: `@inquirer/prompts": "^8.4.2`

Even though default is an optional property in the input options, it can't be set to undefined. It must either be a string or left out entirely, otherwise the TypeScript Compiler complains. This is inconsistent with how`number` from `@inquirer/prompts` works, since when hovering over it, the type is shown as `number | undefined`

This behavior forces me to write things like this when `defaultValue: string | undefined`
```typescript
const answer = await input({
    message,
    required,
    ...(defaultValue !== undefined ? { default: defaultValue } : {}),
  });
```
While, for number, I can just do this:
```typescript
const answer = await number({
    message,
    required,
    default: defaultValue,
  });
```

Otherwise I would get this TypeScript error:
```
Argument of type '{ message: string; required: boolean; default: string | undefined; validate: (value: string) => true | "This value is required."; }' is not assignable to parameter of type 'InputConfig' with 'exactOptionalPropertyTypes: true'. Consider adding 'undefined' to the types of the target's properties.
  Types of property 'default' are incompatible.
    Type 'string | undefined' is not assignable to type 'string'.
      Type 'undefined' is not assignable to type 'string'.
```

I believe this could be solved pretty easily by changing the InputConfig type in [packages/input/src/index.ts](https://github.com/SBoudrias/Inquirer.js/blob/main/packages/input/src/index.ts) to have `default?: string | undefined`

PR #2182, opened 2026-06-29
fix: accept explicit undefined as default in number, confirm, editor

mahirhir:
## Summary

#2111 widened `InputConfig['default']` to `string | undefined` so the option accepts an explicitly-undefined value under `exactOptionalPropertyTypes: true` (the inconsistency reported in #2110). The other prompts that expose a scalar `default` still declare it as a bare optional, so they reject an explicit `undefined` the way `@inquirer/input` did before #2111:

- `@inquirer/number`: `default?: number`
- `@inquirer/confirm`: `default?: boolean`
- `@inquirer/editor`: `default?: string`

This applies the same widening to those three, so `default: value` type-checks when `value` is `T | undefined`. Each gets a test mirroring the one added in #2111.

The choice-based prompts (`select`, `rawlist`, `search`, `expand`) have the same gap on their `default` option. I kept this PR to the scalar-default prompts that match `@inquirer/input`, and am happy to follow up on the rest.

## Test plan

- [x] `yarn tsc`
- [x] `yarn vitest --run packages/number packages/confirm packages/editor`
- [x] `yarn oxlint` / `yarn eslint`

### DIFF
--- a/packages/confirm/confirm.test.ts
+++ b/packages/confirm/confirm.test.ts
@@ -95,6 +95,18 @@ describe('confirm prompt', () => {
     expect(getScreen()).toMatchInlineSnapshot(`"✔ Do you want to proceed? No"`);
   });
 
+  it('accepts explicit undefined as default', async () => {
+    const defaultValue: boolean | undefined = undefined;
+    const { answer, events } = await render(confirm, {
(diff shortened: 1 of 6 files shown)


SBoudrias (approved):
Thanks!

PR #2191, opened 2026-07-07
fix(external-editor): set `shell` to `true`

avivkeller:
On Windows devices, `shell` must be `true` to execute certain commands via `node:child_process`. For instance, `child_process.spawnSync("code")` will fail and `child_process.spawnSync("code", { shell: true })` will not.

This updates the external editor package to set `shell: true` rather than `shell: false`.

### DIFF
--- a/packages/external-editor/src/index.ts
+++ b/packages/external-editor/src/index.ts
@@ -109,7 +109,7 @@ export class ExternalEditor {
     const promise = new Promise<void>((resolve, reject) => {
       try {
         const editorProcess = spawn(this.editor.bin, this.editorArgs(), {
-          shell: false,
+          shell: true,
           stdio: 'inherit',
         });
         editorProcess.once('error', (launchError) => {


SBoudrias:
Thanks for the report. The Windows `.cmd` shim issue is real, but I don't think we can switch this to `shell: true` globally.

`shell: true` means the editor command is interpreted by `cmd.exe`/`sh` instead of being launched directly. That changes error and exit-code behavior, and it reintroduces shell parsing for editor args/temp paths. CI is already showing that: `runAsync()` stops rejecting for a missing editor, and a non-zero editor exit can be reported as success.

For your local setup, the safer workaround is to point `VISUAL`/`EDITOR` at the real executable instead of the `.cmd` shim. For VS Code on Windows, for example:

```powershell
setx VISUAL '"C:\Program Files\Microsoft VS Code\Code.exe" --wait'
```

or for the per-user install:

```powershell
setx VISUAL '"C:\Users\<you>\AppData\Local\Programs\Microsoft VS Code\Code.exe" --wait'
```

That avoids `code.cmd` and lets Inquirer keep spawning the editor directly.

A README note about preferring the real Windows editor executable over `.cmd`/`.bat` shims would be worthwhile. A library-side fix would need to target Windows command shims specifically, not enable shell mode for every editor launch.

SBoudrias:
Documenting this more cleanly in our README here https://github.com/SBoudrias/Inquirer.js/pull/2198

PR #2223, opened 2026-08-06
fix(@inquirer/i18n): accept the yes/no answers each locale hints at

Jaybhade:
## The bug

`@inquirer/i18n` translates the confirm hint but never taught the prompt to read the
answer back. `@inquirer/confirm` only matches `/^(y|yes)/i` and `/^(n|no)/i`, so every
locale whose words for yes/no don't start with `y`/`n` silently mis-reads the user.

Worst case, the prompt returns the opposite of what was typed:

```js
import { confirm } from '@inquirer/i18n/es';

await confirm({ message: '¿Continuar?', default: false });
// renders:  ? ¿Continuar? (s/N)      <- tells the user to press "s" for sí
// user types "s", presses enter
// → false
```

`s` matches neither pattern, so the answer falls through to `default`. Same for French
`o`/`oui` behind the `(o/N)` hint, and Portuguese `s`/`sim` behind `(s/N)`.

Chinese is worse — 是 and 否 are unparseable in either direction, so `@inquirer/i18n/zh`
always resolves to its default no matter what the user answers:

```js
import { confirm } from '@inquirer/i18n/zh';

await confirm({ message: '确认?', default: false }); // user types 是 → false
await confirm({ message: '确认?', default: true }); //  user types 否 → true
```

The custom-locale example in `packages/i18n/README.md` has the same hole: it hints
`J/n` for German, and `j` was never accepted.

## The fix

`@inquirer/confirm` now derives its accepted answers from the hint it displays. The hint
is exposed as a new `confirmHint` theme style (following the existing `SelectTheme` /
`CheckboxTheme` pattern):

```ts
theme: {
  style: {
    confirmHint: (defaultValue) => (defaultValue === false ? 'y/N' : 'Y/n'),
  },
},
```

The first token of the hint is matched as "yes", the second as "no" — prefix-based and
case-insensitive, mirroring how `yes`/`no` already behave. The English `y`/`yes` and
`n`/`no` remain accepted on top, so existing behaviour is byte-for-byte unchanged and the
y/n shortcuts every locale inherits today keep working.

`@inquirer/i18n` then overrides `confirmHint` with each locale's own hint, and the old
`defaultAnswer` translation hack in `create.ts` disappears:

```ts
confirmHint: (defaultValue) =>
  defaultValue === false ? locale.confirm.hintNo : locale.confirm.hintYes,
```

Because the hint drives both the display and the parser, they can never drift apart, and
every custom locale is fixed retroactively — the German example in the README now accepts
`j`/`n` with no extra configuration. No new prompt option is exposed.

## Verification

`yarn pretest && yarn test` — green. `yarn vitest --run packages` reports 405 passing,
1 skipped; the CJS/ESM integration suites pass too.

New tests, all of which fail on `main`:

- `packages/confirm/confirm.test.ts` — the accepted answers follow the theme's
  `confirmHint`, and inputs matching neither the hint nor y/n fall back on the default.
- `packages/i18n/test/confirm.test.ts` — now globs over every locale module (so new
  locales are covered automatically): each locale shows its localized hint, accepts its
  localized yes/no labels, and still resolves the English `y`/`n` shortcuts.

Also driven through a real pty, since the unit tests use a simulated one:

```
? ¿Continuar? (s/N) s   →  ✔ ¿Continuar? Sí     RESULT=true
? 确认? (是/否) 是       →  ✔ 确认? 是            RESULT=true    (default: false)
? 确认? (是/否) 否       →  ✔ 确认? 否            RESULT=false   (default: false)
```

### DIFF
--- a/packages/confirm/README.md
+++ b/packages/confirm/README.md
@@ -78,13 +78,36 @@ type Theme = {
     interval: number;
     frames: string[];
   };
+  keywords: {
+    yes: string;
+    no: string;
+  };
   style: {
     answer: (text: string) => string;
     message: (text: string, status: 'idle' | 'done' | 'loading') => string;
     defaultAnswer: (text: string) => string;
+    confirmDefault: (text: string) => string;
   };
 };
 ```
+
+The `keywords` property defines the words accepted as "yes" and "no". Matching is
+prefix-based and case-insensitive, and the first character of each word is shown in
+the hint (e.g. `Y/n`). The matched word is also displayed once the prompt is answered.
+Overriding `keywords` is how `@inquirer/i18n` localizes the prompt:
+
+```ts
+confirm({
+  message: '¿Continuar?',
+  theme: {
+    keywords: { yes: 'Sí', no: 'No' },
+  },
+});
+```
+
+The `confirmDefault` style marks the character representing the default answer in the
+hint. By default it uppercases the character (`Y/n`); for scripts without case
+(e.g. Chinese) it highlights the character with a color instead.
 
 # License
 
--- a/packages/confirm/confirm.test.ts
+++ b/packages/confirm/confirm.test.ts
@@ -1,6 +1,11 @@
-import { describe, it, expect } from 'vitest';
+import { describe, it, expect, vi, afterEach } from 'vitest';
+import { styleText } from 'node:util';
 import { render } from '@inquirer/testing';
 import confirm from './src/index.ts';
+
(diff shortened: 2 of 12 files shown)


SBoudrias (approved):

SBoudrias:
Sorry for the slow merge, I went back and forth on the right implementation for this. But I think this last state is where it needs to be to simplify the implementation.

PR #2236, opened 2026-08-19
docs: add inquirer-grouped-checkbox to community prompts

patik:
Adds [inquirer-grouped-checkbox](https://github.com/patik/inquirer-grouped-checkbox) to the community prompts list.

It is a checkbox prompt whose choices are organized into labeled groups. The group headers are navigable and selectable, so pressing space on one toggles every item in that group, and each header carries its own running total. If you turn on `searchable`, typing filters across all of the groups at once.

The package is built on `@inquirer/core` v12, is published to npm, and ships types for both ESM and CJS. The sample output in the diff is copied from an actual run rather than written by hand.

### DIFF
--- a/packages/prompts/README.md
+++ b/packages/prompts/README.md
@@ -581,6 +581,22 @@ Navigate a tree of choices, expanding and collapsing branches with the arrow key
 ----------------
 ```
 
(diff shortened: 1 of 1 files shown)


SBoudrias:
Thanks!

PR #2237, opened 2026-08-19
feat(@inquirer/password): add ctrl+space toggle to reveal password

### LINKED ISSUE #1330: Feature request idea(input): toggle the transform function when CTRL+SPACE is clicked

manrods:
Closes #1330. Supersedes #1334 (stalled/outdated — different codebase since then, so this is a fresh implementation rather than a rebase of it).

## What this does

Adds an opt-in `allowShowPassword` config option to `@inquirer/password`. When enabled, pressing `ctrl+space` temporarily reveals the typed value in plaintext; pressing it again re-masks it. A themed help tip (`(ctrl+space to show)` / `(ctrl+space to hide)`) is appended to the prompt so the shortcut is discoverable without reading docs. Default is `false`, so existing behavior/snapshots are unchanged for everyone who doesn't opt in.

## Addressing the three concerns from the issue thread

1. **Discoverability** — reuses the existing help-tip mechanism (the same one that already renders `[input is masked]`) to show a live, state-aware hint (`show` vs `hide`) next to the prompt whenever the feature is enabled.
2. **Common/standard shortcut** — I looked for prior art before picking one: GUI password fields (browsers, password managers, OS credential prompts) universally use a mouse-driven eye icon, not a keyboard shortcut; TUI/CLI prompt libraries I checked (bubbletea/huh, survey, promptui, gocui) don't expose a runtime reveal toggle at all — echo mode is fixed at construction. So there doesn't appear to be an existing convention to defer to. I went with `ctrl+space` because it's mnemonic (space already means "toggle" elsewhere, e.g. checkbox), doesn't collide with the shell's `ctrl+r` reverse-search or the emacs-keybinding set this repo already supports, and matches what #1334 originally proposed — happy to bikeshed this further if you have a preference.
3. **Opt-in** — off by default via `allowShowPassword`; when unset, the keypress handler doesn't register the toggle at all and no hint is rendered, so behavior for existing consumers is 100% unchanged.

## Caveat worth flagging

Some terminal emulators/OSes intercept `ctrl+space` for IME/input-source switching (notably on macOS and some Linux DEs) before it reaches the process. This is documented in the README as a best-effort limitation — masking still works normally, the reveal shortcut just won't fire in that terminal.

## What changed

- `packages/password/src/index.ts` — `allowShowPassword` config, `revealed` state, `ctrl+space` handling via `@inquirer/core`'s `isSpaceKey`, render changes for both the `mask` and no-`mask` paths, no plaintext/tip leak once the prompt is `done`.
- `packages/password/README.md` — documents the new option, the theming hook (`showHideTip`), and the terminal-compatibility caveat.
- `packages/password/password.test.ts` — new cases: off-by-default no-op, reveal/hide toggling for both `mask: true` and no-`mask`, no leak past submission, and no interference with the existing `validate` flow.

## Test plan

- [x] `yarn vitest --run packages/password` — 10/10 passing
- [x] `yarn pretest` (tsc + oxlint + oxfmt + eslint) — clean
- [x] `yarn test` (full unit + integration suite) — 394 passing / 1 skipped, 0 failing
- [ ] Manual verification of `ctrl+space` delivery across real terminals (xterm, iTerm2/Terminal.app, Windows Terminal) — not yet done on my end; flagging in case you or another reviewer can confirm on a platform I don't have handy.

### DIFF
--- a/packages/password/src/index.ts
+++ b/packages/password/src/index.ts
@@ -10,33 +10,41 @@ import {
 } from '@inquirer/core';
 import { cursorHide } from '@inquirer/ansi';
 import type { PartialDeep } from '@inquirer/type';
+import { styleText } from 'node:util';
 
 type PasswordTheme = {
   style: {
     maskedText: string;
+    keysHelpTip: (keys: [key: string, action: string][]) => string | undefined;
   };
 };
 
 const passwordTheme: PasswordTheme = {
   style: {
     maskedText: '[input is masked]',
+    keysHelpTip: (keys: [string, string][]) =>
+      keys
+        .map(([key, action]) => `${styleText('bold', key)} ${styleText('dim', action)}`)
+        .join(styleText('dim', ' • ')),
   },
 };
 
 type PasswordConfig = {
   message: string;
   mask?: boolean | string;
+  toggleMask?: boolean;
   validate?: (value: string) => boolean | string | Promise<string | boolean>;
   theme?: PartialDeep<Theme<PasswordTheme>>;
 };
 
 export default createPrompt<string, PasswordConfig>((config, done) => {
-  const { validate = () => true } = config;
+  const { toggleMask = true, validate = () => true } = config;
   const theme = makeTheme<PasswordTheme>(passwordTheme, config.theme);
 
   const [status, setStatus] = useState<Status>('idle');
   const [errorMsg, setError] = useState<string>();
   const [value, setValue] = useState<string>('');
+  const [revealed, setRevealed] = useState(false);
 
   const prefix = usePrefix({ status, theme });
 
@@ -61,6 +69,8 @@ export default createPrompt<string, PasswordConfig>((config, done) => {
         setError(isValid || 'You must provide a valid value');
         setStatus('idle');
       }
+    } else if (toggleMask && key.ctrl && key.name === 't') {
+      setRevealed((prev) => !prev);
     } else {
       setValue(rl.line);
       setError(undefined);
@@ -69,23 +79,33 @@ export default createPrompt<string, PasswordConfig>((config, done) => {
 
   const message = theme.style.message(config.message, status);
 
+  const showPlaintext = toggleMask && revealed && status === 'idle';
+
   let formattedValue = '';
-  let helpTip;
-  if (config.mask) {
+  if (showPlaintext) {
+    formattedValue = value;
+  } else if (config.mask) {
     const maskChar = typeof config.mask === 'string' ? config.mask : '*';
     formattedValue = maskChar.repeat(value.length);
   } else if (status !== 'done') {
-    helpTip = `${theme.style.help(theme.style.maskedText)}${cursorHide}`;
+    formattedValue = theme.style.help(theme.style.maskedText);
   }
 
   if (status === 'done') {
     formattedValue = theme.style.answer(formattedValue);
+  } else if (!config.mask) {
+    formattedValue += cursorHide;
   }
 
-  let error = '';
-  if (errorMsg) {
-    error = theme.style.error(errorMsg);
-  }
+  const content = [prefix, message, formattedValue].filter(Boolean).join(' ');
+  const bottomContent = [
+    errorMsg ? theme.style.error(errorMsg) : '',
+    toggleMask && status === 'idle'
+      ? theme.style.keysHelpTip([['ctrl+t', 'toggle visibility']])
+      : '',
+  ]
+    .filter(Boolean)
+    .join('\n');
 
-  return [[prefix, message, config.mask ? formattedValue : helpTip].join(' '), error];
+  return [content, bottomContent];
 });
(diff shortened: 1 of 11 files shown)


packages/password/src/index.ts
```diff
@@ -14,18 +15,22 @@ import type { PartialDeep } from '@inquirer/type';
...
     maskedText: string;
+    showHideTip: (visible: boolean) => string;
   };
 };
 
 const passwordTheme: PasswordTheme = {
   style: {
     maskedText: '[input is masked]',
+    showHideTip: (visible: boolean) =>
+      visible ? '(ctrl+space to hide)' : '(ctrl+space to show)',
```
SBoudrias: We should display keyboard commands in the bottom line and style it like we do other prompts. See `checkbox` for a more complex example.

packages/password/src/index.ts
```diff
@@ -14,18 +15,22 @@ import type { PartialDeep } from '@inquirer/type';
...
   },
 };
 
 type PasswordConfig = {
   message: string;
   mask?: boolean | string;
+  allowShowPassword?: boolean;
```
SBoudrias: We should have a default for this value; IMO we can safely default it to `true`.

packages/password/src/index.ts
```diff
@@ -69,23 +80,31 @@ export default createPrompt<string, PasswordConfig>((config, done) => {
...
   }
 
-  let error = '';
-  if (errorMsg) {
-    error = theme.style.error(errorMsg);
-  }
+  const content = [prefix, message, formattedValue].filter(Boolean).join(' ');
+  const helpLine =
+    allowShowPassword && status === 'idle'
+      ? theme.style.keysHelpTip([['ctrl+space', showPlaintext ? 'hide' : 'show']])
```
SBoudrias: We'll need the labels to be i18n-able.

packages/password/src/index.ts
```diff
@@ -69,23 +80,31 @@ export default createPrompt<string, PasswordConfig>((config, done) => {
...
   let formattedValue = '';
-  let helpTip;
   if (config.mask) {
     const maskChar = typeof config.mask === 'string' ? config.mask : '*';
-    formattedValue = maskChar.repeat(value.length);
+    formattedValue = showPlaintext ? value : maskChar.repeat(value.length);
   } else if (status !== 'done') {
-    helpTip = `${theme.style.help(theme.style.maskedText)}${cursorHide}`;
+    formattedValue = showPlaintext ? value : theme.style.help(theme.style.maskedText);
   }
```
SBoudrias: This should start from `if (showPlaintext)` in its own condition, then `else if` blocks for the other 2 conditions.

packages/password/src/index.ts
```diff
@@ -4,39 +4,48 @@ import {
...
+      keys
+        .map(([key, action]) => `${styleText('bold', key)} ${styleText('dim', action)}`)
+        .join(styleText('dim', ' • ')),
   },
 };
 
 type PasswordConfig = {
   message: string;
   mask?: boolean | string;
+  allowShowPassword?: boolean;
```
SBoudrias: Let's rename this option to `toggleMask`. IMO it's clearer about what the user gets (a toggle over the existing mask), it's short and idiomatic, and it sidesteps the "allow" permission framing entirely. We'll also want the theming hook name to be aligned; so let's rename `showHideTip` → `toggleMaskTip` (or maybe `maskToggleTip`?)

SBoudrias:
Following up on the shortcut — let's swap `ctrl+space` for `ctrl+t` before merge.

**Why `ctrl+space` is risky:**
- On macOS and many Linux DEs (IBus, Fcitx), `ctrl+space` is bound to IME / input-source switching, so the keypress never reaches the process — exactly the terminal caveat you already flagged in the README.
- Even when it does arrive, terminals encode `ctrl+space` as the NUL byte (indistinguishable from `ctrl+@` / `ctrl+2`), so detection is fragile across emulators.

**Why `ctrl+t`:**
- Not intercepted by any OS-level shortcut on macOS, Linux, or Windows — it arrives as a clean `ctrl+t` keypress everywhere.
- Mnemonic: **T**oggle.
- No collision with Inquirer's existing keybindings (emacs `ctrl+p`/`ctrl+n`, vim `hjkl`, enter/tab/backspace/space).

With the key change, update the help tip label to `(ctrl+t toggle visibility)`

SBoudrias:
Thanks!!

PR #2254, opened 2026-09-05
fix(confirm): ignore surrounding whitespace in answers

sabtain05:
## Summary

- Trim confirm prompt input before matching yes/no keywords
- Add regression coverage for answers with surrounding whitespace

## Why

Users can accidentally include spaces when typing or piping answers into a prompt. Previously, values such as `  yes  ` did not match a keyword and silently fell back to the configured default. Trimming the input preserves existing matching behavior while accepting the same answer with incidental whitespace.


SBoudrias (approved):
Thanks, code LGTM!

PR #2255, opened 2026-09-06
fix(@inquirer/core): clear hook effects before settling prompts

### LINKED ISSUE #1816: Review async effect teardown
Right now we force prompt cleanups to be asynchronous. Though that can cause issue since keypress will keep triggering (see issue with Unix yes). It might be better to let hooks choose how they'll cleanup, sync or async.

JoelTowell:
## Description

Closes #1816.

`createPrompt` currently clears hook effects after the prompt promise has already settled. That means listeners installed by hooks can still handle input emitted in the same tick after `done()` or after the prompt is aborted.

This moves hook cleanup into the shared settlement path instead:

- resolve: clear hook effects, then resolve
- reject/cancel: clear hook effects, then reject

The final interface cleanup still stays in `finally`, so `screen.done()` and `output.end()` continue to run after the prompt reaches its done state.

I added regression coverage for both cases: settling from a `useKeypress` handler, and aborting the prompt.

## Commands run

- `corepack yarn pretest`
- `corepack yarn vitest --run packages/core/core.test.ts`

### DIFF
--- a/packages/core/core.test.ts
+++ b/packages/core/core.test.ts
@@ -392,6 +392,57 @@ describe('createPrompt()', () => {
     await expect(answer).resolves.toEqual('foo');
   });
 
+  it('useEffect: a throwing cleanup rejects the prompt and drops the answer', async () => {
+    const Prompt = (config: { message: string }, done: (value: string) => void) => {
+      useKeypress((key: KeypressEvent) => {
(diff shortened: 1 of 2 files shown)


SBoudrias (approved):
Thanks! This is a very clean PR! 👏🏻

Issue #1459, opened 2024-07-04
Bump `yoctocolors-cjs` to `2.1.2` for Deno compatibility


bombillazo:
Due to the lack of implementation of the tty library in yoctocolors, a patch was pushed to allow graceful error handling for the library with Deno. This package uses the yoctocolors package with the old version that causes inquerer to crash in Deno when using the latest version (since the move to yoctocolors)

bombillazo:
Alos, using `yoctocolors` will effectively make colors unusable in Deno until [this issue](https://github.com/denoland/deno/issues/24407) in the Deno repo is resolved. The previous color library worked for Deno.

SBoudrias:
Do you also have a ticket on yoctocolor repo?

bombillazo:
I'm sorry for not getting back to you sooner!

I'm using `@inquirer/prompts,` and it's worked great so far. I haven't had any issues, but now the colors in the terminal don't work when running with Deno since `yoctocolors` uses a color implementation that Deno does not currently support. It used to work when you were using `chalk` though.

bombillazo:
Hey, @SBoudrias, sorry for not getting back to you sooner!

I'm using `@inquirer/prompts,` and it's worked great so far. I haven't had any issues, but now the colors in the terminal don't work when running with Deno since `yoctocolors` uses a color implementation that Deno does not currently support. It used to work when you were using `chalk` though.

SBoudrias:
That's unfortunate, but given the population of Node vs Deno users, I think the decision to switch our color library does bring more value to more users. And we see Deno's working on filling the gap with the `tty` module, so sounds like it's a matter of months maybe? I think that's fair enough from the point of view of this library.

Issue #1085, opened 2022-02-15
Deno support


Kreijstal:
Would it be possible if Iniquirer.js also supported Deno?

SBoudrias:
I'm not quite sure what would be involved. But I think this could be complicated since Inquirer depends heavily on Node.js readline module.

Kreijstal:
when importing the module let the user provide a Readline implementation
https://deno.land/x/readline@v1.0.0
if none provided, then feel free to require("readline")

LitoMore:
Hi @Kreijstal, @guillaume86. Simply import `npm:@inquirer/prompts` to your Deno project.

See https://docs.deno.com/runtime/manual/node/npm_specifiers.

Closing.

Kreijstal:
now do browser support, say xterm.js

PR #2269, opened 2026-09-17
docs: document Deno installation

tomas-zijdemans:
## What

Adds a `deno` column to the installation tables in `@inquirer/prompts` and `@inquirer/i18n`, plus a one-line note that Deno scripts need `--allow-env`.

Deno already works — #1085 was closed with "Simply import `npm:@inquirer/prompts` to your Deno project" — but the install tables still list only npm/yarn/pnpm/bun, so neither the npm page nor the repo landing page (a symlink to `packages/prompts/README.md`) mentions it.

```sh
deno add npm:@inquirer/prompts
```

## Why the `--allow-env` note

Under Deno's permission model a bare `deno run` fails before the first prompt renders: `@inquirer/figures` reads `process.env` to detect unicode support, and `node:util`'s `styleText` reads it for color support. `--allow-env` on its own is enough. Without the note the documented install leads straight into a permission error.

## Verification

Against `@inquirer/prompts@8.7.2` on Deno 2.9.6, driving a real pty with `expect`:

- `deno add npm:@inquirer/prompts` in an empty directory creates `deno.json` + `deno.lock` and resolves 8.7.2
- `input`, `select` (arrow-key navigation, i.e. raw-mode TTY) and `confirm` all render and resolve correctly
- `--allow-env` is the minimum permission set; nothing else is requested
- Colors are at parity with Node: `styleText('green', …)` emits identical SGR codes under both runtimes when `TERM` is set. The `yoctocolors` problem from #1459 is moot now that core styles via `node:util`.

## Notes

Docs-only; no source changed, so I didn't run the unit or integration suites. `oxfmt` doesn't format markdown.

I included `@inquirer/i18n` because its install table is identical and would otherwise be the only one missing Deno — happy to drop it if you'd rather keep this to `prompts`.


SBoudrias:
I'm not against adding Deno specific instructions, but only if I can guarantees the library work. Right now there's no CI runs against deno or anything like that. For me this is a show stopper to document anything deno specific within the READMEs.

PR #2271, opened 2026-09-17
ci: add Deno compatibility test suite

SBoudrias:
## What

Adds a Deno compatibility suite to CI so all published packages keep working under Deno, continuously rather than by manual validation.

- New `integration/deno` workspace (sibling of the cjs/esm suites) with spot-checks: `deno check` over all cases, plus one prompt run per type — `input`, `confirm`, `number`, `select`, `checkbox`, `rawlist`, `expand`, `password`, `search`, an i18n-localized prompt (fr locale detection via `LC_ALL`), the legacy `inquirer` package, `@inquirer/figures` symbol fallback, and a prompt-error path (non-zero exit + message on stderr).
- New `Deno Tests` job in `main.yml`, matrixed over `v2.7.x`, `v2.8.x`, `v2.9.x` and `latest`, following the existing e2e pattern: `yarn isolate-monorepo-package` so tests run against the **packed/published artifacts** (dist + rewritten exports), not workspace sources.

## Design notes

- **One prompt per process.** Sequential prompts in a single process don't share piped stdin reliably; each case runs as its own `deno run --allow-env` child (same approach as the esm suite's `yes | node` test).
- **`--allow-env` is the only permission granted**, deliberately. If a future change needs more permissions than we document, CI fails and flags it — permission creep becomes a conscious decision.
- **Case files are type-checked twice**: by `tsc` (added to the root tsconfig include) and by `deno check`, catching declaration-file issues under Deno's own resolver.
- **Contributors without Deno aren't blocked**: the driver detects a missing `deno` binary and skips with a notice; `yarn test` still passes locally. CI runs the suite fully.
- 30s per-case kill guard so a hung prompt fails instead of stalling CI.

## Findings (worth knowing)

- **Deno ≥ 2.7 is the compatibility floor.** On ≤ 2.6, every prompt fails with `EOPNOTSUPP: Operation not supported on socket` — `deno_node`'s readline shim unconditionally enables raw mode and throws on non-TTY stdin. Verified on 2.0.6 / 2.1.14 / 2.4.5 / 2.5.7 / 2.6.10 (fail) vs 2.7.14 / 2.8.3 / 2.9.x (pass). The matrix floor is set accordingly.
- Non-TTY `search` doesn't apply the filter before Enter selects (under Node the same case just hangs); the `number` retry loop doesn't recover buffered input across lines. Both are non-TTY quirks present in both runtimes, not Deno regressions — excluded from the suite. TTY/interactive usage was validated against real ptys in #2269.

Relates to #2269 (Deno install docs), #1085.

## Verification

- Full matrix locally against packed artifacts via `isolate-monorepo-package`: Deno **2.7.14, 2.8.3, 2.9.6** → 14/14 tests each; 2.0–2.6 → fail with EOPNOTSUPP as documented above.
- `yarn oxlint .`, `yarn oxfmt --check .`, `yarn eslint .`, `yarn tsc` — clean.
- `yarn test` — full suite green (vitest + turbo, including the new workspace at repo root).


SBoudrias:
Restructured per review — answers to both questions:

**Why `YARN_ENABLE_IMMUTABLE_INSTALLS: false` was there:** the isolated temp project has no lockfile, and Yarn defaults to immutable installs in CI, so the post-isolation `yarn install` would fail without it. It was cargo-culted from the e2e job (which needs it for the same reason). It's now confined to the `deno-pack` job with a comment; the repo-level install still passes `--immutable`.

**Why `setup-node` was there:** `yarn tsc` (build) and `isolate-monorepo-package` (packing) are Node tooling, so it leaked into the test job. Now Node is confined to a single `deno-pack` job:

- **`deno-pack`** — install, `tsc`, isolate, install the isolated project, upload it as a tarred artifact (~2MB).
- **`deno` matrix** — `setup-deno` + `download-artifact` + `deno test --allow-env --allow-run=deno`. **No checkout, no Node, no Yarn, no env vars.** The whole test phase runs in the Deno world, including the driver itself (`deno test` running the node:test API, which Deno implements natively). Children are restricted to spawning `deno` only via `--allow-run=deno`.

Verified locally against packed artifacts with the new runner: Deno 2.7.14 / 2.8.3 / 2.9.6 → 14/14 each, including a cold-cache simulation (fresh `DENO_DIR` + artifact tar roundtrip).

Two incidental fixes in the same push:

1. The driver's `deno check` step now pins its cwd (it previously inherited the invoking directory, so running the suite from the repo root failed).
2. Dropped the `tsconfig.json` include for the case files. Root cause discovered while chasing a reproducible `deno check` failure at the repo root: Deno applies the covering `tsconfig.json`'s `compilerOptions` when checking matched files — and this repo's base config sets `lib: ["es2023"]`, which has no `console`. Cases are now type-checked solely by `deno check` (CI, published artifacts) and linted by oxlint; they're ignored in eslint's typed-lint scope since the project service requires tsconfig coverage that conflicts with Deno's config discovery.

SBoudrias:
**Follow-up: the CI `v2.7.x` flake and its fix.** The first run of the pure-Deno matrix failed 3/14 tests (`input`, `expand`, `password`) with 30s hangs. Diagnosis from the job log: the three hung children each waited the full `CASE_TIMEOUT` while the other 11 passed, and the total job time (~31s) proved the tests had run **concurrently** — Deno's node:test compatibility layer overlaps `it()` bodies, unlike Node's sequential execution. On the 2-core CI runner with 14 cold-starting Deno children, Deno 2.7 occasionally lost piped stdin data (the prompts rendered and waited forever — 2.8/2.9 dodged it, likely by faster child startup).

Fixed by serializing test bodies through a shared promise queue (`sequential()` wrapper in the driver) — behavior is unchanged under `node --test` (already sequential), and it makes the suite deterministic under `deno test` on all matrix versions. Matrix legs now complete in 7–12s.

SBoudrias:
Dropped the node-based `test` script per review — the suite now only ever runs under Deno:

- `test` script removed, so `yarn test` / `turbo test` no longer executes it via node (turbo shows the workspace with a `<NONEXISTENT>` task, i.e. skipped).
- Added `test:deno` (`deno test --allow-env --allow-run=deno integration.test.mjs`) as the discoverable local entry point for contributors who have Deno installed.
- The driver's `hasDeno` graceful-skip logic was dead code after this change, so it's removed along with the node-runner documentation.

Verified: full `yarn test` green (deno package not executed), `yarn workspace @inquirer/deno-integration-tests test:deno` → 14/14, and the isolated CI path across the matrix floor (2.7.14 / 2.8.3 / 2.9.6) → 14/14 each.

